PR #844, opened 2013-11-21
Rowspan

bastienJS:
I would like to pull a feature request that you implement a rowspan over 2 or more rows.

This feature greatly help me improving my cell templates when I have a double lessons to preparel

It should be possible to decide during runtime wether to render a row with a single cell or a merge cell/cells depending on a viewmodels property.

image

### DIFF
--- a/Brainstorm.md
+++ b/Brainstorm.md
@@ -6,8 +6,9 @@
 ## Outstanding Problems
 
 * pinned columns performance
-* ???
-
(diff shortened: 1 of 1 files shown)


trupti18:
Hi All,
Has this(ROWSPAN)  feature been included in the grid or not?
This is a very useful feature and it should be part of standard ui-grid code.

renfengmei:
i am looking for this feature, has it done? and how to use?

Issue #581, opened 2013-07-22
ng-grid with ui-bootstrap: Grids in hidden tabs are not shown


zussel:
Having multiple grids on a page each in a seperate tab, data of hidden grids is not shown. It appears when resizing the window.

See it here: http://plnkr.co/edit/S1Po1qugA03Ntc4vF0nt?p=preview

I tried to use the select event, but that works only with one hidden grid not with multiple grids.

PR #991, opened 2014-02-13
Fix issue #581

getuliojr:
Fix the problem of the ng-grid not showing when inside of a tab or a div that is hidden. The problem surged because of the class ng-hide that had an !important in its class. Fixed the problem by adding an important at the element when getting real width.


ackimwilliams:
This issue was not addressed. I am still experiencing it.

PR #1169, opened 2014-05-07
fixed retrieving value for group header from booleans

lario:
When grouping by a column with boolean values the grouping is not correct.

http://plnkr.co/edit/FeTf4DN7hIx7ZJSdQgTO?p=preview

When grouping by the 'test' column instead of getting 3 groups: 'true', 'false' and 'null' 'false' and 'null' are grouped together into 'null'

https://github.com/angular-ui/ng-grid/blob/master/src/classes/rowFactory.js#L198 should check if val != null

### DIFF
--- a/src/classes/rowFactory.js
+++ b/src/classes/rowFactory.js
@@ -195,7 +195,7 @@
                 var col = filterCols(cols, group)[0];
 
                 var val = $utils.evalProperty(model, group);
-                val = val ? val.toString() : 'null';
+                val = val != null? val.toString() : 'null';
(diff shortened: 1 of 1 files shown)


norsang:
This commit breaks group by field if the field is missing in any of the row. This is because of val.toString() which doesn't work on undefined.

http://plnkr.co/edit/9TaFbiAIsVSAObPpcgO4?p=preview

Should be changed to:
<pre>val = (val === "" || val === null || val === undefined) ? 'null' : val.toString();</pre>

PR #1392, opened 2014-08-11
Bugfix: Allows a column name to be any string, including special characters.

Manduro:
Use the brackets syntax instead of the dot syntax for selecting a column so it allows for column names with special characters.

For example, using an UUID as a column name such as "3f73c374-bf4e-4c64-8d07-d5ef722b5cba":

row.entity.3f73c374-bf4e-4c64-8d07-d5ef722b5cba would fail because it interprets the dashes as minus signs.

row.entity["3f73c374-bf4e-4c64-8d07-d5ef722b5cba"] works.


lvarayut:
+1

PR #1921, opened 2014-10-26
Use copies of GridRows to eliminate DOM thrashing

markrendle:
I've been experimenting and profiling (with IE11), and the way the `renderedRows` array was being updated was causing the `uiGridViewport` contents to be completely re-rendered by Angular on scrolling.

This change basically creates a separate set of `GridRow`s in the `renderedRows` array, then copies the values from `newRows` into them instead of replacing the elements in the array.

I don't know what I've missed with this, I've probably broken a ton of stuff, but... check the performance difference out in Chrome. :dash:


swalters:
Closing this PR with the assumption that track by $index gives us the same result.  @markrendle if you see it differently, please re-open.

Issue #1838, opened 2014-10-16
headerRowHeight not work


dagouaofei:
I want to change the header width, but it not take effort. See http://plnkr.co/edit/eyDrjy4YFB35LnG5AiIR?p=preview

PR #2060, opened 2014-11-11
fixed issue #1838, grid option: headerRowHeight

HampusLilja:
Issue #1838 fixed.
When setting the grid option headerRowHeight the height is now set to that value.


nileshmali:
It is possible using CSS but then `headerRowHeight` option is of no use. Currently this option does nothing.

kolesnik:
Guys, let's replace "headerRowHeight" option from API documentation with comment about CSS

c0bra:
@kolesnik Sounds good to me.

c0bra:
This is done.

PR #2118, opened 2014-11-16
Adding variables to subgrid scope

jpuri:
The PR has code changes for:
1. Grid option 'expandableRowScope' can now be used to add variables to scope of expanded grid.

Now we have 3 options for expanded grid, I wanted to group them into single grid option object, but I am afraid that subgrid feature is already in use at many places and this change might break code.

Also, might be being able to map 'expandableRowScope' to a scope variable of controller will be more useful, but that would have additional overhead of watchers. Please suggest.
1. Column moving feature has issues on IE:
   1. Moved columns are not positioned correctly.
   2. Styles of moving column is not visible.
      The PR will fix issue 1 above.


asafliv:
Hi!

Can someone please explain a full solution for this issue? in this thread there are 5 different solutions.
I've tried them all and couldn't get my subgrid ng-click button to call a function from the outer scope.

Please I'm lost.

PR #2218, opened 2014-11-24
New option: sectionHeaders - display column headers within section labels

russell-miller:
At Rentrak we had a need for an Angular grid that could display column headers in sections.

Our fork of ui-grid does so by swapping out the headerTemplate with a slightly different one that iterates over a new option `sectionHeaders` to eventually find values within `columns` array(s) - those being the names of real grid columns that it renders as it normally would.

Example:

```javascript
$scope.gridOptions = {
  data: [],
  columnDefs: [
    { name: 'col1' },
    { name: 'col2' }
  ],
  sectionHeaders: [
    { displayName: 'Section', columns: [ 'col1', 'col2' ] }
  ]
};
```

Another need we had was to be able to take this as deep as we want per grid. So this is defined recursively in the new template. This means you can put sections within sections using the attribute `subSections`.

Example:

```javascript
$scope.gridOptions = {
  data: [],
  columnDefs: [
    { name: 'col1' },
    { name: 'col2' },
    { name: 'col3' }
  ],
  sectionHeaders: [
    { displayName: 'Super Section', subSections: [
      { displayName: 'Section A', columns: [ 'col1', 'col2' ] },
      { displayName: 'Section B', columns: [ 'col3' ] }
    ]
  ]
};
```

Screenshot:
section-headers

*Possible Concerns*:
 * maybe this should have been a separate module instead of part of core? not sure how much more work would be entailed
 * there was a seemingly minor feature loss for any grids with section headers - the following style is not applied to the first column header in the template `ng-style="$index === 0 && colContainer.columnStyle($index)"`
 * no tests for the recursively defined directive - would be open to suggestions for how to add testing

[Review on Reviewable](https://reviewable.io/reviews/angular-ui/ng-grid/2218)


mportuga:
@PaulL1 You marked this as waiting-response, but it seems that @russell-miller has already addressed all of @c0bra's concerns. It looks good to me, but I was wondering if you had any lingering concerns. Same goes for @c0bra.

russell-miller:
@mportuga I forgot I still had this request open... I ended up taking a different approach actually. There are conflicts with master now, and I hadn't fully tested that my changes didn't break other features of ui-grid. I would vote to just decline the request, though you may feel free to pick up where I left off.

cc: @PaulL1

mportuga:
@russell-miller Thank you for your reply. I am will simply decline it now as you've taken a different approach. I hadn't noticed the conflicts earlier, so thank you for pointing that out as well. I am new to the ui-grid team and I am trying to clean up all of these outstanding Pull requests.

PaulL1:
@mportuga: back then it was just a process to keep a track of why something wasn't progressing.  I have no opinion on it today.  :-)

PR #2326, opened 2014-12-11
Add feature: Entity dependent edit drop down options

### LINKED ISSUE #2309: Drop down with options dynamic by row
With the current implementation you set the editDropdownOptionsArray like follows

```
editDropdownOptionsArray: $scope.selectOptions      
```

However if the options are different depending on the row is there a way to handle that?  I thought about doing it in a beginCellEdit event but apparently there is nothing to set it onto that will affect a change in the original definition.

Is there currently a way to do this?  It seems like it would be a pretty common use case.

wgorder:
Add the ability to populate cell drop dowmn menus from the row.entity by providing a path
to an options array on the entity. This allows for options to differ on a row by row basis.

closes #2309

### DIFF
--- a/misc/tutorial/201_editable.ngdoc
+++ b/misc/tutorial/201_editable.ngdoc
@@ -40,6 +40,7 @@ __ColumnDef Options__:
    `'code'` instead you can use it without having to reprocess the array
 - `editDropdownValueLabel` (default: `'value'`) Controls the value label in the options array - if your array happens to
   contain `'name'` instead you can use it without having to reprocess the array
+- `editDropdownRowEntityOptionsArrayPath` can be used as an alternative to editDropdownOptionsArray when the contents of the dropdown depend on the entity backing the row.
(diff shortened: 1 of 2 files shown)


chandanjena:
I have a simple requirement. I am using angular ui-grid to display records fetched from database. I have a drop down field named Status. This drop down values are A, B, C, D, E. I need to display all the values if the logged in user ($scope.profile = Admin) is Admin. But if the user is Supervisor ($scope.profile = Supervisor) then, only A,B and C should be visible in the drop-down.
How can I achieve this? I'm very new to Angular as you could have found by now.

here is my columnDefs for the drop-down.
{ name: 'Status', field: 'Status', displayName: 'Status', editableCellTemplate: 'ui-grid/dropdownEditor',
			      cellFilter: 'mapStatus', editDropdownValueLabel: 'status', editDropdownOptionsArray: [
			      { id: 'A', status: 'A' },
			      { id: 'B', status: 'B' },
			      { id: 'C', status: 'C' },
			      { id: 'D', status: 'D' },
			      { id: 'E', status: 'E' }]
			    }

Thnaks in advance.

PR #2390, opened 2014-12-18
2.x - Expandable rows and actions column

gwin003:
Added ability to expand rows using an actions column in the grid
Actions column contains an edit, delete and expand button
Each of the buttons can be hidden or disabled
Able to supply callbacks for the edit and delete buttons so you can handle functionality in your application
HTML template that shows in the expandable div can be defined inside your ng-grid directive, and has access to the row entity

Note: The action column icons uses FontAwesome, so that will be required to use the buttons, unless the template is overridden.

[Review on Reviewable](https://reviewable.io/reviews/angular-ui/ng-grid/2390)


moberemk:
This hasn't been touched in a while, but it's functionality which my project could definitely use--any chance this will get merged in soon @c0bra?

gwin003:
@moberemk Have you taken a look at what I did on https://github.com/gwin003/ng-grid? My main concern is what I did was too specific for our issue.

JLLeitschuh:
Your second commit says what the PR is not what it adds/removes.
Your message for each commit should do (at least some of) the following:
Answer the following questions:

1. Why is this change necessary?
  This question tells reviewers of your pull request what to expect in the commit, allowing them to more easily identify and point out unrelated changes.

2. How does it address the issue?
  Describe, at a high level, what was done to affect change. Introduce a red/black tree to increase search speed or Remove \<troublesome gem X\>, which was causing \<specific description of issue introduced by gem\> are good examples. If your change is obvious, you may be able to omit addressing this question.

3. What side effects does this change have?
 This is the most important question to answer, as it can point out problems where you are making too many changes in one commit or branch. One or two bullet points for related changes may be okay, but five or six are likely indicators of a commit that is doing too many things.

This was taken from here: https://robots.thoughtbot.com/5-useful-tips-for-a-better-commit-message

faris-git:
I'm new to ui-grid. I'm looking for a feature expandable columns with combination of infinite scroll. Is this issue related to new feature which support expanding columns(having sub columns)?

mportuga:
@gwin003 I am closing this PR due to inactivity. I realized this was partially our fault since there weren't a lot of active developers for a bit, but since you weren't able to address @JLLeitschuh's concerns. @faris-git If you would like this feature feel free to open a new PR for it at your leisure. @gwin003 is also free to do the same if he wishes.

Issue #2357, opened 2014-12-16
scrolling with large amount of columns,columns disappear after sort  


gotenxds:
Hi,
In the demo application I move the scrollbar to last column "coll499" when i click on the column header 
to sort the data,the column disappaer
the grid before i clicked
![image](https://cloud.githubusercontent.com/assets/3519520/5452385/f2e7a2d6-8522-11e4-91d1-21d707178256.png)
after the click
![image](https://cloud.githubusercontent.com/assets/3519520/5452375/ba55ce5c-8522-11e4-82cd-9aea031208bb.png)
I am trying this from the  [this url!](http://ui-grid.info/docs/#/tutorial/191_horizontal_scrolling) from ui-grid website.

note:in my project its also disappear when try to edite the one of the last few column

leo009394:
hi @c0bra 
I still have found the same issue when a cell is in edit. Have you fixed this issue or I missed something
Here is the <a href="http://plnkr.co/edit/QzLtW9rvgTkSVdazRlCJ?p=preview">plunker</a>

stefan-leye:
I have the same issue

Niranthari:
hi @c0bra  i still have the issue regarding the scroll bar disappears while i am trying to edit the cell could you pls help me.its very urgent.Thanks in advance

PR #2425, opened 2014-12-23
Regression Fix for #2357 the scrollPercentage is not correct 

jcompagner:
if there are added rows, because then the scrollPercentage should be recalculated.

so its an enhancement on this fix:
https://github.com/angular-ui/ng-grid/commit/bc2c68aba48cdd5a3e840a589fb90cdba363618c


c0bra:
I've merged  e497eac. Thanks @jcompagner!

Issue #2186, opened 2014-11-21
Any plans for copying cell data from grid?


marcuswhit:
I'm fairly sure this isn't currently possible, so please correct me if wrong, but are there plans to enable ctrl+c copying of grid data that is selected?  Ideally this would copy as csv data I guess.

PaulL1:
I believe this has been addressed, the PR has been merged.  Closing.

PR #2464, opened 2014-12-29
Enh(cellNav) #2186 select multiple cells

mathiasdose:
Add modifierKeysToMultiSelectCells option which enables user to,
when including cellNav, hold the modifier key to select multiple
cells with the mouse. Add API method getCurrentSelection which
enables the user to get the cells currently selected.


PaulL1:
The tests look complete, the code looks clean, it's localised to cellNav.  So....merging.  Thanks for this @mathiasdose

Issue #2483, opened 2015-01-05
Create a Grid Footer similar to the 2.x grid footer


swalters:
2.x had the ability to enable a grid Footer that would show:
Total Rows
Selected Rows
Filtered Rows

Create a GridFooter in core that append to the grid and does not scroll horizontally with the column scroll.

camacazze:
Does this mean that the Grid Footer is not working correctly??

I am asking bcoz the Tutorial documentation says that it does.

I am facing a problem where the footer is not working.

Pls. reply.

PaulL1:
It is working for me.  What problem exactly are you having?  The functionality that you see in the tutorial should be working, so long as you've grabbed the unstable version.

PaulL1:
I think this is now implemented.  Closing.

PR #2484, opened 2015-01-05
feat[Core] commit for #2483.  Adds a showGridFooter option that will cre...

swalters:
...ate a grid footer with the Total Items and Shown Items displayed.  Also shows Selected Items if using Selection feature.

 Breaking Changes in this commit:
 showFooter option renamed to showColumnFooter
 footerRowHeight option renamed to columnFooterHeight


swalters:
Can't merge this yet.  I am seeing issues when columns are pinned right due to the grid footer interfering with the right render container

PR #2490, opened 2015-01-05
Fix #2489: Column moving not working well with horizontal scrolling

### LINKED ISSUE #2489: Column moving not working well with horizontal scrolling
If the table has horizontal scrolling and the columns are movable they will get into positions that are different from the ones intended when dragged to the right or left with the scroll as can be seen in the following plunker:
http://plnkr.co/edit/pBotvpDTiUdXn2U1uIW8?p=preview
Furthermore in Chrome any text over which the mouse cursor passes inside and outside of the table gets selected when moving the columns.

edmondpr:


jpuri:
Hi @edmondpr , 
Thanks a lot for this nice change to move column mode. Just a small refactoring I have put as suggestion. Else it looks great.

PaulL1:
I'm OK to merge this, but I see it now has a conflict with a change in master.  Is it possible to rebase off master?

PR #2496, opened 2015-01-07
Fix #2461: let the parent container scroll also on touchmove event

### LINKED ISSUE #2461: Parent/page scrolling doesn't work when grid scrollbars are disabled
Hi, I disabled grid scrollbars using

```
$scope.gridOptions.enableHorizontalScrollbar = uiGridConstants.scrollbars.NEVER;
$scope.gridOptions.enableVerticalScrollbar = uiGridConstants.scrollbars.NEVER;
```

and I expected the grid scroll events to be disabled as well. Imho, in this case, the page (or parent div) scrolling should be working when onMouseOver the grid.

Is this a bug? Or how can I prevent the grid from stealing the scroll events that should be propagated to its parent?

VojtechVitek:
A follow-up commit to 924af44b7ac7c3087d24427cc5904707da5cdf85 @kenwarner

### DIFF
--- a/src/js/core/directives/ui-grid-render-container.js
+++ b/src/js/core/directives/ui-grid-render-container.js
@@ -172,8 +172,6 @@
                 event = event.originalEvent;
               }
 
-              event.preventDefault();
-
               var deltaX, deltaY, newX, newY;
               newX = event.targetTouches[0].screenX;
               newY = event.targetTouches[0].screenY;
@@ -205,6 +203,11 @@
                 scrollEvent.x = { percentage: scrollXPercentage, pixels: deltaX };
               }
 
+              // Let the parent container scroll if the grid is already at the top/bottom
+              if ((scrollEvent.y && scrollEvent.y.percentage !== 0 && scrollEvent.y.percentage !== 1) ||
+                  (scrollEvent.x && scrollEvent.x.percentage !== 0 && scrollEvent.x.percentage !== 1)) {
+                event.preventDefault();
+              }
               scrollEvent.fireScrollingEvent();
             }
             


PaulL1:
OK, looks useful.  I won't merge at this time because I know @swalters is doing a lot of change and I don't want to break his merge.  But we'll look after that.

VojtechVitek:
Any update on this?
@kenwarner @PaulL1

PaulL1:
I believe this code is already included - I saw it when I was working through the scrolling changes.  Are you able to look at the current code base and check that it would merge correctly?

VojtechVitek:
Seems like it's not included. I will rebase tomorrow and update this PR.

VojtechVitek:
@kenwarner @PaulL1 rebased.

PaulL1:
It looks to be there in master at present - a slight variation on the code but very similar:

```
              // todo: this isn't working when scrolling down.  it works fine for up.  tested on Chrome
              // Let the parent container scroll if the grid is already at the top/bottom
                if ((scrollEvent.y && scrollEvent.y.percentage !== 0 && scrollEvent.y.percentage !== 1 && containerCtrl.viewport[0].scrollTop !== 0 ) ||
                    (scrollEvent.x && scrollEvent.x.percentage !== 0 && scrollEvent.x.percentage !== 1)) {
                  evt.preventDefault();
              }

              scrollEvent.fireThrottledScrollingEvent();
```

As such, closing.

VojtechVitek:
@PaulL1, the code that you mentioned is for mousewheel scroll event (https://github.com/angular-ui/ng-grid/blob/master/src/js/core/directives/ui-grid-render-container.js#L123) and fixes the #2461 issue for desktop:
```
 // Scroll the render container viewport when the mousewheel is used
            $elm.bind('wheel mousewheel DomMouseScroll MozMousePixelScroll', function(evt) {
```

However, this patch fixes the touchmove event instead (https://github.com/angular-ui/ng-grid/blob/master/src/js/core/directives/ui-grid-render-container.js#L173) and thus fixes the same #2461 issue, but on touchpads (mobile/tablet).

Can you please re-open?

PaulL1:
Ah, OK then.  I'll merge.

VojtechVitek:
Thanks!

PR #2501, opened 2015-01-07
2.0.8 Pagination fix

DaDanny:
I found a bug with ng-grid that was not displaying the correct amount of serverItems for a list.

This was because $scope.totalServerItems in the footer.js file was never defined. 

To fix this bug, I changed it from $scope.totalServerItems to $scope.pagingOptions.totalServerItems

[Review on Reviewable](https://reviewable.io/reviews/angular-ui/ng-grid/2501)


PaulL1:
@c0bra may look at this on the next 2.x merge he does (if any).  I note that:

- it has 5 commits, normally we'd only merge things where the commits are squashed. 
- there seems to be a weird change in `ng-grid.debug.js` that I don't quite understand, I'd have assumed that was a built file
- the travis build failed, I'm not sure if that's normal on 2.x or is a problem.

PR #2513, opened 2015-01-08
Scroll error, add check for x/y

brianchance:
Scrolling horizontally does not set x which throws error, added check for x/y.


PaulL1:
This looks reasonable, and I think will be still needed after @swalters merges his scroll changes.  But I will wait for him to merge that, and then maybe he can quickly check this will still merge and is still needed.

swalters:
Thanks for the PR.  I found the same issue while reworking scroll bars so this PR won't be needed.

PR #2516, opened 2015-01-08
add utf-8 support on csv export

rlebosse:
This sets exports to use UTF-8. This prevents encoding errors in some spreadsheet programs such as excel.


PaulL1:
Thanks @rlebosse

PR #2520, opened 2015-01-09
select cells on mousedown, so it works also when mouseup is fired on different cell ( from the same row )

gboros:
this is needed, so for example, if you start to select the text content of a cell with the mouse, and the
mouseup will be on the next/prev cell from the same row, the selection is still set


c0bra:
Do we want to globally allow selecting multiple cells via the mouse? If so, do we also need to allow the same sort of selection via keyboard nav?

PaulL1:
I think this is fixed with the newer touch event handling - the down event is on a specific cell, but the up event is anywhere on the document.  Suggest we can probably close this.

PR #2558, opened 2015-01-15
Fix #2297 (header/footer): change clearfix to ui-grid-clearfix

### LINKED ISSUE #2297: Improper use of non-existent "clearfix" class
Both the `ui-grid-footer` and `ui-grid-header` templates have a `clearfix` class on them. The grid defines no "clearfix" class nor can we rely on it being there.

These references should probably be updated to `ui-grid-clearfix` which DOES exist, but who knows what effect that could have. Someone will need to test this out to make sure there's no problems with the rendered elements.

PaulL1:


PaulL1:
I've completed the change as requested in #2297, and found no ill effects in working through the tutorials.   Merging and will wait for any reports of issue.

PR #2568, opened 2015-01-15
Created expand all rows at the top row

drixta:


PaulL1:
Looks good, and clean and well documented code.  I'll merge.

PaulL1:
You may want to consider unselecting 'expandedAll' when they collapse one or more rows.  It's not imperative, but I ended up doing it on selectAll because it just looks funny otherwise.

PR #2578, opened 2015-01-16
Add Index for Row Indexing

jhawlwut:
Allows for the row to be indexed with a new column corresponding to the data in the view.
Example of adding the column using a controller:
colDefs.push({'name': 'Index', 'displayName': 'Index', 'width': '60', 'visible': true,
               'cellTemplate': '<div class=\"ui-grid-cell-contents\">{{grid.options.paginationPageSize*(grid.options.paginationCurrentPage-1)+row.index+1 }}</div>', });

screen shot 2015-01-16 at 1 26 07 pm

### DIFF
--- a/src/js/core/factories/GridRow.js
+++ b/src/js/core/factories/GridRow.js
@@ -47,13 +47,21 @@ angular.module('ui.grid')
     // Default to true
     this.visible = true;
 
-  /**
+    /**
     *  @ngdoc object
     *  @name height
     *  @propertyOf  ui.grid.class:GridRow
     *  @description height of each individual row
     */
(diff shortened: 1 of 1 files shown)


PaulL1:
Is there a reason this needs to be in the grid code, rather than just being a column people add to their data if they want it?

jhawlwut:
Can't account for others data but we used this to build the index column for our internal use(s). The reason we did this was because our user data may not contain an index attribute; therefore, making it necessary for each row. In ng-grid, we were using "this.rowIndex = rowIndex;" and couldn't find it in the current ui grid. For us, this was bringing back that feature from ng-grid.

PR #2591, opened 2015-01-19
Fix Error: container.visibleRowCache is undefined.

Ahmed-Gemmy:
Replace
container.visibleRowCache.length = 0;
With
 container.visibleRowCache= [];
Because empty array is being considered as undefined which doesn't have a property called length

### DIFF
--- a/src/js/core/factories/Grid.js
+++ b/src/js/core/factories/Grid.js
@@ -1148,7 +1148,7 @@ angular.module('ui.grid')
       var container = self.renderContainers[i];
 
       container.canvasHeightShouldUpdate = true;
-      container.visibleRowCache.length = 0;
+      container.visibleRowCache = [];
     }
     
     // rows.forEach(function (row) {


PaulL1:
Looks reasonable.  I have a small concern that there might be a watch on visibleRowCache, and I think setting to a new array instead of emptying the existing one might break that (hypothetical) watch. But I don't see any reason we'd be watching visibleRowCache, so merging.

PaulL1:
@Ahmed-Gemmy : this broke some things in cellNav/edit, so I've changed the code to:
```
      if ( typeof(container.visibleRowCache) === 'undefined' ){
        container.visibleRowCache = [];  
      } else {
        container.visibleRowCache.length = 0;  
      }
    }
```

Could you recheck the situation that was giving you the error to make sure it still achieves what you were intending?

Ahmed-Gemmy:
Fantastic .. this is much more accurate 
Thanks Paul.

PR #2629, opened 2015-01-25
Fix #1671 (gridMenu): move columnMenu in rightmost column for gridMenu

### LINKED ISSUE #1671: Extend grid menu
The aim of the grid menu is to have a home for user actions that can be taken on the grid as a whole.

The most immediate need is to provide a way to show columns that have been hidden, similar to in 2.x.  But other features might also make use of it, including the exporter.  It might also be desirable to let an end user control some core gridOptions, for example to turn filtering on and off (as an example).

The general look and feel of the menu would be similar to 2.x - a menu button in the top right, which floats above the core grid.  The columns would scroll underneath it.  The main issue with this is that it would overlay the column menu for the rightmost column - which might be in the body renderContainer, or might be in the right renderContainer.  The header (or at least the columnMenuButton) for this column would need to be moved inward a little.

The menu itself should use the existing gridMenu, and should be switchable (so it can be turned off) and customisable (so the developer can add items to it if they wish).

PaulL1:


gboros:
I think this fix may cause problems if the last column has a smaller width; as you can see in this
screenshot, the columnMenu will move too much to the right, moving to another column
wrong-last-column-menu

if I just remove that hard-coded margin for the last column, it seems to render fine
correct-last-column-menu

PR #2660, opened 2015-01-28
added "isRowSelectable" to options

pedroclayman:
The "isRowSelectable" is a new option that enables to add some logic to whether a row is selectable or not.

The value assigned to the option has to be a function that accepts a row as parameter and has to return a boolean that specifies if the given row is selectable or not.

Here is a sampe of such a function
```javascript
var options = {
  isRowSelectable: function(row) {
    // every row that has an entity with the name "Grid" is selectable
    return row.entity.name === 'Grid';
  }
}
```


PaulL1:
Looks good.  Merging.

PR #2663, opened 2015-01-29
Remove $scope.$emit('hide-menu'); fixes #2662

### LINKED ISSUE #2662: Possible to leave grid menu open after changing visibility of a single column - so can change visibility multiple columns?
I've got a dataset which returns ~30 columns, as mocked-up here: http://plnkr.co/edit/Kr2ppkiwxHOU0iWOxaO0?p=preview

My end users are telling me that when they are using my tool, they often have to hide/remove 5 to 10 columns; and opening the grid menu each time slows them down significantly. The old UI I had for them (using a different library) had each of the columns' name with a checkbox beside it, and they could clickity-click remove 10 columns in 2 seconds, where now each column takes 2 seconds to remove. 

Is there any way to keep a constantly-open grid menu outside of the results/table, so that it would be more repeatedly accessible?

juliusakula:
Removing this line of code keeps the menu open until the user is done with it. This allows for multiple actions while the menu stays open.

### DIFF
--- a/src/js/core/directives/ui-grid-menu.js
+++ b/src/js/core/directives/ui-grid-menu.js
@@ -236,8 +236,7 @@ function ($compile, $timeout, $window, $document, gridUtil, uiGridConstants) {
               }
 
               $scope.action.call(context, $event, title);
-
-              $scope.$emit('hide-menu');
+              
             }
           };
 


PaulL1:
@juliusakula : I think we need to decide whether we'd want to leave the menu open for all actions, or just for column hiding related actions.  I think the existing behaviour is more what people would expect in most cases.  I'll consult the guys to see what they thing.

farooqu:
I think leaving the menu open for all actions makes sense. You can imagine a grid being loaded with many hidden columns and a user may want to un-hide multiple columns. Not having to re-open the menu each time would make lots of sense.

juliusakula:
I still think this fix would be helpful

PaulL1:
I'm planning to implement somethign similar when I get around to it, but this one I think I can't merge because:
1. It applies to all menus including the column menus.  I don't think that makes sense, it needs to apply only to the gridMenu
2. I'd like it to apply only to the column show/hide within the grid menu, not to other things like exporting a file
3. I'd like it to be switchable so that people can choose to keep the current behaviour if they want.

PaulL1:
I've implemented a similar concept, but with an option on it so that only those menus and menu items that we want to keep open would stay open.  Closing this.

Issue #2661, opened 2015-01-28
RowTemplate is not applied when the data is changed


blsr:
RowTemplate is not applied when the data is changed.
How can I fix this?
Please check this plunker
http://plnkr.co/edit/cHTyGLT44v69rAWzkWGv?p=preview

PaulL1:
I need to write a tutorial on dataChange I think - it's included in some of the other tutorials, and in the API documentation, but it's clearly not visible enough.

If you're dynamically changing your data you need to tell the grid that you have done so, which you do using the `notifyDataChange` api.  The grid doesn't have a watcher on every data cell looking for change, as that causes a lot of extra work during scrolling and row virtualisation, so we rely on you telling us when you've changed any data.

Refer: http://plnkr.co/edit/3QslUtHHGL8xUnEkJdJS?p=preview

PaulL1:
Closed above.

PR #2681, opened 2015-01-31
Fix #2678 (doco): new dynamic data tutorial

PaulL1:


PaulL1:
Damn.  Done it again.  This actually fixes #2661

PR #2695, opened 2015-02-03
Fixed NOT_EQUAL filter

axelstudios:
`uiGridConstants.filter.NOT_EQUAL` was actually returning exact matches

### DIFF
--- a/src/js/core/services/rowSearcher.js
+++ b/src/js/core/services/rowSearcher.js
@@ -209,7 +209,7 @@ module.service('rowSearcher', ['gridUtil', 'uiGridConstants', function (gridUtil
     }
 
     if (filter.condition === uiGridConstants.filter.NOT_EQUAL) {
-      return angular.equals(value, term);
+      return !angular.equals(value, term);
     }
 
     if (typeof(value) === 'number'){


PaulL1:
It's not clear to me that this is the problem.  In this plunker: http://plnkr.co/edit/78tvqOJ0mkM2OYbeMpea?p=preview I've changed the filter on e-mail to NOT_EQUAL.  If the problem were as you suspect, then it should be working as equals instead of not-equals, but if I put `bishopcarr@digirang.com` into the filter, it doesn't return that row.

I'll look at it more closely later.

axelstudios:
It's funny, I did that *exact* same test down to using that same email address..  I realize now that the problem in that case is that `@` is being escaped, so maybe testing with !regex would be a better solution than !angular.equals

PaulL1:
It looks to me like it's escaping the `.`, not the `@`.  But that may be browser specific.  Let me try !regex.

PaulL1:
OK, that works when I use:

```
    if (filter.condition === uiGridConstants.filter.NOT_EQUAL) {
      var regex = new RegExp('^' + term + '$');
      return !regex.exec(value);
    }
```

I'll do a separate pull request with that.

axelstudios:
Thanks!

PR #2716, opened 2015-02-06
Add Japanese translation

tanak:


PaulL1:
Thanks for this @tanak

Issue #2535, opened 2015-01-12
Infinite scroll 'blank block'


cookednoodles:
When using the infinite scroll at http://ui-grid.info/docs/#/tutorial/212_infinite_scroll

If you scroll down, content will load but not be shown until the 'up' scrollbar is pressed.

Arkh1:
+1 to this.  I am seeing the same problem with infinite scroll.  I also see a similar problem without infinite scroll on a grid with a large data set.

rkilarski:
I just noticed this as well, and looks like this was broken between 1/2 and 1/15.

c0bra:
@Arkh1 is it still breaking for you on the large data set without infinite scrolling? I couldn't get it to happen.

Arkh1:
Actually after switching over to the unstable version of ui-grid I no longer experience the issues I was having with large datasets.

PaulL1:
OK, excellent, closing.

cookednoodles:
This issue is still reproducable, I can see it on the demos and when moving over to unstable on my project. I have reproduced it on both chrome and firefox, all you need to do is go to http://ui-grid.info/docs/#/tutorial/212_infinite_scroll and scroll down via scrollbar or the mouse.

![bug](https://cloud.githubusercontent.com/assets/226331/5973799/c833273c-a86c-11e4-9515-0b5c13c9aa8a.png)

When this happens, the canvas has a height of 6000px, the first element has a margin-top of 4350px in the style. If I remove the margin on both, the scroll bar is very jumpy and initially at the wrong postion, but I can see content 'normally' for a while until it loads more content and the height changes again.

rkilarski:
My group actually has a fix for this issue, we are working through it and will post a fix asap.  

We're also working on code to enable infinite scroll in the UP direction.  Just FYI.

chussenet:
Cool  Ryszard !

Yes ,currently the infinite scrolling is not working . Should be taking out of the feature implemented until a basic sample works such as the sample from the tutorial !

Not working or missing features .
1-blank block after scrolling until a request is triggered to the backend to load more records.
2-keeping in memory only a subset of the records loaded to minimize footprint  
3-scrolling up and down should trigger a call to the backend if the records previously loaded were reclaimed for memory management.
4-currently ,the implementation is pushing records to the end of the array . May need more sophisticated to push ,insert in specific positions. (see slickgrid for example)

Thank you
Claude Hussenet

cookednoodles:
[...]

PR #2730, opened 2015-02-09
Enable Infinite Scroll Up 

mloutraris-ge:
Additionally fix display issue for infinite scroll 'blank block' #2535

Code changes enable infinite scroll up. Changes are as follows:

misc/tutorial/212_infinite_scroll.ngdoc
Update documentation to load the data for up and adjust the load data
for down to append to exiting data, not overwrite the hashKeys

src/features/infinite-scroll/js/infinite-scroll.js
Updated file to add new needLoadMoreDataTop method and supporting code
to call method.

src/features/infinite-scroll/test/infiniteScroll.spec.js
Updated unit test to test for needMoreLoadDataTop

src/js/core/constants.js
Added new constant for scroll direction

src/js/core/directives/ui-grid-viewport.js
Added supporting logic to set scroll direction properly

src/js/core/directives/ui-grid.js
Added new function to adjust scroll position after an infinite scroll
data load up or down.

src/js/core/factories/Grid.js
Added logic to reset scroll direction to none after a debounce as well
as increased debounce to 1000 to allow for timeout added to reset scroll
position to execute

src/js/core/factories/GridRenderContainer.js
Added logic too datawatch function to calculate the row to scroll to
after a data load. The up scroll is tricky because you ahve to find the
row you were on in the new data array. This was done by searching for
the hashKey of the row we were on. Supporting code was also added. The
"blank block" issue was fixed here by allowing the logic to execute if
infinite scroll is enabled.

### DIFF
--- a/misc/tutorial/212_infinite_scroll.ngdoc
+++ b/misc/tutorial/212_infinite_scroll.ngdoc
@@ -32,10 +32,20 @@ Specify percentage when lazy load should trigger:
         { name:'name' },
         { name:'age' }
       ];
-      var page = 1;
+      var page = 0;
+      var pageUp = 0;
       var getData = function(data, page) {
         var res = [];
-        for (var i = 0; i < page * 100 && i < data.length; ++i) {
+        for (var i = (page * 100); i < (page + 1) * 100 && i < data.length; ++i) {
+          res.push(data[i]);
+        }
+        return res;
+      };
+
+      var getDataUp = function(data, page) {
+        var res = [];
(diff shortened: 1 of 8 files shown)


PaulL1:
This code looks clean and well documented, and has unit tests, thanks for that.

I can see it fiddles with the scrolling a little bit, @swalters you may want to cast an eye over that, it has a specific if clause that checks for infinite scroll being enabled, but I'm aware that we had problems with blank rows in other situations than just infinite scroll - maybe that code should be generalised?

Given the code is localised to infinite scroll - merging.

PaulL1:
Note that a chunk of this code has been backed out so as to remove the infiniteScroll logic that had crept into core.  However, the concepts and intent I think have been retained.

PR #2751, opened 2015-02-11
Fix (cellNav): don't clear focus on scroll

PaulL1:


adilbenmoussa:
Thanx for the quick reaction and the fix!

PR #2756, opened 2015-02-12
i18n: add pinning and pagination for German

laliluna:


PaulL1:
Excellent, thank you.

Issue #1926, opened 2014-10-27
When the mouse is on the grid, the container can't scroll


calebkiage:
Hello, I have disable grid scrolling but when the mouse pointer is in the grid, the container cannot scroll. Any ideas on how I can disable this behavior?

PaulL1:
I believe this is now resolved - could people verify?

calebkiage:
It seems that this has been partially fixed. When scrolling up, the page continues scrolling when the grid's scroller reaches the top which is good. But the same doesn't happen when the grid's scroller is at the bottom. See [102 Sorting](http://ui-grid.info/docs/#/tutorial/102_sorting)
Also, I have noticed that grids don't have enough items to scroll still give issues with scrolling the page. See [101 Intro to UI-Grid](http://ui-grid.info/docs/#/tutorial/101_intro)

PaulL1:
Interesting.  Grids with little data such as the 101 tut allow the page to scroll downwards, but not upwards.  Grids with more data such as 102 sorting allow the page to scroll upwards (when reaching the end of the scroll), but not downwards.

PaulL1:
Thinking further on this, it could be the case that the grid decides whether to scroll itself or propagate to the parent based on how far through the grid scroll it currently is.  If it's 100% and you scroll down, it propagates to the page.  If you're at 0% and you scroll up, it propagates to the page.

So, when you have a grid with less than a page of data, it's always at 100% scroll.  So it won't propagate the scroll up to the page.  And currently we have a miscalculation on the vertical scroll so that when you reach the end of the data, you're not at 100% scroll.  So when you have more than a page of data, currently you maybe cannot get to 100% so that it propagates the scroll down to the parent.

A theory at this stage, no more.

joshmackey:
I'm also getting this problem. I've set the grid to be as tall as needed. We don't want the grid to be scrolling, only the page as a whole. However, whenever the mouse is over the grid, it disabled the whole page scrolling even if the grid itself doesn't scroll any.

PaulL1:
OK, my theory on it not getting to 100% scroll is incorrect, it's a rendering glitch rather than a logic glitch.  That is to say, the scroll bar on my Mac doesn't go all the way to the bottom, but if you look closely it's going all the way to the end of the scroll - perhaps it's leaving space for the horizontal to slide in underneath it?

[...]

PR #2762, opened 2015-02-12
scroll bug when scrolling down  #1926

foxx9:
>When scrolling up, the page continues scrolling when the grid's scroller reaches the top which is good. But the same doesn't happen when the grid's scroller is at the bottom

It seems that the fireThrottledScrollingEvent function had a conflict with the browser scrolling behavior  #1926


PaulL1:
Very excellent.  I tried and failed to fix that one!!

PR #2772, opened 2015-02-13
Add date flag to filters to be allow columns to be filtered by dates.

amyboyd:
Example use:

image

Code:

    columnDefs: [
        // ...
        {
            cellFilter: 'date:\'d MMM yyyy H:mm\'',
            displayName: 'Order Date',
            field: 'date',
            filters: [
                {
                    condition: uiGridConstants.filter.GREATER_THAN_OR_EQUAL,
                    placeholder: 'From',
                    flags: { date: true }
                },
                {
                    condition: uiGridConstants.filter.LESS_THAN_OR_EQUAL,
                    placeholder: 'To',
                    flags: { date: true }
                }
            ]
        }
        // ....
    ]

### DIFF
--- a/src/js/core/services/rowSearcher.js
+++ b/src/js/core/services/rowSearcher.js
@@ -134,7 +134,7 @@ module.service('rowSearcher', ['gridUtil', 'uiGridConstants', function (gridUtil
           newFilter.condition = rowSearcher.guessCondition(filter);
         }
 
-        newFilter.flags = angular.extend( { caseSensitive: false }, filter.flags );
+        newFilter.flags = angular.extend( { caseSensitive: false, date: false }, filter.flags );
 
         if (newFilter.condition === uiGridConstants.filter.STARTS_WITH) {
           newFilter.startswithRE = new RegExp('^' + newFilter.term, regexpFlags);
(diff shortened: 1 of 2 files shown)


sunz7:
Nice feature! Thank you

PaulL1:
Looks excellent.  Merging.

src/js/core/services/rowSearcher.js
```diff
@@ -221,6 +221,12 @@ module.service('rowSearcher', ['gridUtil', 'uiGridConstants', function (gridUtil
       }
     }
 
+    if (filter.flags.date === true) {
+      value = new Date(value);
+      // If the term has a dash in it, it comes through as '\-' -- we need to take out the '\'.
+      term = new Date(term.replace(/\\/g, ''));
```
SSchaaf: This filter is runnig ok. But will be better if we can use with formts like dd/mm/yyyy, mm/dd/yyy, etc..

In my case this filter obly works with date format yyyy-m-d (image). new Date parse this kind of format but not the others above.

image


Why not include moment.js in this project to support more formats ?

I think that we can use cellFilter (that is angular filter used to format the column date data) with moment.js ? 

I was playing with it and I think that could be a good solution.

This is my code :

```
if (column.cellFilter) {
	term = moment(term.replace(/\\/g, ''), column.cellFilter.split(':')[1].toUpperCase()).toDate();
} else {
	term = new Date(term.replace(/\\/g, ''));
}
```

Issue #2755, opened 2015-02-12
Cell navigation in combination with cell Edit is broken


pedroclayman:
see http://ui-grid.info/docs/#/tutorial/205_row_editable

If you navigate to a cell that lies somewhere outside of the first page (scrolling occurs) and hit F2 then the cell does not become editable. Some weird scrolling happens instead.

PaulL1:
This looks to have been fixed in #2790, could you retest?

pedroclayman:
Fix works. Problem solved.

PaulL1:
Thanks,

PR #2790, opened 2015-02-17
Fixing editOnFocus when the view is scrolled

### LINKED ISSUE #2750: enableCellEditOnFocus for date objects cells has a weird behavior
Hi Guys,

We are newbies on the ui-grid, and we are facing an issue with inline editing of the a date object cell. please check this plnkr: http://plnkr.co/edit/Zh8dh6CBifaCwFHxct63?p=preview

In my application i'm getting the behavior when the cell is edit mode and the user tries to scroll.

Tested on:
Chrome: Version 40.0.2214.111 (64-bit)
Safari: Version 8.0.3 (10600.3.18)

Thanx in advance!
Kind regards,

axelstudios:
This should fix #2750 and #2755.  There are separate issues related to `scrollToIfNecessary`'s boundary values, but I'll make a separate PR with those fixes

### DIFF
--- a/src/features/cellnav/js/cellnav.js
+++ b/src/features/cellnav/js/cellnav.js
@@ -306,7 +306,7 @@
 
                 /**
                  * @ngdoc function
-                 * @name scrollToFocus
+                 * @name scrollToIfNecessary
                  * @methodOf  ui.grid.cellNav.api:PublicApi
                  * @description brings the specified row and column fully into view if it isn't already
                  * @param {GridRow} row grid row that we should make fully visible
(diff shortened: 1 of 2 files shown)


axelstudios:
On a somewhat related note, it looks like the `$scope` argument should also be removed from [these two calls](https://github.com/angular-ui/ng-grid/blob/933124a6a70eb95f4878ed67daa8381f598b64b1/src/features/saveState/js/saveState.js#L487-L489).

PaulL1:
Some of this was handled in #2755 I think, I need to look and see what is left that needs dealing with.

PaulL1:
Both of the related issues are closed.  Has this been handled through the other changes that have been made?

PaulL1:
Closing, both because it may have been fixed in the related issues, and because I think @swalters changes to scrolling have made this redundant.

PR #2800, opened 2015-02-18
Fix dynamic header heights

c0bra:
Previously header cell heights were calculated based on a "max height is
king" algorithm where cells could grow to match the largest height, but
could never shrink, e.g. this let cells with no filter box be the same
height as those with no filter box.

This presented a problem with toggling features like filtering. If
suddenly no columns were filterable this would leave a big empty space at
the bottom of each header cell.

This change makes header cells display-type "table-cell", which lets them
render like a TD with full height, and places them in table-row and
table divs. Replacing "width" with "min-width" and "max-width" forces the
header cells to be the correct width.

Also the `columnStyle` method has been replaced with a
`columnWrapperStyle` method.

### DIFF
--- a/src/js/core/directives/ui-grid-render-container.js
+++ b/src/js/core/directives/ui-grid-render-container.js
@@ -100,7 +100,7 @@
                 var newScrollLeft = args.getNewScrollLeft(colContainer,containerCtrl.viewport);
 
                 // Make the current horizontal scroll position available in the $scope
-                $scope.newScrollLeft = newScrollLeft;                
+                $scope.newScrollLeft = newScrollLeft;
 
                 if (containerCtrl.headerViewport) {
                   containerCtrl.headerViewport.scrollLeft = gridUtil.denormalizeScrollLeft(containerCtrl.headerViewport, newScrollLeft);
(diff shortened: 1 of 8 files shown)


lecoleco:
Hi, I'm set enableRowHeaderSelection:true and makin filter dynamic with ng-select. (ui grid last version)

For make that, the only way i found is remove the column without the filter and add the column again now with the filter.

If the option enableRowHeaderSelection is true, the header heights not working good,
show the filter when add and hide the input filter when remove but not change height any more.

I change this point to get the lowest and work's good. 
I have another way to show and hide the filter, please tell me.
Thanks 

                 if (container.headerHeight < maxHeaderHeight) {
                    container.explicitHeaderHeight = container.headerHeight;
               }
                if (container.headerCanvasHeight < maxHeaderCanvasHeight) {
                      container.explicitHeaderCanvasHeight = container.headerCanvasHeight;
                }

cgoss:
I believe this may have caused a bug.

I have just started to work with this control and I am at 101 in the tutorial and I get a header row that is the entire size of the viewport with a row the size of the horizontal scroll bar. 

This code renderContainer.explicitHeaderCanvasHeight is undefined and the inherited height is that of the parent container element.


  if (renderContainer.explicitHeaderCanvasHeight) {
                ret += '\n .grid' + uiGridCtrl.grid.id + ' .ui-grid-render-container-' + $scope.containerId + ' .ui-grid-header-canvas { height: ' + renderContainer.explicitHeaderCanvasHeight + 'px; }';
              }
              else {
                ret += '\n .grid' + uiGridCtrl.grid.id + ' .ui-grid-render-container-' + $scope.containerId + ' .ui-grid-header-canvas { height: inherit; }';
              }

PR #2806, opened 2015-02-18
Fix Bug, Change movingColumn to follow cursor

ChrisSchmidt1991:
MovingColumn position stays centered on the cursor unless outside
bounaries. Bug was from cursor moving up/down across the header row that pushed
the movingColumn away from the cursor.

### DIFF
--- a/src/features/move-columns/js/column-movable.js
+++ b/src/features/move-columns/js/column-movable.js
@@ -287,7 +287,7 @@
                         movingElm.css(movingElementStyles);
                       };
 
-                      var moveElement = function (changeValue) {
+                      var moveElement = function (changeValue, cursorLocation) {
                         //Hide column menu
                         uiGridCtrl.fireEvent('hide-menu');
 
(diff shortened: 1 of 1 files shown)


PaulL1:
@jpuri, does this make sense to you, and appear to work reasonably well?

jpuri:
@PaulL1 , sorry for the delay. I will check this today and tell you.

ChrisSchmidt1991:
Another way to reproduce this issue is drag the column till the cursor is outside and to the left/right of the grid. then move cursor back into grid. There will be a gap between the cursor and the moving column. http://ui-grid.info/docs/#/tutorial/310_column_moving

PaulL1:
Looks like this bug still exists, even after the recent changes to column moving.  I think we need the fix, but someone needs to rebase the PR - it no longer automatically merges.

swalters:
@ChrisSchmidt1991 Closing this PR as @PaulL1 is looking to refactor moving columns in a few days.

Issue #2811, opened 2015-02-19
infinite scroll doesn't work well with rowIdentity


lvostinar:
the commit from https://github.com/angular-ui/ng-grid/issues/2535 should use rowIdentity instead of $$hashKey

c0bra:
What does "doesn't work well" mean?

c0bra:
You'll want to squash these 3 commits into one.

lvostinar:
Is only one commit on different branches but we can close this one, I created a pull request.

PR #2812, opened 2015-02-19
fix for #2811

lvostinar:
Fix infinite scrolling when rowIdentity is used.

### DIFF
--- a/src/js/core/factories/GridRenderContainer.js
+++ b/src/js/core/factories/GridRenderContainer.js
@@ -382,7 +382,7 @@ angular.module('ui.grid')
         if ( self.grid.scrollDirection === uiGridConstants.scrollDirection.UP ) {
           findIndex = rowIndex > 0 ? self.grid.options.excessRows : 0;
           for ( i = 0; i < rowCache.length; i++) {
-        	 if (self.grid.options.rowIdentity(rowCache[i].entity) === self.grid.options.rowIdentity(self.renderedRows[findIndex].entity)) {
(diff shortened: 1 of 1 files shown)


c0bra:
This commit fails to pass the style check, please address that. Also, please put text in your pull request description.

lvostinar:
Done.

c0bra:
@lvostinar I'll check this out locally and get it merged. Do I just need to use infinite scroll with my own custom rowIdentity to cause the problem to occur? What problem/error should I expect to happen? I would like to get a unit test written for it if I can.

lvostinar:
I was getting a lot of flickering because rowIndex was always set to 0.

jcompagner:
yes we have our own rowIdent function that returns a unique id that we already know like this:

					rowIdentity: function(o) {
						return o._svyRowId;
					},
					rowEquality: function(row1,row2) {
						return row1._svyRowId == row2._svyRowId; 
					}

so our rows don't really have a $$hashKey because that is not really generated. we provide our own.
But that code doesn't use our rowIdentity function that we provide but depends on that hashKey thats why it never hit, and the range that we try to calculate there was never done correctly

jcompagner:
It is not fully clear to me how getRowIdentity and the rowIdentty really work
We implement the rowIdentity (that use used a lot for example in the RowHashMap class/object)

But getRowIdentity() is only use in  Grid.prototype.modifyRows function
We don't hit that code because for us self.options.useExternalSorting is true so it always goes into the first if of that modifyRows function.
Thats why for use the $$hashKey is really never generated (and i guess if we didn't have externalSorting enabled we also had to provide also the getRowIdentity which then does the same as rowIdentity?

c0bra:
I'm just going to go ahead and merge this.

PR #2828, opened 2015-02-23
Bug fixes for 509e007 and visibility change

cookednoodles:


c0bra:
@cookednoodles Looks like your changes are failing to past jshint or the jscs style check. You can run `grunt test:single --angular=1.3.6` locally to try and find the issues.

cookednoodles:
This looks like these would have been erroring on the old code too, could you confirm these look like FPs?

PR #2837, opened 2015-02-24
No data element: Displayed when no rows are being rendered

pedroclayman:
The template 'ui-grid/uiGridNoData' is being displayed when no rows are being rendered. The default template can be overridden using the property 'noDataTemplate' in options.


pedroclayman:
I forgot to implement some attribute to turn this feature off (enableNoData). I am planning to fix this in the following days

c0bra:
@pedroclayman there's a method in gridUtil named `getTemplate()` which will handle the various types of template properties people will put in.

swalters:
@pedroclayman Thanks for the PR.  Please look at @c0bra's last comment and submit a new PR that we can merge.  Thanks.

PR #2845, opened 2015-02-25
Bring back the 2.x menu behavior of custom menu templates.

farooqu:
In 2.x it was possible to specify a custom menu template. In 3.0 the one grid menu has been split into 3 directives and custom templates are not possible. This pull request adds the ability to specify custom templates for menuButton, menu, and menuItem.


PaulL1:
OK, this code looks clean and well documented.  I'd like to see unit tests, but I know that the unit tests in general in this area aren't in great shape.  It looks like a lot of code change but most of that is just changing from link to pre/post, and associated indenting changes, this mostly looks to just be including templates as it says.  

Thanks for the contribution, merging.

PR #2873, opened 2015-02-28
Fix (exporter): Fix #2313 export for IE9

### LINKED ISSUE #2313: grouping ui-grid ?
Currently 3.0 ui-grid non-functional grouping and it is a very important function. Can you tell me when the test is functional grouping.

PaulL1:
Information and code provided by @alexandery.  I have no ability to test
it, but it should only impact IE9, which apparently doesn't work otherwise
anyway.

The main concern here is that this creates and destroys an iFrame in IE9,
and that iFrame apparently messes up formatting, so requires application of
width: auto !important; on the ui-grid-viewport.  That appears to break formatting
in chrome on all features tutorial, so I haven't applied it.  Presumably that means
formatting will be ugly in IE9.

### DIFF
--- a/src/features/exporter/js/exporter.js
+++ b/src/features/exporter/js/exporter.js
@@ -734,6 +734,18 @@
 
         /**
          * @ngdoc function
+         * @name isIE
+         * @methodOf  ui.grid.exporter.service:uiGridExporterService
+         * @description Checks whether current browser is IE and returns it's version if it is
+        */
+        isIE: function () {
+            var myNav = navigator.userAgent.toLowerCase();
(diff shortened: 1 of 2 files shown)


alexandery:
@PaulL1 I am surprised that it breaks something in Chrome - I believe we have tested it in Chrome and FF and it was working fine. But then again - that was unstable RC18, so you guys might have changed the CSS since then.

Just an FYI - the formatting was getting messed up only then displaying row header checkboxes, which could have been touched as well. All in all - the CSS is not critical, I believe.

I will be able to test it only on Monday, so if the latest code makes it to Tutorial pages, I will try it out there in IE9.

P.S. Not sure if this is the best spot to communicate on the issue.

Issue #2312, opened 2014-12-09
IE9 - no way to save the results of CSV export 


alexandery:
Clicking the link that gets generated by the grid results in navigation to an empty page in IE9 - no way to save the file even from the context menu "Save Target As...". Firefox prompts you to save the file.

alexandery:
@PaulL1,

Thanks for the update. I've tried the latest unstable version today (v3.0.0-RC.18-f8a68cc). Unfortunately the export functionality doesn't seem to work for IE9 still. Upon exporting IE9 pops up a new, completely empty window.

The code generates CSV data, but when it gets to 'downloadFile' method, the execution gets to lines 14081-14082 and continues from there.

Would appreciate any pointers or suggestions on how to go about testing and figuring out what's wrong there.

Thanks!

PaulL1:
@alexandery: what is at lines 14081-14082?  I did some changes on downloadFile a few days ago, but I doubt they fixed it.  If you can tell me what line it is in that the error occurs, I can have a look.

alexandery:
@PaulL1 Well, I thought you could check the file to get the content of the lines. However, I have implemented the feature myself. I hope you can integrate this into the component to make sure that others can use it (and that I won't need to apply these changes next time I upgrade).

Below is the code to implement it on IE9 (possibly below as well). All of it goes into ui.grid.exporter. 

1. Main csvExport() method has been fitted with 'filename' parameter to allow the calling code to specify the filename:

```javascript
        csvExport: function (grid, rowTypes, colTypes, filename) {
          var exportColumnHeaders = this.getColumnHeaders(grid, colTypes);
          var exportData = this.getData(grid, rowTypes, colTypes);
          var csvContent = this.formatAsCsv(exportColumnHeaders, exportData, grid.options.exporterCsvColumnSeparator);
          
          this.downloadFile (grid.options.exporterCsvFilename, csvContent, filename);
        }
```

2. New method to check for IE version

```javascript
      	/**
         * @ngdoc function
         * @name isIE
         * @methodOf  ui.grid.exporter.service:uiGridExporterService
         * @description Checks whether current browser is IE and returns it's version if it is
		*/
        isIE: function () {
    		var myNav = navigator.userAgent.toLowerCase();
    		return (myNav.indexOf('msie') != -1) ? parseInt(myNav.split('msie')[1]) : false;
		}
```

3. Modified downloadFile() method
```javascript
        downloadFile: function (fileName, csvContent, filename) {
[...]

PR #2878, opened 2015-03-02
Fix export bug around isIE

tanak:
isIE() returns false when using browsers other than Internet Explorer,
but it worked as if using IE.

I tested in IE 11, Firefox 36.0, and Chrome 40.0.2214.115 m.

This pull request is related with #2312 and #2873.

### DIFF
--- a/src/features/exporter/js/exporter.js
+++ b/src/features/exporter/js/exporter.js
@@ -760,6 +760,7 @@
           var a = D.createElement('a');
           var strMimeType = 'application/octet-stream;charset=utf-8';
           var rawFile;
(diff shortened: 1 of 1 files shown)


PaulL1:
Thanks for this fix @tanak, very useful.  Sorry for the breakage, I didn't pick it up when reviewing @alexandery's code fragment.

alexandery:
Sorry about that guys. I've posted a reply to: https://github.com/angular-ui/ng-grid/issues/2312 but am seeing now that @tanak has actually fixed the issue in the source code already.

PR #2891, opened 2015-03-03
Fix left scroll typo from 509e007

cookednoodles:
I have split up my PRs from 2828 to make things easier. This fixes the left scroll when adding cols.


PaulL1:
I think I merged an identical change under #2914.  Even though yours was provided first.  Sorry.  :-)

PR #2892, opened 2015-03-03
Fix the condition that triggers the sort to change

cookednoodles:
Fix the condition that triggers the sort to change, it was firing incorrectly on every col


PaulL1:
Looks good, thanks.

PR #2893, opened 2015-03-03
Fix #2838 - Do selection on device appropriate events, not both

### LINKED ISSUE #2838: Touch event row selection bug
When selecting a row with UI-Grid in touch mode the registerRowSelectionEvents function is triggering both the touchstart/end and the click events. 

This means that the selectCells function is being called twice, the first time it selects the row, the second time it unselects it.

cookednoodles:
At the moment, the selection/deselection is happening twice on touch devices as both the click and touch events are being registered.
This will only use the touch events when touch is supported, and click when it isn't.

### DIFF
--- a/src/features/selection/js/selection.js
+++ b/src/features/selection/js/selection.js
@@ -795,9 +795,13 @@
             function registerRowSelectionEvents() {
               if ($scope.grid.options.enableRowSelection && !$scope.grid.options.enableRowHeaderSelection) {
                 $elm.addClass('ui-grid-disable-selection');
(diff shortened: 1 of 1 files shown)


PaulL1:
We're having issues with touch enabled devices.  A lot of Windows laptops are touch enabled, but people use them with a mouse rather than touching the screen.  Disabling the click event means those laptops won't work with the mouse I think.

PaulL1:
Closing this because I think this approach doesn't work, there is a different option we're trying at #3018 if you wanted to try that out.

Issue #1195, opened 2014-05-19
Firefox editableCellTemplate Issue


zlaja1983:
In Firefox, if the editableCellTemplate is of:

\<input type="number" ng-class="\'colt\' + col.index" ng-input="COL_FIELD" ng-model="COL_FIELD" />

The first time a user clicks, it allows him to modify an entry. Any subsequent attempts, it does not enter the edit mode. We are currently using: version 2.0.7 of the grid. It is worth noting that it works in Chrome and IE without an issue.

There is an error in the Firefox console once trying to edit the same cell:
TypeError: current is null angular.js:9186

Here is a jsfiddle to illustrate the problem in Firefox: Try modifying the 'age' column.

http://jsfiddle.net/zlaja1983/38hPb/

EvanWinstanley:
I know this is an old issue so you guys might have moved on, but I was able to resolve this issue with a simple `$timeout` (see pull request above).

Basically for reasons I couldn't quite figure out, Firefox was firing a blur and then a focus when editing a cell with `<input type=number />`. Adding the timeout resolved the issue. I tested for `enableCellEditOnFocus=true|false` (single click and double click editing) in FF and Chrome and everything seemed to still be working correctly in both browsers.

Hope this helps someone.

alainbryden:
Thank you Evan. This has continued to be an issue for us so we'll be trying your fix.

swalters:
Closing old 2.x issue as all current development is on 3x. 

If someone really needs this fix for 2.x, my advice is to fork and fix it.  Sorry, I know it's not ideal, but it is the best we can offer with limited resources.

PR #2902, opened 2015-03-04
Bug fix #1408 and #1195

### LINKED ISSUE #1408: FireFox editableCellTemplate throws error when the cell is edited second time
I have a cell template of type number, It works normally when I click and edit first time. The second time on same cell throws (TypeError: current is null          angular.js:12923).  However same code works on Chrome and (even on) IE and older version of FireFox (v20).

There is another user who faced a same issue and his jsfiddle: http://jsfiddle.net/zlaja1983/38hPb/
http://stackoverflow.com/questions/23939184/firefox-editablecelltemplate-only-enabling-editing-one-time

AngularJS v1.2.21
ng-grid-2.0.12
jQuery v2.1.1
firefox v30

EvanWinstanley:
Wrap blur in a timeout which resolves an issue in Firefox where blur is triggered on focus for input type=number

[Review on Reviewable](https://reviewable.io/reviews/angular-ui/ng-grid/2902)

### DIFF
--- a/src/directives/ng-input.js
+++ b/src/directives/ng-input.js
@@ -1,4 +1,4 @@
-ngGridDirectives.directive('ngInput', [function() {
+ngGridDirectives.directive('ngInput', ['$timeout', function($timeout) {
     return {
         require: 'ngModel',
         link: function (scope, elm, attrs, ngModel) {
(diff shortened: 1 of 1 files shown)


PaulL1:
It looks like the unit tests are failing on this.

EvanWinstanley:
Fair enough, I had a lot of trouble actually running the unit tests locally.I'll have another look at it when I get a chance.

Issue #2890, opened 2015-03-03
Bug GridRenderContainer.prototype.adjustScrollVertical


jmmartinez84:
This code:
    <pre>if (typeof(scrollTop) === 'undefined' || scrollTop === undefined || scrollTop === null) {
      scrollTop = (this.getCanvasHeight() - <strong>this.getCanvasWidth())</strong> * scrollPercentage;
    }</pre>
Should be:
     <pre> if (typeof(scrollTop) === 'undefined' || scrollTop === undefined || scrollTop === null) {
      scrollTop = (this.getCanvasHeight() - <strong>this.getViewportHeight())</strong> * scrollPercentage;
    }</pre>

PaulL1:
Any chance you could do a PR for this - perhaps referring to the first timers guide if you need to?  Changes are much more likely to get merged if submitted as a PR.

PaulL1:
Closing - the changes were merged.

Issue #2889, opened 2015-03-03
Bug:  Grid.prototype.redrawInPlace


jmmartinez84:
This issue hide first column when I add - remove an item.
This code:
    <pre>if (rowsAdded) {
        container.adjustRows(container.prevScrollTop, null);
        container.adjustColumns(<strong>container.prevScrollTop</strong>, null);
      }</pre>
Should be:
     <pre>if (rowsAdded) {
        container.adjustRows(container.prevScrollTop, null);
        container.adjustColumns(<strong>container.prevScrollLeft</strong>, null);
      }</pre>

PaulL1:
Any chance you could do a PR for this - perhaps referring to the first timers guide if you need to?  Changes are much more likely to get merged if submitted as a PR.

PaulL1:
I've merged the suggested code, thanks and closing.

PR #2909, opened 2015-03-04
Issues #2889 #2890

jmmartinez84:
https://github.com/angular-ui/ng-grid/issues/2889
https://github.com/angular-ui/ng-grid/issues/2890


c0bra:
Not sure what's going on here but the PR looks like a mess.

PR #2915, opened 2015-03-05
Update GridRenderContainer.js

jmmartinez84:
https://github.com/angular-ui/ng-grid/issues/2890

### DIFF
--- a/src/js/core/factories/GridRenderContainer.js
+++ b/src/js/core/factories/GridRenderContainer.js
@@ -311,7 +311,7 @@ angular.module('ui.grid')
     }
 
     if (typeof(scrollTop) === 'undefined' || scrollTop === undefined || scrollTop === null) {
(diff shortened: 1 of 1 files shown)


PaulL1:
OK, happy with the PR.  But I don't know what it actually does.  @jmmartinez84, could you explain the defect we're fixing, and @swalters, are you OK with this change?

jmmartinez84:
There is no easy way to call that function externally, with a scrollTop undefined value, because that function is only called in the "scroll" event, anyway, that scrollTop value is not well calculated. It's not a big deal, but it's a bug.

<pre>
angular.element('.ui-grid-viewport').isolateScope().rowContainer.adjustScrollVertical(null,0.1,true)
</pre>

c0bra:
Confirmed that this would be a bug, though I haven't seen it having any weird affects...

Issue #2796, opened 2015-02-18
Custom filter ui


yccteam:
I could not find how to implement custom filter ui (i.e. a multiselect, a drop down or a time/date/datetime ui).
I was thinking of something in the form of:
'''filter: {
          condition: function myLogic(searchTerm, cellValue){
                    // logic comes here
          },
          placeholder: 'ends with',
          ui: uiGridMultiSelect
}'''
The example above would create a multi select comprised of the values available in this column to filter with.
Any way to this right now?

PaulL1:
This is a feature that we don't currently have, but I think it would be useful, so I'll mark it for future implementation.

PaulL1:
I've added a dropdown.  I didn't find an easy way to add a multi-select dropdown to the core, because we're using native controls and don't want to introduce dependencies on external libraries.  I'm still thinking on a way to generalise this to allow other controls to be injected.

shrihari1857:
has the issue of displaying Angularjs bootstrap datepicker on headerCellTemplate on ng-Grid been resolved?

PaulL1:
I've provided the ability to have custom filter templates.

shrihari1857:
I wanted to have a bootstrap datepicker on headerCellTemplate, got it resolved a few days back by adding overflow: visible to the ngHeaderContainer class, as shown below.
.ngHeaderContainer {
  position: relative;
  overflow: visible !important;
  font-weight: bold;
  background-color: inherit;
}

Issue #2886, opened 2015-03-03
Custom filter in headerCellTemplate


pszawlis:
I created a grid with custom external filter. It works perfectly fine, as long as it is not part of the headerCellTemplate. When it is inside:

1. Data is filtered correctly but the display is not adjusted
2. I didn't manage to apply the same formatting as for Grid Mene, i.e. the drop down overlapping the header and grid cells and not changing header height. 

Please take a look at the Plunker http://plnkr.co/edit/z2Ua5CAypvjSNPb9Psjv?p=preview

thomas-koch-comsolit:
this might be a duplicate of #2796

pszawlis:
Issue #2796 is a feature request, my case (point 1.) is probably a bug (or hopefully my wrong understanding).

The bottom line is: I was able to create filter using AngularjJs mulit select and it works well as long as it is not part of  headerCellTemplate.

PaulL1:
Changes were done on the header height recently that may have fixed point 1.

igwejk:
This is a duplicate of [#2731](https://github.com/angular-ui/ng-grid/issues/2731).

@pszawlis see my comment there to understand how you can achieve the behaviour you want.

PaulL1:
I think the new filterHeaderTemplate option resolves this: http://ui-grid.info/docs/#/tutorial/306_custom_filters

PR #2918, opened 2015-03-05
introduce directive property for column filter definition

thomas-koch-comsolit:
This should solve #2796 by adding a 'directive:' property to the filter definition, e.g.:
```javascript
columnDefs: [{
  name: 'eventData',
  filter: {
    directive: 'my-own-datepicker'
  }}]
```

If directive is defined than this directive is used to render the filter instead of the default text input field.

So we propose:
a) to add the above directive configuration property
b) to do a 'deep watch' on the 'term' property and allowing arbitrary JavaScript objects on the the 'term' property. This of course only makes sense with external filtering.
c) to allow custom filtering directives to add data to the filter definition. In our case we would like to add a 'type' property that is than also send to the server for server side filtering. Thus the server code nows what filter directive was used and how to interpret the term property.
d) to share the scope with the custom filter directive to make c) possible and it didn't work without that for us.

This pull request is only meant as a proof-of-concept and we'd be glad to also provide documentation and tests if you accept the idea.

Thank you.

### DIFF
--- a/src/js/core/directives/ui-grid-filter.js
+++ b/src/js/core/directives/ui-grid-filter.js
@@ -1,0 +1,18 @@
+(function(){
+  'use strict';
+
+  angular.module('ui.grid').directive('uiGridFilter', function ($compile, $templateCache) {
+
+    return {
+      compile: function() {
+        return {
+          pre: function ($scope, $elm, $attrs) {
+            var template = $scope.colFilter.filterTemplate || $scope.col.filterTemplate;
+
+            $elm.append($compile(template)($scope));
+          }
+        };
+      }
+    };
+  });
+})();
--- a/src/js/core/services/gridClassFactory.js
+++ b/src/js/core/services/gridClassFactory.js
@@ -160,12 +160,16 @@
           }
 
           var
-            templates = ['HeaderCell', 'Cell', 'FooterCell'],
+            templates = ['HeaderCell', 'Cell', 'FooterCell', 'Filter'],
             templateGetPromises = [];
 
           for (var i = 0; i < templates.length; ++i) {
             preprocessTemplate(templates[i], colDef, col);
           }
+
+          angular.forEach(colDef.filters, function(filter, i) {
+            preprocessTemplate('Filter', filter, col.filters[i]);
+          });
 
           // Create a promise for the compiled element function
           col.compiledElementFnDefer = $q.defer();
--- a/src/js/core/services/gridClassFactory.js
+++ b/src/js/core/services/gridClassFactory.js
@@ -101,9 +101,6 @@
          * @param {object} gridOptions reference to grid options
          */
         defaultColumnBuilder: function (colDef, col, gridOptions) {
-
-          var templateGetPromises = [];
-
           /**
            * @ngdoc property
            * @name headerCellTemplate
@@ -112,11 +109,6 @@
            * is ui-grid/uiGridHeaderCell
            *
            */
-          if (!colDef.headerCellTemplate) {
-            col.providedHeaderCellTemplate = 'ui-grid/uiGridHeaderCell';
-          } else {
-            col.providedHeaderCellTemplate = colDef.headerCellTemplate;
-          }
 
           /**
            * @ngdoc property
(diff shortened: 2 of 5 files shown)


PaulL1:
I think I'm OK with the concept of a custom directive, although we'd need to carefully document what behaviours that directive must provide.

I'm not sure on the deep watch - I presume that's because you want to trigger filtering any time anything in the term changes?  Is that instead something we could roll into the directive - so make the directive explicitly call a method when it's ready for the filtering to run?  Filtering is reasonably expensive with large data sets, I've never been super happy with running it on every keypress.  And we generally try to avoid watches where-ever we can, I think deep watches are things we avoid more than shallow watches.  The other guys might have opinions on this too (and they may differ from mine).

We'd probably want to see unit tests with a feature like this, and a tutorial.

But thanks very much for the contribution, it's good to have more contributors committing big chunks of functionality.

swalters:
After discussing with contributors, we would rather see colDef.filterTemplate option. This should meet your needs as you can code a directive that is used in the filterTemplate and each column can have a different filterTemplate.

Can you look at approaching from that direction?  The first task would be to take the existing code and moving to a default filterTemplate.

pszawlis:
I tried to apply solution from the fork to the problem described in #2886.

Unfortunately it does not solve the issue with the display, my filter list is still hidden behind ui-grid cells. Is there any special formatting that I should apply for my custom directive?



I tried to reproduce the issue with the [plunker](http://plnkr.co/edit/PsTzjVhpxQ0lkvXvysJV?p=preview) substituting standard reference to ui-grid-unstable.js with the ui-grid.min.js built from the fork, but I keep getting this error


`"Error: [$injector:unpr] Unknown provider: aProvider <- a <- uiGridHeaderCellFilterDirective"`

PaulL1:
a.  I'd try the unminified version from the fork, then you might get a more useful error message
b.  There are a set of overflow: hidden attributes set in the grid, you basically have to override all these to get it working.  There was another issue recently with this exact same problem, I'll look for it, but I recollect three places - ui-grid-viewport, ui-grid-header and ui-grid-cell or something like that.

PaulL1:
#2932

pszawlis:
Thanks for help, changing the css in the original plunker solved the layout issue, please see amended version of the original plunker. Nevertheless, grid is still not properly redrawn after filtering.

http://plnkr.co/edit/z2Ua5CAypvjSNPb9Psjv?p=preview

Unfortunately plunker with custom filter doesn't work with unminified version. it looks like it fails with dependency injection, but I have no idea why.

` Uncaught Error: [$injector:modulerr] Failed to instantiate module mainApp due to:
Error: [$injector:nomod] Module 'mainApp' is not available! You either misspelled the module name or forgot to load it. If registering a module ensure that you specify the dependencies as the second argument.`

http://plnkr.co/edit/PsTzjVhpxQ0lkvXvysJV?p=preview

PaulL1:
That error is usually a typo in your plunker.

pszawlis:
The plunker with custom filter works now in terms of display. 

http://plnkr.co/edit/PsTzjVhpxQ0lkvXvysJV?p=preview

sachinny21:
Will this be a part of the next stable build as a stand alone feature?

pszawlis:
I found the reason of the wrong filtering. 

I made an error of referring to mainCtrl from headerCellTemplate (or filter directive in second solution). After moving to different controller and using $scope.grid.appScope everything works just fine. 

Probably using broadcast would be better here, but it works

You can check two examples 

headerCellTemplate
http://plnkr.co/edit/lcoTtIdg723yHXtwsKjl?p=preview

filter directive
http://plnkr.co/edit/PsTzjVhpxQ0lkvXvysJV?p=preview

pszawlis:
I tried to add resizable columns with this approach and unfortunately it breaks. As soon as grid is to wide to fit the container header stops adjusting. You can check it here with my mulit-selects.

http://plnkr.co/edit/lcoTtIdg723yHXtwsKjl?p=preview

PaulL1:
I've written something similar, using this code as an example, but using the concept of a template instead of a directive (semantics to some extent, but fits with our design patterns).

I've included a tutorial: http://ui-grid.info/docs/#/tutorial/306_custom_filters

Let me know if this meets your needs.

PR #2927, opened 2015-03-06
Do deep watch on filter term

ghost:
extracted this from #2918 to discuss the issue of a deep watch on filter.term here.

I forgot to to reply to @PaulL1 in the commit message: "Filtering is reasonably expensive with large data sets, I've never been super happy with running it on every keypress."

With angular 1.3 we have ngModelOptions with debounce. So filtering is not triggered with every keypress but only after the user finished typing. BTW: Why don't you raise the minimum angular version for ui-grid to 1.3?

[Review on Reviewable](https://reviewable.io/reviews/angular-ui/ng-grid/2927)

### DIFF
--- a/src/js/core/directives/ui-grid-header-cell.js
+++ b/src/js/core/directives/ui-grid-header-cell.js
@@ -239,12 +239,16 @@
             if ($scope.filterable) {
               var filterDeregisters = [];
               angular.forEach($scope.col.filters, function(filter, i) {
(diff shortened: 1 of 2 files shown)


ageblade:
Any progress on this? maybe it could be configured in the filter object so it will be disabled by default.

najafsen:
+1

mportuga:
It appears this PR has not received any updates or responses in over a year, and the issue in question is closed. I am closing it for now. @najafsen and @ageblade Feel free to open your own PRs if you so desire.

Issue #1818, opened 2014-10-14
3.0: Document the grid initialisation sequence for new contributors


PaulL1:
We need to document the sequence that the grid runs through to initialise, including when to use each of grid.refresh, grid.refreshCanvas, grid.refreshRows, and what some of the core methods like modifyRows and setVisibleRows need to do.

As the usage of the grid ramps up we'll both have more bugs and probably more contributors, having this in place would make it much easier for people to come up to speed with the project.

PaulL1:
Completed - refer http://ui-grid.info/docs/#/api/design-rendering-cycle

thatguyyoulove:
@PaulL1 This page shows up blank for me. Was it removed?

PaulL1:
The document is still here: https://github.com/angular-ui/ui-grid/blob/master/design/RENDERING_CYCLE.md

Not sure why it's stopped showing in the design docs, presumably someone has changed something in the config.

PR #2940, opened 2015-03-08
Enh(core): add doco to refresh methods, move to queueGridRefresh

PaulL1:
Make queueRefresh (which calls refreshCanvas) and queueGridRefresh
(which calls grid.refresh) available on the api.  Document when
to use which, and what they do.

Change the core application to use queueGridRefresh in preference
to refresh, which means we get a debounce if this was being called
multiple times in a digest - for example once for each column.

Didn't apply to areas that appeared to be waiting on the promise
that it returns - e.g. didn't apply to pinning.


PaulL1:
Part of #1818

PR #2982, opened 2015-03-11
Adding functionality to help fix bug with row template and selection functionality.

axschech:
When adding a custom row template while using ui.grid.selection it causes the rowHeaderSelection checkmarks to stop functioning correctly.

This change will:
	- Set isRowSelectable to return true if a row template is passed in and there is no current function for isRowSelectable
	- explicity set the value of selected before it is passed into setSelected (in toggleRowSelection)


axschech:
See #2985

PR #3001, opened 2015-03-13
Fix editDropdownFilter

### LINKED ISSUE #3000: editDropdownFilter does not work
editDropdownFilter does not work because the position of filter expression is not correct.

src/features/edit/templates/dropdownEditor.html
```html
<div>
  <form name="inputForm">
    <select ng-class="'colt' + col.uid" ui-grid-edit-dropdown ng-model="MODEL_COL_FIELD" ng-options="field[editDropdownIdLabel] as field[editDropdownValueLabel] CUSTOM_FILTERS for field in editDropdownOptionsArray"></select>
  </form>
</div>
```
should be
```html
<div>
  <form name="inputForm">
    <select ng-class="'colt' + col.uid" ui-grid-edit-dropdown ng-model="MODEL_COL_FIELD" ng-options="field[editDropdownIdLabel] as field[editDropdownValueLabel] for field in editDropdownOptionsArray CUSTOM_FILTERS"></select>
  </form>
</div>
```

tanak:
change CUSTOM_FILSTERS position to enable filters
and add a test for editDropdownFilter

fixed #3000

### DIFF
--- a/src/features/edit/templates/dropdownEditor.html
+++ b/src/features/edit/templates/dropdownEditor.html
@@ -1,5 +1,5 @@
 <div>
   <form name="inputForm">
-    <select ng-class="'colt' + col.uid" ui-grid-edit-dropdown ng-model="MODEL_COL_FIELD" ng-options="field[editDropdownIdLabel] as field[editDropdownValueLabel] CUSTOM_FILTERS for field in editDropdownOptionsArray"></select>
+    <select ng-class="'colt' + col.uid" ui-grid-edit-dropdown ng-model="MODEL_COL_FIELD" ng-options="field[editDropdownIdLabel] as field[editDropdownValueLabel] for field in editDropdownOptionsArray CUSTOM_FILTERS"></select>
   </form>
 </div>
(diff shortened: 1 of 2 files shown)


PaulL1:
@tanak: I am reverting this, I should have checked more closely before I merged it.

The filter, similar to the other filters we offer (and as documented on the property itself) is intended for use for things like i18n - it is an angular filter in the sense that it transforms the value, rather than a filter that restricts the list of items in the object.

The change you have made applies the filter to the entire dropdown array, rather than to each value in the array.  That means you can't use angular-translate as the filter.

I'm not sure that the change you're making makes sense - it seems to be intended to provide a subset only of the values in the dropdown - but since the dropdown array can be set on a per-row basis, it seems to make little sense to filter it, you'd just put fewer values in the array surely?

tanak:
@PaulL1: sorry for breaking the editDropdownFilter.

What I'd like to do here was something like the following.

data is like:

|Name|State|City|
|---|---|---|
|John|San Francisco|Silicon Valley|
|Fred|Louisiana|New Orleans|

when you modify the state in row edit mode, the options of city dropdown will change according to the state.


I know you can set options to each row by using editDropdownRowEntityOptionsArrayPath,
but is it possible to apply to the dynamic change of options?

PaulL1:
I think the way to do that is to set a listener on the cellEdit event (check the edit feature for that listener), and then if the cell edited is a state, then change the corresponding city dropdown options on the entity.  Then it should just work.

tanak:
Thank you for your advice:)

PR #3014, opened 2015-03-15
Refactor(modifyRows): simplify and standardise modifyRows code

PaulL1:


PaulL1:
OK, I think the right approach is to merge and then see if anything unexpected breaks.

PR #3018, opened 2015-03-15
Resize touch events: sample only

PaulL1:
Try to address this issue by setting events for both touch and mouse, but
once a touchdown (or mousedown) event starts, setting only the associated
touch (or mouse) move and up handlers, and disabling the opposite down
handler until this event completes.  Similar approach to hammer.js.

Tests suppressed whilst we see whether this works on a touch device,
if it does then we can work out why this breaks the unit tests, and I can
then provide a proper PR (including look at other features and core, not just
resize).

I accidentally included the modifyRows commit in here too, but I don't think
that will impact testing this, so just ignore that.  The final PR won't include that
code.

### DIFF
--- a/src/features/resize-columns/js/ui-grid-column-resizer.js
+++ b/src/features/resize-columns/js/ui-grid-column-resizer.js
@@ -290,6 +290,7 @@
 
     var downEvent, upEvent, moveEvent;
 
+/*
     if (gridUtil.isTouchEnabled()) {
       downEvent = 'touchstart';
       upEvent = 'touchend';
@@ -300,6 +301,7 @@
       upEvent = 'mouseup';
       moveEvent = 'mousemove';
     }
+ */
 
     var resizer = {
       priority: 0,
@@ -374,7 +376,7 @@
         }
         
         
-        function mousemove(event, args) {
+        function moveFunction(event, args) {
           if (event.originalEvent) { event = event.originalEvent; }
           event.preventDefault();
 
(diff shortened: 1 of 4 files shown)


AgDude:
@PaulL1 I tried this branch, and here is what I found in tutorial 204:

Windows 8 touch-enabled, Chrome, using the mouse Column resize works as it should

iPad, Safari: column resize no longer works. Column resize works on the ipad in the master branch.

Windows 8 using, Chrome, using touchscreen  - column resize does not work either on this branch, or on master (when I touch and hold, chrome brings up the right-click menu). I don't consider this a problem, but if it happens on windows tablets, it might be for some.

PaulL1:
When you say right-click menu - is that the column menu (left click menu?), or is that the native browser right-click menu?  If the latter, not sure we can fix that?

AgDude:
It is the browser right click menu. I don't know if it could be fixed or not, but either way, I would consider it a separate issue.

PaulL1:
OK, I have an iPad, so I can chase that myself.  In general terms, if I fixed the iPad, then would this branch be better or worse than master from your viewpoint?

AgDude:
This works much better than master for resizing columns, yes I would say it is better. I really didn't spend much time looking at it, did you address the related the column click to  sort issue also?

PaulL1:
No, I was going to use the same approach for that though if this was working OK.

PaulL1:
OK, I have replaced this with a newer and better pull request: #3142.

The newer one handles resize, move, column header click, and hopefully selection, and still would greatly appreciate any testing from @AgDude and others.

Since @AgDude pointed out that I can test on my iPad, I think what I have is much more solid.  It works well on the iPad and on my Mac/Chrome (non-touch enabled).  I don't have a device that has both touch and mouse however.

PR #3026, opened 2015-03-17
Fix(sorting): sort all json data

AlphaHinex:
Only first 25 objects are sorted make me confused when using external pagination, so sort all of them.


PaulL1:
Heh.  I was too lazy to do them all, so thanks for this.

PR #3027, opened 2015-03-17
Feat(exporter): add exporterMenuAllData property

AlphaHinex:
Could use this property to show or hide export all as csv&pdf in grid menu. Default value of this property is true (means show). When using external / server-side export, hide the native export all in menu is useful.


PaulL1:
I'm OK with this in that it looks to work although I'm not entirely sure why we'd want to suppress the "all" options separately from the "visible" options.

Issue #2967, opened 2015-03-10
native ui-grid "Export All" feature and external/server-side pagination?


Shanna7463:
Is it possible to use native ui-grid "Export All" feature with external/server-side pagination?  I know I need to implement a customized $http.get call, but I'm just not sure where to add it within the native export implementation.  Has this been done before?

PaulL1:
I think I'd be tempted to use a server-side export in this situation - there are some reasonable pdf generation packages on the server.

If you really want to export client side, then fundamentally you'll have to shift all the data to the client (which takes the point out of pagination somewhat).  I don't think exporter has a hook to say "get all the data", so you perhaps would have to use a custom UI/menu item, and wrap it in a method that first gets all the data into the grid, then calls export.

Shanna7463:
I'm trying to use your advice and implement a custom menu Item, however it doesn't seem to be working correctly.  I update my grid with all data and then I call csvExport:

gridMenuCustomItems: [
                          {
                            title: 'Export All Data',
                            action: function ($event) {

                            	$http.get(url).success(function(data) {
                            		$scope.gridOptions.totalItems = data.totalFeatures;
                            		$scope.gridOptions.data = data.features;
                            		
                              	    var myElement = angular.element(document.querySelectorAll(".custom-csv-link-location"));
                              	    $scope.gridApi.exporter.csvExport( uiGridExporterConstants.ALL, uiGridExporterConstants.ALL, myElement );
                          	  	});
                            }
                          }
                        ],

However, the file only contains the data that was previously displayed in the grid (i.e. I was initially displaying 25 rows, then selected Export All Data, the grid updates with all available data, but the file created only has 25 entries.)  Any idea what I'm doing wrong?

PaulL1:
You may need a timeout or $interval to allow the grid to ingest the data before triggering the export.

AlphaHinex:
Hi @PaulL1 , I did some works on my fork's `export-all` branch to fulfill the same requirement as @Shanna7463 needed. If you think it would do some help, please let me know.

PaulL1:
I'd be happy to take a look at it.  I can only imagine that it'd have a callback for "get the rest of the data", and then it'd wait on that promise before calling export?

AlphaHinex:
yes, user need to supply a callback to "get the rest of the data" if using external pagination, exporter plugin will call it automatically when it found the length of grid data is not equal to total item of grid.

PaulL1:
[...]

PR #3028, opened 2015-03-17
Feat(exporter): export all till all data loaded

AlphaHinex:
When using external pagination and want to use the native export function, this PR will do some help. Also see #2967 .

### DIFF
--- a/misc/tutorial/405_exporting_all_data_complex.ngdoc
+++ b/misc/tutorial/405_exporting_all_data_complex.ngdoc
@@ -1,0 +1,107 @@
+@ngdoc overview
+@name Tutorial: 405 Exporting All Data With External Pagination
+@description
+
+When using build in pagination, the data is fully loaded before export.
+
+For external pagination, use the `exportAll` event to load all grid data and then disable external pagination.
+
+@example
+This shows combined external pagination and sorting.
+<example module="app">
+  <file name="app.js">
+    var app = angular.module('app', ['ngTouch', 'ui.grid', 'ui.grid.pagination', 'ui.grid.selection', 'ui.grid.exporter']);
+
+    app.controller('MainCtrl', [
+      '$scope', '$http', 'uiGridConstants', '$q', function($scope, $http, uiGridConstants, $q) {
+
+        var paginationOptions = {
+          sort: null
+        };
+
+        $scope.gridOptions = {
+          paginationPageSizes: [25, 50, 75],
+          paginationPageSize: 25,
+          useExternalPagination: true,
+          useExternalSorting: true,
+          enableGridMenu: true,
+          columnDefs: [
+            { name: 'name' },
+            { name: 'gender', enableSorting: false },
+            { name: 'company', enableSorting: false }
+          ],
+          exporterAllDataPromise: function() {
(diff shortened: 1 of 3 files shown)


PaulL1:
I'm OK with the concept here, but a couple of things:
1.  I think we should use a promise for the data, rather than doing a $timeout in a loop.  Usually when they request all the data from the server they'll get a promise back, so they can give us that promise (or chain onto it in a then, and give us the resulting promise), and we can just wait on that promise.  

2.  We need to squash all the commits into a single commit before merging.

AlphaHinex:
Is it OK now? I am not familiar with promise, sorry for that.

c0bra:
@PaulL1 any chance you couuld look at this again?

PaulL1:
OK, there's a couple gremlins there, but I need to test it anyway, so I'll merge then update.  I'm not quite sure what the `callback` function does - I think it's not needed with the promise.  And I think in the tutorial we can directly return the $http promise rather than creating a new deferred.  But I'll run it and check.

PaulL1:
I've updated the code a bit for some usage of deferred - I think you'd chained too many things together.

You can directly set the total items and data in the success block of the $http get, and you can return the promise from the $http rather than creating a new one.

I've tidied up a bit in the main code too - you shouldn't need the callback function, you can just put that content into a .then block.  This avoids passing around references too much.

Overall good though - thanks for the contribution.

AlphaHinex:
Thanks for the cleanup works :satisfied:

vijender1256:
@PaulL1 Hi paul, I am looking to implement export all data to pdf on server side paging. I dont see any code here on this page for doing this. Can you please help me to implement this functionality on UI-grid server side paging. appreciate your help.

PaulL1:
There should be code there for export all with paging - there is a callback that is triggered that asks the programmer to get "all data", which can then be exported.

vijender1256:
@PaulL1 I am not seeing any code on this page, can you provide me a link for the code that I can refer to. I greatly appreciate your help :)

vijender1256:
@PaulL1 Thank you, i figured the problem , when binding the Ui-grid properties i have something like this:

 $scope.couponGridOptions = {
            data: 'CouponsList',
            enableRowSelection: true,
            enableRowHeaderSelection: false, etc...

so when i use that property "data" instead of:

  var getPageSuccess = function (result) {
            $scope.couponGridOptions.data = result.Promotions;
            $scope.couponGridOptions.totalItems = result.totalItems;
        };

explicitly loading when page is successful I am had problems.

vijender1256:
@PaulL1 I am having trouble in resetting page number, When I am on page number 2 and I have refresh button which rebinds the ui-grid, when i do that the data is getting rebind but the page number still is 2, how do i reset the page number to 1 after i click on refresh button. Any help is appreciated, lot of people asked this question but I did not find solution. Thanks

PaulL1:
My recollection is that you can set the page number property directly when you reset the group.

vijender1256:
Thank you very much :)  its  $scope.yourGridOptions.paginationCurrentPage = 1; in case for anyone who is looking

PR #3039, opened 2015-03-17
Fix #3036: Update column sorting priority

### LINKED ISSUE #3036: Sort priority should be cleared when a column is unsorted, and only updated if column is not already sorted.
When a column is shift-clicked it cycles the sorting from asc -> desc -> unsorted. Each step of the way the sort priority of the column is incremented. It would seem that the priority should not be updated if the column is already sorted. The sort priority should also be cleared when the column is toggled to unsorted since an unsorted column with a sort priority doesn't make sense.

farooqu:
Update column sorting priority to not increment when sort is toggled, remove when unsorted, and to respect priority of columns with suppressRemoveSort by taking their priority in to account when a new column is sorted.

### DIFF
--- a/src/js/core/factories/Grid.js
+++ b/src/js/core/factories/Grid.js
@@ -1804,8 +1804,10 @@ angular.module('ui.grid')
     if (!add) {
       self.resetColumnSorting(column);
       column.sort.priority = 0;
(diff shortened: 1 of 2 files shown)


PaulL1:
Looks clean and tidy, thanks heaps.

PR #3042, opened 2015-03-17
Dynamic filtering

PaulL1:
Allows filtering to be turned on and off, tutorial updated to demonstrate.


PaulL1:
Bother, failed unit tests, and I cunningly was using the --fast option, so I didn't see the fails in features.  Closing and will update.

PR #2816, opened 2015-02-19
Bringing back menuTemplate property from 2.0.x and adding menuItemTemplate property to GridOptions. Fix for #2428.

farooqu:
In 2.0.x it was possible to specify a custom template for the grid menu using the menuTemplate grid option property. This brings that functionality back.

It also fixes #2428 by allowing the ability to specify a menuItemTemplate property to grid options so as to override the template for all items in the grid menu.

PR #3049, opened 2015-03-18
Fix #2974 and #2816: Enable custom menu templates, and prevent duplicate items in column menu.

### LINKED ISSUE #2974: ColumnMenus: if open second menu without closing first, get duplicate items
Open the first menu, then open the second (or last) without closing the first.
http://plnkr.co/edit/mtEO8fkK6ohqQDpgbyGE?p=preview

farooqu:
This reverts a reversion to bring back the custom menu template functionality from 2.x. It also fixes the bug introduced by that original commit which necessitated the reversion, by putting the show-menu broadcast in a timeout. The ng-if="shown" for uiGridMenu.html had a delay due to $animate. As a result, even though shown was being set to false, the menu was not being removed from the DOM before it was added back in. This caused the previous elements to stick around.

[Review on Reviewable](https://reviewable.io/reviews/angular-ui/ng-grid/3049)


farooqu:
http://plnkr.co/edit/cFAciH?p=preview Plunker updated with the ui-grid.js from my fork.

PaulL1:
OK, looks like it works.  But I'm not clear why it's needed now when it wasn't needed before - is it because the templates are getting retrieved or something?  I'm nervous that the $timeout might fix it in this specific instance, but perhaps if the template retrieve were slower it might not work.

The documentation on each option points to the footer tutorial for more information - is there some more detail somewhere about what these templates can be used for?

farooqu:
I'm not sure why it would be needed now but wasn't in the past. The uiGridMenu showMenu and hideMenu methods both have a timeout to wait a digest cycle for the ng-if to be removed. The behavior I was seeing when debugging was that even though shown in uiGridMenu was being set to false, it was being re-set to true due to the $scope.$broadcast('show menu') before the digest cycle could apply the ng-if and remove the old menu items from the DOM.

farooqu:
Regarding the documentation, that was an oversight in copy pasting the footer doc. I'll try and see if I can get some documentation added this weekend.

Our primary use case was to override the menu button's toggle method to expand it to include the ability to close the menu using the "ESC" key, as well as to modify the look of the menu to display a check-box with the label next to each item.

tuhoojabotti:
Would be great to see this option. We have styled the grid to match our material design app, but it would be nice to be able to override the menu to use material components.

mportuga:
@farooqu You appear to have some merge conflicts. Are you still interested in merging this? If so, would you mind resolving the conflicts so that I or someone else can  review it now?

mportuga:
Closing due to inactivity.

PR #3056, opened 2015-03-19
Update saveState.js

NicolasN:
I don't know why and I don't understand why, but if I save column.visible then savedColumn.visible is always set to true !?! 
But if I save column.colDef.visible it's ok :/

### DIFF
--- a/src/features/saveState/js/saveState.js
+++ b/src/features/saveState/js/saveState.js
@@ -320,7 +320,7 @@
             savedColumn.name = column.name;
             
             if ( grid.options.saveVisible ){
-              savedColumn.visible = column.visible;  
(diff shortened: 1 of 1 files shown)


PaulL1:
This looks to be breaking the unit tests, so won't merge.  The aim is the save the current state, not the colDef state. Having said that, there is some funny treatment of col.visible, it usually updates colDef.visible as well for a reason I haven't worked out.

If you can provide a plunker that illustrates your problem, I'll take a look at a different way to fix this.

NicolasN:
I made a mistake in my own code, so there is no problem with ui-grid. Sorry for the noise ...
Anyway thanks for your work, ui-grid is a huge step forward compared to ng-grid.

PR #3060, opened 2015-03-19
Fix #3034: don't allow a negative viewport height

### LINKED ISSUE #3034: Negative viewport height
We've run into an issue where it is possible to resize a grid such that the getViewportHeight() method returns a negative value (basically anytime the gridHeight is less than the headerHeight).

Granted, this is an edge case but it would be nice if ui-grid could protect against this since a negative value is not valid for the CSS height property.

Seems like this should be beyond trivial (`return (viewPortHeight > 0) ? viewPortHeight : 0;`) in both getViewportHeight methods, but I'm not familiar with the ui-grid code so I don't know all the implications of changing this.

priceld:
It doesn't make sense that the viewport height would ever be less than zero.

The change in `getViewportHeight()` exposed a problem in cellnav which I fixed and also updated the unit tests to be a little more meaningful.

[Review on Reviewable](https://reviewable.io/reviews/angular-ui/ng-grid/3060)


PaulL1:
@swalters: I think this one probably needs your review as I think you've recently been in this code.

c0bra:
@priceld I'm really sorry it's taken me so long to get to this but I'm checking it over now.

It looks like there's was a lot of refactoring in the scrolling code around the same time that you submitted this PR. For instance, the `1scrollToIfNecessary` method has been refactored from the Scrolling feature to the core Grid class.

Would you be interested in trying to rebase on current master and adjust your changes to suit?

c0bra:
I'm having a hell of a time figuring out the merge for this. For some reason the diff when I try to apply your commit locally is way way different than what's here.

If you want I can try to pull out your code changes myself and commit them, or you can resubmit your PR.

priceld:
Sorry for taking so long.  I'd be glad to merge my changes locally and open an new PR.  It may be a week or so before I can get to it though.

mportuga:
@priceld I am closing your PR for now due to merge conflicts. But feel free to reopen it when you are ready.

PR #3069, opened 2015-03-20
Fix #3061: Add option to opt out of initial grid resize

### LINKED ISSUE #3061: Option to opt out of initial grid resize for small grid heights
First, I should clarify that we handle our grid size/resize a little bit differently than what is perhaps intended.  Our webapp has a layout (like jQuery UI Layout) and our grids can live in any one of the layout panes (north, south, east, west, or center).  We have the grids fill the entire pane (in CSS we set top, bottom, left, and right to 0) and when the pane gets resized, we fire "handleWindowResize()" on the grid to update the size of the grid.

However, when a grid is initialized in a pane that is exceptionally short, ui-grid detects this and gives the grid a bigger default height automatically.  This is coming from a section of code in the grid's "post" function (src/js/core/directives/ui-grid.js).  
```
// If the grid isn't tall enough to fit a single row, it's kind of useless. Resize it to fit a minimum number of rows
if (grid.gridHeight < grid.options.rowHeight) {
  // resize
}
```
This code attempts to resize the grid for you by adding a "height" style property to the grid's DOM element.

This breaks our usage since ui-grid detects the grid's height by asking the DOM element (`gridUtil.elementHeight($elm)`) and whereas before it was dynamic it is now set statically in the grid element itself.

It seems like perhaps the best workaround for us, is to be able to opt out of this initial grid resize.  Granted it's a bit obscure, but still we'd like to add an option to allow users to bypass this part of the grid's initialization.

Thoughts?

priceld:


PaulL1:
Looks good, merging.  Thanks.

PR #3076, opened 2015-03-20
Update rowSorter.js

bringking:
Updated rowSorter to use an object instead of an array. This fixes an edge case where, you have a column named "sort", and the code would basically overwrite the default Array.sort function. Here is a plunk demonstrating the bad behavior.

http://plnkr.co/edit/feSTTpNUEGoq4nnAz5KK?p=preview


PaulL1:
Nice catch - that could have confused someone for a long time!!

PR #3085, opened 2015-03-23
Fix(infiniteScroll): fix #2730, fix #2827, BREAKING changes

### LINKED ISSUE #2827: Infinite Scroll not working when reusing a grid
Infinite scroll doesn't seem to work when reusing a grid which previously hit the end.... Makes sense that the callback would stop triggering if the listener is unregistered or something, but how to make this work? 

I have a search form with some parameters in, i click search and then it queries the database and returns results which are put into this grid. I do a search, look at the data/scroll etc, then do another search. The data in the grid is wiped and the new data put in. (I.e. using same grid) - from this I notice 2 problems:

1) When I perform a second database search the scroll position does not reset to the top – and without using a new directive (cellNav) to scrollTo, I don’t know how.

2) Similar to this, infinite scrolling seems to stop working when you do a new search.

Is there a way to recreate the DOM element, or get the grid refreshed (do it's initialization and call registerGridApi again)?

PaulL1:
Substantial refactor of infinite scroll.  Provides new functionality, a much more
detailed tutorial, and removes the code that crept into core in #2730 (which I merged
when I shouldn't have).  Clarifies the handling, removes a viewPort directive that was
unnecessary, makes the handling for scroll up and down consistent.

Strictly speaking the API hasn't changed, but there are some new settings and the
default behaviour has subtly changed in that scrollUp isn't enabled by default.  This
should be more logical to users, but probably does require a review of the tutorial and API.

### DIFF
--- a/misc/tutorial/212_infinite_scroll.ngdoc
+++ b/misc/tutorial/212_infinite_scroll.ngdoc
@@ -2,85 +2,104 @@
 @name Tutorial: 212 Infinite scroll
 @description
 
-The infinite scroll feature allows the user to lazy load their data to gridOptions.data
+The infinite scroll feature allows the user to lazy load their data to gridOptions.data.  
 
-Specify percentage when lazy load should trigger:
-<pre>
-  $scope.gridOptions.infiniteScroll = 20;
-</pre>
+Once you reach the top (or bottom) of your real data set, you can notify that no more pages exist
+up (or down), and infinite scroll will stop triggering events in that direction.  You can also
+optionally tell us up-front that there are no more pages up through `infiniteScrollUp = true` or down through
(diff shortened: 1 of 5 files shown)


PaulL1:
@mloutraris-ge, @novice3030: there are significant changes to infinite scroll in here.  It would be good to get a review and check it's still working well for you.  Note that there are some API changes that probably don't impact anyone but might - mostly in terms of default behaviour.

@swalters: I've stripped the infinite scroll stuff out of core - this impacts the ui-grid directive and the gridFactory.  I suspect you are touching these in your scroll refactoring, so worth noting that they go away - you can safely delete them.

mloutraris-ge:
We will check it out and let you know. Thanks!

PR #2914, opened 2015-03-05
Update Grid.js

jmmartinez84:
https://github.com/angular-ui/ng-grid/issues/2889

PR #3090, opened 2015-03-23
fix(Grid): column rendering issue when adding row

jonathanchris:
Looks like this was just a typo, but fixes column rendering when there is horizontal scrolling and scrolled far right. Hard to reproduce this bug, but this fixes the issue I was having.


jonathanchris:
I was working off an older version, I'm realizing now that this was already fixed with this #2914. I'm going to close this.

PR #3100, opened 2015-03-24
added support for scrollbars.WHEN_NEEDED back in

jcompagner:
It only puts out "auto" for vertical or horizontal in styling. For calculation heights it works as if it was NEVER.

added test for them also

[Review on Reviewable](https://reviewable.io/reviews/angular-ui/ng-grid/3100)

### DIFF
--- a/src/js/core/constants.js
+++ b/src/js/core/constants.js
@@ -107,8 +107,8 @@
     },
     scrollbars: {
       NEVER: 0,
-      ALWAYS: 1
-      //WHEN_NEEDED: 2
+      ALWAYS: 1,
+      WHEN_NEEDED: 2
     }
   });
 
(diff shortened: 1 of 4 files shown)


swalters:
I think calculation heights must be exact.  They must know if a scroll bar is present or not. Am I wrong?

jcompagner:
don't think that matters, what happens now is that if i design a ui-grid to have 4 columns each 100px wide, so totally 400 and ui-grid it self is also exactly that size 
then the scrollbar will go a bit over the last column. But what can we do here? 

with the previous code (which is stil a lot in the code base i think, commented out on various places) we had those "native" scrollbars directives then we needed to do a lot more calculations to exactly get everything right

now the overflow is really on the divs that holds the viewport (or is the viewport) so the browser takes care of it.

i do see in the code that we do calculate with some scollbarHeight and scrollbarWidth (again a lot commented out). But what to do there? don't think that code is called all the time to relayout everything when suddenly the scrollbars do appear

Issue #3084, opened 2015-03-22
ui-grid-move-columns issue -Not able to move columns properly 


njosh:
This issue is also replicable on tutorial pages

http://ui-grid.info/docs/#/tutorial/310_column_moving

when i move columns multiple times the border start getting darker and after sometime not able to move any column.Please advice

Thanks
![image](https://cloud.githubusercontent.com/assets/11484810/6768552/302020da-d099-11e4-9630-7759ac8675f4.png)

c0bra:
Wow, that is ridiculous. Definitely a bug; thanks for reporting.

jpuri:
This looks like something new, I will check.

jpuri:
The problem was this "$on" in line 148 in ui-grid-header-cell.js.

                $scope.$on('$destroy', function () {
                  $contentsElm.off('mousedown touchstart');
                });

Because of this old click handlers of header cell were not removed and were creating trouble for move column when header was clicked.

jpuri:
The reason is apparently this line of code, this is leading to event handlers being bind multiple number of times:

            var dataChangeDereg = $scope.grid.registerDataChangeCallback( updateHeaderOptions, [uiGridConstants.dataChange.COLUMN]);

@PaulL1 , this line was added as part of commit "Enh(filtering): Fix #2067 allow enabling filtering dynamically". We need to move all event bindings outside "updateHeaderOptions", so that events are not bind repeatedly. Please suggest.

PaulL1:
@jpuri: most likely what's happening is that the dataChangeCallback is being called (via notifyDataChange), and it's registering the handlers again without removing the old ones.  The logical thing to do would be to modify the updateHeaderOptions function to remove any existing handlers before registering new ones.

jpuri:
Hi @PaulL1 ,

What is happening is when column is moved first time one event is generated, when second time and third time 4 and number of events keep doubling. Thus move column is broken.

The reason could be when we move column, Grid is also re-rendered so some of the event handlers are just left there. If I comment the line above, its fixed.

We can do either of these:
1. Bind event only once and move code of event binding outside this function 'updateHeaderOptions'.
2. Each time function is called bind event only if it was not bind before.

As you mentioned is not possible to remove existing binding.

PaulL1:
It should be possible to remove the existing binding - you call $off and pass in the same function that we used in $on.  That means that the function needs to be declared on the $scope so that we can always get the same instance.

[...]

PR #3114, opened 2015-03-25
Fix for #3084

jpuri:
Move column was broken because of click events not getting detached. PR has fix for the issue.


jpuri:
Seems to be some issue with the commit, please wait as I update the PR.

PR #3117, opened 2015-03-26
Added Czech localization

pionl:
Hi,

I've added Czech localization file for multiple key combination (cs, cz, cs-cz, cs-CZ). Tested on latest version.

Martin.


PaulL1:
Thanks.

PR #3120, opened 2015-03-26
refactor(core): Change scrolling events. Various performance improvement...

swalters:
...s in scrolling

Before this change, scrolling Events were being fire for each render container.  Now, a ScrollBegin and ScrollEnd event is fired once for the grid.  ScrollTo methods were moved from cellNav into core and also changed to return a promise.

 BREAKING CHANGE:
 Two events are now emitted on scroll:

 grid.api.core.ScrollBegin
 grid.api.core.ScrollEnd

 Before:
 grid.api.core.ScrollEvent
 After:
grid.api.core.ScrollBegin

ScrollToIfNecessary and ScrollTo moved from cellNav to core and grid removed from arguments
Before:
grid.api.cellNav.ScrollToIfNecessary(grid, gridRow, gridCol)
grid.api.cellNav.ScrollTo(grid, rowEntity, colDef)

After:
grid.api.core.ScrollToIfNecessary(gridRow, gridCol)
grid.api.core.ScrollTo(rowEntity, colDef)

GridEdit/cellNav
When using cellNav, a cell no longer receives focus.  Instead the viewport always receives focus.  This eliminated many bugs associated with scrolling and focus.

If you have a custom editor, you will no longer receive keyDown/Up events from the readonly cell.  Use the cellNav api viewPortKeyDown to capture any needed keydown events.  see GridEdit.js for an example


swalters:
MY COMMENTS IN CAPS ON LEFT
From @PaulL1 on gitter:

I think the summary is:
DONE we should clean up the commented out code. Not critical, but we'll have to do it later if not now

DONE it'd be nice to take $timeout out of uiGridEditor - it's not used just included. Minor, not critical

DONE scroll propagation to parent. I think this we should fix before merging - it took us ages to get it right and I'm not keen to go backwards again. But I could be talked out of this and raise a hi-pri issue instead

TODO rtl is busted at least for MacBook, I'd raise a bug for this. But I'm also aware that @c0bra normally fixes the scrolling stuff and doesn't have a MacBook, so I'm not sure how we'll fix it again if we merge with it broken
tabbing into the grid is broken. That's not a huge problem, so I'd raise a bug for it. But it's a regression

WORKS ON MY BOX scrollTo on the cellNav tutorial is a blocking defect, I reckon we should fix that before merging

BY DESIGN minor issue with selecting a row sometimes causing a scroll move unexpectedly. Ugly but not critical. It does this a bit also with editOnFocus as you tab across the grid. Also ugly but not critical

?? if you have a cell in edit mode and scroll away from the grid and attempt to focus a non-grid widget, the focus shifts back to the grid. Again probably minor but ugly

BY DESIGN if you have a cell in edit mode, then scroll, the cell loses edit mode (edit on focus http://localhost:9003/docs/#/tutorial/309_editable_with_cellnav), but keeps focus. That's annoying

TODO if you have a cell in edit mode then scroll as above, you don't appear to have a keyboard way to get focus back - if you press tab it tabs off the entire grid and you can't tab back into it, space doesn't seem to work, cursor keys scroll the grid. I think we have focus on the container but not on a cell

rkilarski:
Hello!  You said:

<blockquote>If you have a custom editor, you will no longer receive keyDown/Up events from the readonly cell. Use the cellNav api viewPortKeyDown to capture any needed keydown events.</blockquote>

The keys I need to capture are the Enter and right/left arrows.  Since those are considered navigation keys in uiGridCtrl.cellNav.handleKeyDown(), they aren't emitted via viewPortKeyDown.

What can I do in order to allow handling Enter and the horizontal arrows?

Rich

PR #3132, opened 2015-03-27
added filter attribute 'disableCancelFilterButton'

pedroclayman:
added filter attribute 'disableCancelFilterButton':
if set to true then the 'x' button that clears the filter is not rendered.

select filter default empty option:
I have added an empty option into the filter select which is necessary if the clear button is disabled.


PaulL1:
@pedroclayman: I'm OK to merge this, but unfortunately it looks to have a merge conflict now.  Any chance you could resolve that conflict?

pedroclayman:
there are some failing tests on master apparently

Issue #2616, opened 2015-01-23
Sort by clicking column header is broken on touch-enabled Windows device


AgDude:
I observed this in my own project, as well as in the tutorial here: http://ui-grid.info/docs/#/tutorial/102_sorting

I can sort by using the column menu, but not by clicking on the header. I am using Windows 8.1 and Chrome 40.

jpuri:
It works for me in the tutorial link you shared above except if column has:
enableSorting: false 

I checked in chrome version 40.0.2214.91.

AgDude:
I can confirm that this issue does not occur on Linux Chrome 34. On windows I am using the same version of chrome as @jpuri. I suspect this is related to #2550. 

I have windows 8.1, running on a laptop with touch support. I am not using the touchscreen (connected to external monitors), but I wonder if this is a bug in how chrome is detecting a "touch" device...

PaulL1:
@AgDude: that sounds like a reasonable supposition.  But unfortunately I don't have a touch enabled windows machine to check.  Are you able to put some breakpoints in the code and find out what sort of event (if any) is being fired when you click on the column header?

I'd suggest in ui-grid-header, around about the lines that show:
```
            if ($scope.sortable || $scope.colMenu) {
              // Long-click (for mobile)
              var cancelMousedownTimeout;
              var mousedownStartTime = 0;
```

Looking at the code, I suspect you're right on your diagnosis, because it does:
```
              var downEvent = gridUtil.isTouchEnabled() ? 'touchstart' : 'mousedown';
              $contentsElm.on(downEvent, function(event) {
```
And I'm guessing that the event you're getting is a mousedown, but because your device is touch enabled we're waiting on a touchstart.

I wonder what would happen if we always just waited on both.....what harm that would cause.

AgDude:
You are correct about the events. ```downEvent``` evaluates as "touchstart", but since I am using a mouse it would never occur. I don't see any harm in listening for both.

c0bra:
I've changed the title to clarify the issue.  Unfortunately listening to both can be problematic because BOTH events will fire during a touch/click sequence, causing extra event handling and breaking stuff.

I don't have a Windows touch device to test with. @AgDude Do you know if there's any way to do emulation, maybe in IE11?

AgDude:
@c0bra Chrome has emulation mode, it is in a tab next to the console. The bottom option is a "touch notebook". I  can't verify it will work, since I am on a touch laptop at the moment.

[...]

Issue #2550, opened 2015-01-14
Resize Columns in Chrome on Windows 8 (touch-enabled) not working


jaskaye17:
There appears to be an issue re-sizing the column widths using the ui-grid-resize-columns directive in Chrome. The functionality appears to work fine in IE and Firefox. I expect the column width to readjust on drag; however, nothing  happens in Chrome when I try to drag with the 'drag icon'. This can be seen at the following plnkr:

http://plnkr.co/edit/iC3IKKVOVoPoX9T9qOXV?p=preview

This may have a relation to a separate issue that I have noticed with ui-grid-move-columns directive not working properly in Chrome, but, working in Firefox and IE. The issue can also be seen in the above plnkr. A recent bug issue was filed and resolved which appeared to address this issue. I am still; however, noticing the problem described above #2354.

PaulL1:
It's working for me in Chrome on the tutorial: http://ui-grid.info/docs/#/tutorial/204_column_resizing

Resize is also working for me with your plunker.  What version of Chrome, and what OS are you using?

jaskaye17:
Hey @PaulL1 Thanks for the prompt response!

I am using Chrome: Version 39.0.2171.99 m and windows: Version 8.1 Pro

PaulL1:
And are you also getting this problem on the tutorial pages, or only in the plunker?  I have 39.0.2171.95 (64-bit), but on OSX.  I wouldn't expect that to make a difference, but I see Chrome is currently updating itself, so maybe when that finishes I'll have .99 and it might break.

PaulL1:
Are you using expandable?  I see issue #2277, which may be related.

jaskaye17:
I am having the same problem directly from the tutorial page as well as from the plunker. In this demo/test I am not using expandable; just what is in the demo and plunker without any tinkering. Very strange. My coworker, who also has the same computer setup/dev environment, has the same issue using the feature in chrome.

cmaxo:
I have the same issue. If I run Firefox it totally works. My chrome version is 39.0.2171.99 m on windows pro 8.1.

AgDude:
I can confirm this on windows 8.1 Chrome Version 40.0.2214.91 

It is working fine on Ubuntu with Chrome Version 34.0.1847.116

I have observed this on the 204 tutorial page.

c0bra:
[...]

Issue #2303, opened 2014-12-09
Move columns not working well on touch devices


jpuri:
In grids with huge horizontal scroll the moved columns ar e not getting repositioned correctly. I am working on this issue.

jpuri:
Hi,

I am now working to fix this up.

jpuri:
I was trying to use touchEvents (touchstart, touchmove and touchend/ touchcancel). Strange but fact is that touchend after touchmove is not well supported by most of touch devices.

http://stackoverflow.com/questions/11314185/multiple-touches-touchend-event-fired-only-when-a-touchmove-occurs

I am not very sure how can we implement this feature on touch devices. @PaulL1 any suggestion ?

PaulL1:
Absolutely no idea.  :-)  I do know that we need to be careful that just because devices support touch events doesn't mean that they don't also have a mouse - we have a few issues with laptops that are touch enabled but people are trying to use a mouse on them and it's not working.

I think if we're going to try to do this on touch devices we need to write down what semantics we want for each of the touch events on a header.  Because we're trying to do sort, menu open and drag all at the same time, and I'm not sure there's enough touch events to do all of them.

jpuri:
The issue appears to be that different devices are firing different events, but yes its true that our column headers have a lot of events to handle.

@swalters once suggested to implement move columns differently. I am not very sure might be we can come up with a better design for move columns. I have time these days to do any changes or fix issues.

jpuri:
Might be in our docs/ tutorials we can mentions which features will/ will not work with touch devices.

PaulL1:
@jpuri, I think #3142 will fix this.  At least it does on my iPad, not sure about other touch enabled devices.  Do you want to test it out?

PaulL1:
I think this is resolved.  Closing.

PR #3142, opened 2015-03-28
Fix(touch events): touch events on resize, move, header click/sort/menu,...

### LINKED ISSUE #3084: ui-grid-move-columns issue -Not able to move columns properly

PaulL1:
... selection

Fix #3084, #2838, #2616, #2550, #2303.  All relate to handling of events
on touch devices.  General philosophy is to register down events for both
touch and mouse, but once we get a down event, we only listen for move and
up events of the type we got a down on.  This is a similar approach to
hammer.js.

Also includes a big refactor of how move columns listens for events, makes
it more like other handlers.

This all works for me on iPad, and on OSX/chrome (non-touch).  I don't have
the ability to test a machine that has both mouse and touch, so waiting on
someone else to look at that.

### DIFF
--- a/src/features/move-columns/js/column-movable.js
+++ b/src/features/move-columns/js/column-movable.js
@@ -247,220 +247,225 @@
 
               if ($scope.col.colDef.enableColumnMoving) {
 
-                $scope.$on(uiGridConstants.events.COLUMN_HEADER_CLICK, function (event, args) {
+                /*
+                 * Our general approach to column move is that we listen to a touchstart or mousedown
+                 * event over the column header.  When we hear one, then we wait for a move of the same type
+                 * - if we are a touchstart then we listen for a touchmove, if we are a mousedown we listen for
+                 * a mousemove (i.e. a drag) before we decide that there's a move underway.  If there's never a move,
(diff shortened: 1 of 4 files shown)


PaulL1:
I would greatly appreciate a review and test from @AgDude, and I think @c0bra was also looking at things in this area.  

From my viewpoint this is ready to merge - it works on my iPad and it hasn't broken non-touch devices (at least not my MacBook on Chrome).  It now allows move on the iPad as well, which we haven't really had working before.

c0bra:
@PaulL1 @AgDude This looks good for me in Chrome with the touch flags enabled, and in FF no touch events!

Awesome work, Paul.

AgDude:
I had every intention of reviewing this on Saturday, but I didn't get to it. I'll try to look at it this afternoon or tomorrow. But no offense taken will be taken if you merge it before then.

PaulL1:
I'm doing paid work next couple days so prob won't get to merge. If you test it before then great, if @c0bra wants to merge earlier then also great.

PR #3149, opened 2015-03-29
Date Filter support other formats

SSchaaf:
Date Filter only support yyyy-m-d format.

image

This PR uses CellFilter to Date Filter format. So we can use other formats like dd/mm/yyyy, mm/dd/yyy, etc..

It is necessary to include moment.js as a dependecy.


swalters:
I'm sorry, but we don't allow any third party library dependencies in the core.  Ui-Grid's only dependency is Angular.

SSchaaf:
And about date's format ? Any idea to get the same behaivour ?

Issue #3123, opened 2015-03-26
Add pinning to saveState


jonathanchris:
I think the saveState feature should be able to include pinning. This way the grid can restore your pinned columns. Currently the saveState doesn't remember you pinned columns, this feels like a missed opportunity.

PaulL1:
It can easily be added, it's very extensible, it's just that nobody has added it as yet.  A PR would be welcomed.

jonathanchris:
Awesome, coming soon!

PaulL1:
I think the most elegant way to do this (after doing something similar on grouping) is for the pinning feature to offer a "get pinning state" and a "set pinning state", which save state then just calls.  Much tidier than having saveState knowing about the pinning internals.

jonathanchris:
@PaulL1 Take a look and let me know if you have any suggestions/recommendations on this PR. I used the publicApi and created an event and method in the pinning feature. Then utilized this in the saveState feature. I attempted to add testing and documentation as I went along.

PaulL1:
This merged some time ago and seems to be working well.

PR #3161, opened 2015-03-30
feat(saveState): add pinning to save state

jonathanchris:
The idea was to update saveState to include pinning. This turned
out to be more involved than initial anticipated. I ended up
having to update the pinning feature to include a public event
and method so saveState could include pinning as an option.

I updated documentation as I worked through it and even updated the
tutorial to include pinning. This is in reference to an feature
request I made recently #3123.

### DIFF
--- a/misc/demo/grid-save.html
+++ b/misc/demo/grid-save.html
@@ -1,0 +1,83 @@
+<!DOCTYPE html>
+<html class="no-js" ng-app="test"><!--<![endif]-->
+<head>
+    <meta charset="utf-8">
+    <meta content="IE=edge,chrome=1" http-equiv="X-UA-Compatible">
+    <title></title>
+    <meta content="width=device-width" name="viewport">
+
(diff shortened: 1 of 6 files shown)


PaulL1:
Looks good, I'll merge.

I have a reservation about the double refresh before the column shows - I thought I'd fixed things so we didn't need the double refresh.  I'll take a look and potentially remove that bit once it's merged.

jonathanchris:
Yeah i think i put a comment in there about the double refresh. It didn't pass my smell test either, but it was the only way I could get it to work, so I left it in there :smile: 
Thanks again, glad I could be a part of such a great project!

PaulL1:
I've actually fixed that double refresh now too.  I think you submitted this PR before I'd done the code in core that removed the need for it - so we sort of crossed.  Basically I just took it out and it worked, because the newer code was there.

But probably worth you checking that it also still works for you and that I didn't miss anything.

PR #3197, opened 2015-04-02
added support for filter terms equal 0 and false

pedroclayman:
Filtering using a term that equaled 0 or false did not work as expected since the terms have been used directly in logical conditions.


c0bra:
Check out the travis build. Looks like your tests are failing.

pedroclayman:
Seems like I forgot to commit something.

pedroclayman:
I have fixed the problem but messed my branch a bit.
I will recreate a pull request under the same name in a minute

PR #3201, opened 2015-04-02
Added dynamic expanded row height ability

yccteam:
I've added 2 main things here:
1) For each expandable row, specify it's height.
2) And event that triggers before the change, so the height could be set dynamically on the run.

[Review on Reviewable](https://reviewable.io/reviews/angular-ui/ng-grid/3201)

### DIFF
--- a/src/features/expandable/js/expandable.js
+++ b/src/features/expandable/js/expandable.js
@@ -134,6 +134,8 @@
                * @param {object} rowEntity the data entity for the row you want to expand
                */              
               toggleRowExpansion: function (rowEntity) {
+                event.stopPropagation();
+                event.preventDefault();
                 var row = grid.getRow(rowEntity);
                 if (row !== null) {
                   service.toggleRowExpansion(grid, row);
--- a/src/features/expandable/js/expandable.js
+++ b/src/features/expandable/js/expandable.js
@@ -186,11 +186,12 @@
       },
       
       toggleRowExpansion: function (grid, row) {
-          // trigger the "before change" event. Can change row height dynamically this way.
-           grid.api.expandable.raise.rowExpandedBeforeStateChanged(row);
+        // trigger the "before change" event. Can change row height dynamically this way.
+        grid.api.expandable.raise.rowExpandedBeforeStateChanged(row);
         row.isExpanded = !row.isExpanded;
-          if (angular.isUndefined(row.expandedRowHeight))
-              row.expandedRowHeight = grid.options.expandableRowHeight;
+        if (angular.isUndefined(row.expandedRowHeight)){
+          row.expandedRowHeight = grid.options.expandableRowHeight;
+        }
               
         if (row.isExpanded) {
           row.height = row.grid.options.rowHeight + row.expandedRowHeight;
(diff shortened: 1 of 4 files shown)


yccteam:
I've fixed the jsHint error

bierdok:
+1

choomz:
+1

Karov28:
+1

AgDude:
+1 comments are really not helpful. Please don't send us all spam. 

Expandable isn't really my area, but this sounds like a good idea. If someone would take the time to rebase and squash this PR, that would greatly improve its chances of being merged.

yccteam:
@AgDude , 
What's "rebase" and "squash"? Some ec6 stuff ? ;)

I've asked for some feedback. If you could guide me as to what needs to be done, I'd be more than willing to get more into it. I'm using this solution in my app, and it works great. Just tell me what needs to be done and I'm on it. I've done lots of fixes to several bugs, but since my PR's are left to dry, I stopped loading them (tried a new one a week ago... just in case :)).

I know you guys are busy, and might not have the time, but a little feedback of what's wrong might help me solve the issue.

Thanks

AgDude:
@yccteam thanks for your contributions. Rebase and squash are git commands, which help to make your PRs easier to read and merge.

rebase is a means of "syncing your fork", so it merges nicely into the latest master. Here are instructions using merge (I typically use rebase, but it should have the same effect)
https://help.github.com/articles/syncing-a-fork/

squashing combines your commits into one. Here are some methods for squashing: 
http://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git

A PR should typically have only one commit. The exception being if there are two significantly different parts of the PR. An example where two commits would be preferred is fixing a bug in the core, then adding a new feature which needed that bug to be fixed.

https://github.com/angular-ui/ng-grid/blob/master/CONTRIBUTING.md also has some pointers.

If you have trouble with this ask on gitter, I will most likely be there, and if not I am sure someone else can help you.

yccteam:
Thanks. I'll try to get to this one during the weekend.

fastex:
Are you fixes also incorrect measurement if we scroll down grid with expanded rows?

swalters:
@yccteam was taking a look at this today.  I was hoping it fixed the issue on 306 tutorial when you expand all rows and then scroll.  

Can you do a merge with master today and then squash all of these commits into one PR?  If so, I'll merge it.

Phara0h:
I really need this feature, this would be great. Also just saying auto height based on child div would be the best.

Namek:
@yccteam are you going to rebase+squash it?

JLLeitschuh:
Closed by #4343

yccteam:
Hi,
Sorry for the late response. I'm working like 24/7 lately on an offline system, so not much access to the internet.
Glad to see you've managed to merge this. One issue I've found is that the virtual scrolling can act strangely when setting up different heights to the rows.
I've started working on a solution when the project I'm working on started going wild. I've found out that the scrolling calculation is done by multiplying the number of rows by a constant, and it's not taking into account the different rows' heights.
Have you encountered this issue?

JayHuang:
@yccteam yes, this is also the issue my team has had with my own (similar) solution. Would be great if there's a way to resolve this without significantly impacting the virtualization performance.

PR #3211, opened 2015-04-04
added support for filter terms equal 0 and false

pedroclayman:
Filtering using a term that equaled 0 or false did not work as expected since the terms have been used directly in logical conditions.


PaulL1:
Looks great thanks.

PR #3237, opened 2015-04-08
Update i18n nl.js for grouping feature

timstuyckens:
avoid crash on  title: i18nService.get().grouping.group,       when language is set to dutch


PaulL1:
Thanks,

Issue #2700, opened 2015-02-03
grid visual artifact within modal when width is 100%


longshot5112:
I have a modal which contains a grid which should be 100% of the modal width. Upon initial loading of the modal, there is a visual artifact of the grid beyond the boundary of my modal. If I re-size my browser window, the artifact will disappear and the grid will "snap" in place where it should be and start obeying the 100% width boundary of my modal window.

This only seems to happen when loading a grid in a modal. My primary grid on the main page is set to width: 100% and has no issue.

Initial load:
![image](https://cloud.githubusercontent.com/assets/10211493/6024658/6bf0b376-ab92-11e4-97ad-328f9dc703f4.png)

After resizing:
![image](https://cloud.githubusercontent.com/assets/10211493/6024680/98017e50-ab92-11e4-857d-4c50457add26.png)

This is the CSS I have for the grid:
![image](https://cloud.githubusercontent.com/assets/10211493/6024725/ec587d96-ab92-11e4-99c2-feec4fcd0cf1.png)

lifinanalytics:
Hey actually this issue is part of ui-view-port directive, as a part of workaround replace ui-grid div with following code.
 <div   ui-grid="gridOpt" class="grid" style="width:100%" ng-style="getTableStyle()" ></div>
Use following code on your controller.
  $scope.getTableStyle= function() {
                           
                            return {
                               width: "100%",
                               height:"100%"                            
      };
};

lifinanalytics:
Hey actually this issue is part of ui-view-port directive, as a part of workaround replace ui-grid div with following code.
<code>&lt;div  ui-grid="gridOpt" class="grid" style="width:100%" ng-style="getTableStyle()" &lt;div> </code>
Use following code on your controller.
<code>
  $scope.getTableStyle= function() {                           
                            return {
                               width: "100%",
                               height:"100%"                            
      };
};
</code>

tony0918:
I got the same issue. And @lifinanalytics, your workaround doesn't work for me. If the parent container's width has been set by %, Grid cannot get the correct width of parent container's width. When resizing occurs, because of parent container won't be rendered again, Grid can get correct width.
Any other solution or ETA?
Thanks.

httpete:
Do this after setting data, as a workaround.

window.dispatchEvent(new Event('resize'));

lifinanalytics:
[...]

PR #3238, opened 2015-04-08
fix(uiGrid): Wait for grid to get dimensions before rendering

c0bra:
This is a fix for #2700 and a few other issues.

The problem here is that for determining the grid width, gridUtil is detecting that when the grid is initialized, it has an offsetWidth of 0, so it creates a "fake" (cloned) element that it attaches to the document with visibility: hidden in order to measure it.

This causes a problem because with width: 100% when attached to the document body the grid will just fill up the whole space. THEN when the real grid is in the modal it takes up way more space than it should.

This change makes the grid wait up to 2 seconds for the element to have a width > 0 before it initializes the dimensions.

Note that this change also has a breaking change, which is that gridUtil can no longer try to get the width of hidden elements, which is the same as jQuery. Hidden elements really don't have any dimension.

### DIFF
--- a/misc/demo/modal.html
+++ b/misc/demo/modal.html
@@ -1,0 +1,70 @@
+<!DOCTYPE html>
+<html ng-app="app">
+
+  <head>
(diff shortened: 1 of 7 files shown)


c0bra:
I'm going to go ahead and merge since there's no feedback either way.

PR #3241, opened 2015-04-09
Added 'numberStr' option to colDef.type make use of sortNumberStr functi...

dllabs:
As per [discussion with PaulL1](https://github.com/angular-ui/ng-grid/issues/3240)


PaulL1:
Excellent, thanks.

Issue #3246, opened 2015-04-09
[FEAT] - tree-view grid


yccteam:
This feature would be a combination of grouping and expandable row. Each row could be expanded, but the expansion would be considered as a part of the main grid, so their virtualization be calculated together.
So the look and feel would be like a grouped grid, but the rows would look like an expandable grid (the "group headers" would have their own data sources).

Some thoughts:

@PaulL1 :
1) ```
potentially treeview could be added to grouping rather than duplicating. Probably the main change is to use the existing row as a groupHeader rather than inserting a new row. But not quite sure how we'd decide which row is the "parent" row
since at the moment we sort then look for breaks in value. Would the first row with a given value always be the "parent". Or would we have some other way to tell the code which rows are nodes and which rows are leaves?```

2) ```Thinking on the treeview @yccteam, perhaps it could be a lot like grouping, but you could have an optional property on each row ($$treeLevel), and then it could use a rowsProcessor similar to the group rowsProcessor to work out which are nodes and which are leaves (perhaps it could set a property on the row like row.treeLevel, and row.treeNode = true)
Further, we could store the number of child nodes for any given node in row.numChildren, and then use a custom cell template to add the number of children in brackets if the user wanted
You could reuse the header column logic and the indenting of the + from grouping
it'd be nice to have a combined code base, but perhaps it's easier to just copy the bits that are useful.
Since treeview would (I presume) be more of a programatically set thing - I'm assuming you can't just sort by any arbitrary column then turn that into a treeview - so it would probably be a much smaller code base```

PaulL1:
Would you see the expanded sections having their own data sources, or all the data returned as part of one server call?  The latter would be much more efficient, if we did the former we'd need a "get children of this node" type method, and we'd need to original data to tell us which nodes had children under them.  Seems like you'd be making your server a lot more complicated with not much benefit?

PaulL1:
I've started on this: https://github.com/PaulL1/ng-grid/tree/treeView

It doesn't run yet, but the basics are there.  I haven't done any work on dynamic loading.  I need to add a tutorial.  But just so we don't double up, I'm posting.

PaulL1:
I've completed the base implementation of this.  It runs, has unit tests, has a tutorial and supports dynamic loading.  It has some outstandings that we might want to implement as noted in the tutorial: http://ui-grid.info/docs/#/tutorial/215_treeView

PaulL1:
[...]

PR #3259, opened 2015-04-11
Tree view

PaulL1:
Similar to grouping and expandable in some respects, this feature allows
you to have expandable rows within the same grid, with the same template,
and use some of your data rows as parent nodes, and others as children.

Refer #3246

### DIFF
--- a/misc/tutorial/121_grid_menu.ngdoc
+++ b/misc/tutorial/121_grid_menu.ngdoc
@@ -1,5 +1,5 @@
 @ngdoc overview
-@name Tutorial: 304 Grid Menu
+@name Tutorial: 121 Grid Menu
 @description The grid menu can be enabled through setting the gridOption `enableGridMenu`.  This 
 adds a settings icon in the top right of the grid, which floats above the column header.  The
 menu by default gives access to show/hide columns, but can be customised to show additional
--- a/misc/tutorial/209_grouping.ngdoc
+++ b/misc/tutorial/209_grouping.ngdoc
@@ -55,13 +55,12 @@ to allow people to start using it.  Notable outstandings are:
 Options to watch out for include:
 
 - `groupingIndent`: the expand buttons are indented by a number of pixels (default 10) as the grouping
-  level gets deeper.  Larger values look nicer, but mean that you probably need to make your groupHeader
-  wider if you're allowing deep grouping
-- `groupingRowHeaderWidth`: the width of the grouping row header, important as above
+  level gets deeper.  Larger values look nicer, but take up more space
+- `groupingRowHeaderWidth`: the base width of the grouping row header
 - `groupingSuppressAggregationText`: if your column has a cellFilter, the insertion of text (e.g. 'min: xxxx') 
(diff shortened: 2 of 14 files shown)


yccteam:
Looks great.
After reviewing here are my thoughts:
* Remove the ```+``` - since we already have ```rowProcessing```, why not add/remove it during the processor if the row has no visible children? Since we use virtualization, it shouldn't happen so often. Anything wrong with this point of view?
* Sorting - sorting should be for each level. This is where it should be like the ```expandable``` feature, so each subtree would be autonomous with its sorting as well as in its ```columnDef```.  In your current implementation it can run over the roots, sort them out, and within each root, sort the children (recoursively maybe?).  Maybe it should be done only for visible data (how can one access only the visible data?).
* Expand all - doesn't change the state of the header ```+```.
* Different headers for sub trees - as I mentioned, the subtrees can have totally different data sets. For instance - taxi stations with some data on them (location, name, number of taxis, area covered etc.) and taxis (liscensing plate and some more technical details) and for each taxi you can ask for a list of drivers (sometimes taxis have more than one driver) with totally different details needed to show. While it seems as if the expandable might fit in here, it doesn't in the case you want to use the tree as a menu to browse the data without showing scroll within a scroll or use pagination (to actually have the feel of a tree-view menu and not a grid).  You can argue that all expandable grids can be fully expanded but what if those grids have 10,000 data points?

Be happy to read your thoughts about this.

PaulL1:
Interesting thoughts.

1.  On the +. We can only remove it if we remove the ability to dynamically fetch the data, or if we provide some mechanism for the user to tell us whether or not there are children of a given node.  At the moment we need to let people click on it so we can raise the 'rowExpanded' event and let the user go get the data.

2. Sorting.  Yes, but it's hard.  The sorting algorithms just don't work like that, so it'd be a pretty substantial rewrite, probably treeView would have to replace the vanilla sort regime entirely with a recursive tree sort.  So whilst I don't disagree, it's not in my roadmap any time soon!!

3.  Expand all - sounds like a bug, I'll have to look at that.

4.  Different headers - I think this is a use case for expandable, not treeView.  The grid doesn't really work with different rowTemplates for each row.  It can be done I think, but it gives all sorts of problems with column headers and the like.  If this is what you're aiming to do then I think expandable is a better choice.

yccteam:
Thanks.
1. I see your point. The sign should come from the server, which knows if it has children or not. So a simple ```flag``` would suffice and it would be up to the user to handle it correctly.
2. Sorting - I already did something similar with global filtering of an expandable grid.  Shuoldn't be too hard I guess. I actually did that with ```lodash``` but I could make it totally vanilla JS so as not to include an external lib.  What function do I need to override in order to implement this?
3. :)
4. I've implemented an expandable grid, but there's the issue I've mentioned: when I get sub grids with a lot of data, I either need to give them a different scroll or pagination, which prevents the use of the grid as a tree-view menu. Any way around this? To use virtualization for the whole grid (including expanded rows' grids) with only the main grid's scroll?

PR #3266, opened 2015-04-12
A variety of fixes

### LINKED ISSUE #3236: guessSortFn throws error, getSortFn expects null return
Within the rowSorter, the getSortFn expects the guessSortFn to return null if it couldn't guess the correct sort function. Instead it throws an exception. Either guessSortFn should return a default function or the call should be wrapped in a try catch. 

Without this, one can't specify a custom colDef.type without breaking sorting.

PaulL1:
Fix #3236: doco on columnDef types
Fix footer: aggregation wasn't running on first load, changed the throttle to do one last call afterwards, changed throttle to $interval rather than $timeout to avoid breakage on e2e tests (protractor doesn't like $timeout)
Enh e2e tests: added a check for filter dropdowns
Fix e2e tests: fixed a variety of breakage in the e2e tests
Enh gridMenu: support for ordering on menu items, they were not deterministic and therefore sometimes in different orders
Enh saveState: allow saving of treeView state

### DIFF
--- a/misc/tutorial/103_filtering.ngdoc
+++ b/misc/tutorial/103_filtering.ngdoc
@@ -187,13 +187,13 @@ this button.
     var gridTestUtils = require('../../test/e2e/gridTestUtils.spec.js');
     
     describe('first grid on the page, filtered by male by default', function() {
-      it('grid should have six visible columns', function () {
-        gridTestUtils.expectHeaderColumnCount( 'grid1', 6 );
+      it('grid should have seven visible columns', function () {
+        gridTestUtils.expectHeaderColumnCount( 'grid1', 7 );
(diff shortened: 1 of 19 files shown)


Zorkling:
@PaulL1 the remove of .ui-grid-menu overflow:hidden from [Enh(gridMenu): support sort order on the menu items](https://github.com/PaulL1/ng-grid/commit/69117353218241d32f6997d285cf6d77941f53df) seems to break transitions applied to the menu

This plunker shows it in action. I have the overflow hidden css commented out in the main.css file
http://plnkr.co/edit/4ViiBrVvNoKUk4HsNuxm?p=preview

PR #3278, opened 2015-04-14
Adding a grid option for enabling/disabling column virtualization

jjkovacs:
Adding the ability to disable column virtualization. Column virtualization causes lag for columns that are expensive to rebuild (and where pinning is undesired). It also causes grids with a lot of/wide columns to have horizontal scrollbars that are jumpy and do not actually reflect the true width of the grid.

Closed out the pull request for #2969 because I don't think it was going to be merged anytime soon, and I didn't create a separate branch out of 2.x for it (whoops).

[Review on Reviewable](https://reviewable.io/reviews/angular-ui/ng-grid/3278)

### DIFF
--- a/src/classes/grid.js
+++ b/src/classes/grid.js
@@ -52,6 +52,9 @@ var ngGrid = function ($scope, options, sortService, domUtilityService, $filter,
         //Enable or disable HEAVY column virtualization. This turns off selection checkboxes and column pinning and is designed for spreadsheet-like data.
         enableColumnHeavyVirt: false,
 
+        //Enable or disable column virtualization.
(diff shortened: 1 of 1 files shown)


lotz:
I think this feature is definitely useful for situations where column construction is more expensive.  We have a need for it, and really do not want to maintain a separate patched "custom" version of ui-grid.

:+1:

mportuga:
@jjkovacs Could you add documentation for this? You can find instructions on how to do that here:
https://github.com/angular-ui/ui-grid/blob/master/DEVELOPER.md#documentation

mportuga:
Closing since we are no longer supporting the 2.x effort

Issue #1202, opened 2014-05-22
after resize column + http.get causes all column missing (col.width = NaN)


nghenglim:
Code that produces the issue as below, if colDef has no width specified, after resize the column and get the list again makes all column width becomes zero.

```
$scope.gridOptions = { 
    rowHeight: 28,
    headerRowHeight: 24,
    footerRowHeight: 24,
    data: 'records',
    columnDefs:'columnDefs', 
    showFilter:true,
    showFooter:true,
    enableRowSelection: true,
    enableColumnResize: true, //got a bug after resize and load data table empty
    enableHighlighting: true
};
$scope.filter = function() {
    $http.get('/someurl.php')
        .success(function (data) {
            $scope.records = data['result']['items'];
            $scope.columnDefs = data['result']['colDefs'];
        }).
        error(function(data) {
            console.log(data);
    });
};
```

the php file for get request

```
$colDefs = array();
$colDefs[] = array('field'=>'name','displayName'=>'Name');
```

PR #3279, opened 2015-04-14
Fixing column widths snapping back to their pre-resized widths

jjkovacs:
The fix for #1202 caused issues with column widths snapping back to their original values after calls to rebuildGrid.

Basically, what was happening was the columns were being lazy loaded. When resizing the columns, `hasUserChangedGridColumnWidths` gets set to true. When the grid received the new columnDefs, it destroyed the old columns collection, whose widths had already been calculated, and filled the columns collection with a new set of ngColumns. The new column widths were unable to be calculated because `hasUserChangedGridColumnWidths` was still set to true. 

Reverting the original fix for #1202, and instead setting `hasUserChangedGridColumnWidths` to false when building a new columns collection.

[Review on Reviewable](https://reviewable.io/reviews/angular-ui/ng-grid/3279)

### DIFF
--- a/src/classes/grid.js
+++ b/src/classes/grid.js
@@ -391,6 +391,7 @@ var ngGrid = function ($scope, options, sortService, domUtilityService, $filter,
                 cols.push(column);
             });
             $scope.columns = cols;
+            $scope.hasUserChangedGridColumnWidths = false;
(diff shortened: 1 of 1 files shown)


mportuga:
Closing since the issue in question is already closed.

Issue #1221, opened 2014-06-02
Resizing column into left direction will not apply to last column data


konstantinkrassmann:
If i resize the column to the left until the last column does move, the last column's won't enlarge to meet the right border. 

Plunker: 

http://plnkr.co/edit/FiNAPALZph53t0mmr0fL?p=preview

Just resize 2nd column to the left. 
Check the last column data.

Image: 

http://s7.directupload.net/images/140602/5txafjjn.png

PR #3280, opened 2015-04-14
reverting the fix for #1221

### LINKED ISSUE #3044: ngGrid- version 2.x - Horizontal scrollbar appears even if column widths fit the viewport.
When grid has vertical bar displayed, horizontal scroll is always visible even column widths fit the viewport.  This plunker shows the example of issue:
http://plnkr.co/edit/ZllYpsC815Rp0UrdznlA?p=preview

Look at the code:
var rightPad = 0;
                if ((i === cols.length - 1) && (sumWidth + col.width < grid.elementDims.rootMaxW)) {
                    rightPad = grid.elementDims.rootMaxW - sumWidth - col.width;
                }
                css += "." + gridId + " .col" + i + " { width: " + (col.width + rightPad) + "px; left: " + sumWidth + "px; height: " + rowHeight + "px }" +
                    "." + gridId + " .colt" + i + " { width: " + (col.width + rightPad) + "px; }";

rightPad doesn't take scrollH into account.  Therefore the column with is a little bit over viewport.

Seems a similar scrollbar issue was reported in version 3.0 and fixed.

jjkovacs:
The fix for this issue seems to have introduced various other issues. I've noticed that the column headers sometimes get out of alignment with their corresponding columns, especially when a grid contains pinned columns. This should also fix #3044.

Is the issue expressed in #1221 expected to be base behavior of a data grid? I wouldn't expect the width of the last column in a grid to be dependent on the width of the other columns by default.

If the expectation is that the row striping extends all the way across the grid, that can be done purely with css:
```css
.gridStyle .ngCanvas {
  min-width: 100%;
}

.gridStyle .ngRow {
  min-width: 100%;
}
```

The last column can also be set to fill the remaining width of the row with css as well:
```css
.gridStyle .ngRow .ngCell:last-child {
  width: auto !important;
  right: 0px;
}

.gridStyle .ngRow .ngCell:last-child .ngCellText {
  width: inherit;
}
```

The following plunker demonstrates this:
http://plnkr.co/edit/lqMtOz7DzPC9t9zmBRbX?p=preview

[Review on Reviewable](https://reviewable.io/reviews/angular-ui/ng-grid/3280)

### DIFF
--- a/src/services/DomUtilityService.js
+++ b/src/services/DomUtilityService.js
@@ -126,12 +126,8 @@
         for (var i = 0; i < cols.length; i++) {
             var col = cols[i];
             if (col.visible !== false) {
-                var rightPad = 0;
-                if ((i === cols.length - 1) && (sumWidth + col.width < grid.elementDims.rootMaxW)) {
-                    rightPad = grid.elementDims.rootMaxW - sumWidth - col.width;
-                }
-                css += "." + gridId + " .col" + i + " { width: " + (col.width + rightPad) + "px; left: " + sumWidth + "px; height: " + rowHeight + "px }" +
-                    "." + gridId + " .colt" + i + " { width: " + (col.width + rightPad) + "px; }";
+                css += "." + gridId + " .col" + i + " { width: " + col.width + "px; left: " + sumWidth + "px; height: " + rowHeight + "px }" +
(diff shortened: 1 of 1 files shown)


mportuga:
The issue in question appears to have been closed already. Closing this PR.

Issue #2646, opened 2015-01-27
CSV Export


gunjanr17:
What versions of .xls and .xlsx the current csv export is compatible with?

PaulL1:
None - it exports in csv format, which excel is capable of importing (and on most machines where excel is installed, will open in excel automatically)

cookednoodles:
However, in some european countries it won't open correctly. Accents and special characters will be messed up in excel, because of utf-8 not being set. Excel will also want a semi-colon as a delimiter, not a comma.

gunjanr17:
I think there was already a issue logged for adding UTF-8 support to CSV which is closed, so I suppose we already have UTF-8  being set. 
Does CSV export consider the locale settings for delimiter?

PaulL1:
There is an option to override the delimiter, and I thought the character set got changed.  However, when we redid the download link we may have lost that.

cookednoodles:
It seems it was removed again in ttps://github.com/angular-ui/ng-grid/commit/aad0bd6156fcf3b8930f4291fc1588e7864eea4c I have no idea on the best way to do this as it appears that each application behaves differently.

PaulL1:
I did the change in that commit, and at the time I did it that first character was turning up in the CSV import as an unprintable in the first cell of the CSV.  Taking it out fixed the problem.  I think I worked out that the "\ufeff" was just another way to set character set, and the change to the download link in #2931 I added `var strMimeType = 'application/octet-stream;charset=utf-8';`, which also sets the character set.  The two I think were conflicting.

So I think that we're still set to utf-8.

PaulL1:
I think excel is quite happy with a comma as delimiter, as that is the csv specification.  Is anyone using rc19 and can confirm that it is or is not utf-8?

c0bra:
`file -bi filename` reports "text/plain; charset=us-ascii", but that's after downloading on Windows and copying over Samba to my linux machine. There's no really set-in-stone way to tell the correct encoding of a file.

I'm going to go ahead and close this. If the encoding or something else is still a problem then a new issue can be opened.

Cog-g:
I can confirm that I have the same problem of encoding. And reversing https://github.com/angular-ui/ng-grid/commit/aad0bd6156fcf3b8930f4291fc1588e7864eea4c resolve this problem.

PaulL1:
[...]

PR #3290, opened 2015-04-16
exporterOlderExcelCompatibility fix

ravishivt:
The exporterOlderExcelCompatibility option wasn’t being used to determine whether to prefix the CSV file with "\uFEFF" or not.  See #2646 for more info.


PaulL1:
Looks good, thanks.

PR #3301, opened 2015-04-17
Pagination tests working again

### LINKED ISSUE #3173: Fix pagination tests
The file `pagination.spec.js` had a comma in the name. It used to be: `pagination.spec,js`, which meant karma was not picking it up for inclusion.

I have fixed the extension but the tests are broken, so I have disabled them.

@derrabus This is probably something you need to look at, correct?

derrabus:
Fixes #3173.


PaulL1:
Excellent, thanks.

Issue #3184, opened 2015-04-01
Lots of blank space past last row when infinitely scrolling


davidkaminsky:
I'm seeing a problem that I'm assuming is related to the fix for this issue (https://github.com/angular-ui/ng-grid/issues/3119). When I've reached the end of my results when infinitely scrolling the scrollbar is still in the middle of the viewport and I see a lot of empty space below the entries.

I am telling dataLoaded that it should no longer be able to scroll and that seems to be working as I'm not seeing anymore calls to get more data.

vm.gridApi.infiniteScroll.dataLoaded(false, (page * numResultsPerPage) < result.numTotalUsers);

![screen shot 2015-03-30 at 2 38 56 pm](https://cloud.githubusercontent.com/assets/1380808/6945045/e9ef02f6-d84b-11e4-9800-30c413030d9f.png)

luciformed:
I  have the same issue, and based on my investigation it's because of the way viewport offsets the rows from the top

davidkaminsky:
I'm still seeing this issue on the stable version rc.21. Is this supposed to be fixed?

PaulL1:
This looks to have been fixed on 18th April.  Are people still experiencing it?  If so, could someone please provide a plunker so that it can be diagnosed?

c0bra:
No response in > 2 weeks. Closing.

valtersboze:
i got the same problem when i was trying to remove the weird 1px padding-top on ui-grid-canvas. so after i added "padding: 0"  i got same empty block after the content making the scrollable content 2x longer.. if anyone experiencing the same, remove anything in custom css related to ui-grid-canvas and that should fix the problem.

PR #3304, opened 2015-04-17
Fix for extra-whitespace issue in the viewport #3184

luciformed:
issue #3184

### DIFF
--- a/src/js/core/directives/ui-grid-viewport.js
+++ b/src/js/core/directives/ui-grid-viewport.js
@@ -28,7 +28,7 @@
           $scope.rowContainer = containerCtrl.rowContainer;
           $scope.colContainer = containerCtrl.colContainer;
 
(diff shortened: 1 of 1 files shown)


c0bra:
You're going to have to give some actual information with this PR. Changing from margin-top to translateY is fairly significant and I'd like to hear how this solves the problem and why using margin-based positioning doesn't work.

I wouldn't also want to make sure that this does not negatively affect performance, esp. on mobile devices.

luciformed:
The problem is, for whatever reason, when offset is done via margin-top, scrollHeight of grid viewport becomes canvas height + that margin-top value , hence the additional whitespace at the end of the list. I will try to find out why exactly that happens with my grid and not in ui-grid example pages

luciformed:
Ok, the thing i was missing, is 'padding-top:1px' on canvas element.

Issue #3281, opened 2015-04-14
'!important' in cell.less & selection.less


Zorkling:
The !important in cell.less for the checkbox column seems unnecessary and is causing some difficulty for me while trying to create style overrides. 

Changing the selector for 'ui-grid-row-header-cell' to '.ui-grid-row .ui-grid-cell.ui-grid-row-header-cell' should allow it to take precedence over the stripped row colors without the need for the !important.

Additionally in selection/less/selection.less the selector for a selected row could be modified from '.ui-grid-row-selected > [ui-grid-row] > .ui-grid-cell' to '.ui-grid-row.ui-grid-row-selected > [ui-grid-row] > .ui-grid-cell' to accomplish the same goal.

Please let me know if theres any considerations i missed that would make this change problematic and i will look into it for a more complete solution. If possible I would like to use the default css so that I am able to keep up to date with the latest version and not have compatibility issues, rather than using the customizer or compiling the less myself.

Fantastic job with the 3.0 update, and im looking forward to the release!

(using v3.0.0-rc.20-b0dc47d)

c0bra:
Thanks for the report. I'll look at updating this. If you were wanting to have a go at a PR it would be most welcome.

Zorkling:
Ill give it a shot, and let you know if im able to make progress on it. Thanks for the response!

Zorkling:
@c0bra Completed the changes and made a pull request. Let me know if it looks good and if theres something I can improve upon with my PR process!

Zorkling:
Thanks @c0bra !

PR #3312, opened 2015-04-19
Fix(less): removed importants in favor of higher priority selector #3281

Zorkling:
Fix for [#3281](https://github.com/angular-ui/ng-grid/issues/3281)

'!important' in cell.less & selection.less

In addition to updating the css selectors and removing !importants, I added a less variable for the .ui-grid-row-header-cell background color


c0bra:
Perfect, thank you!

PR #3315, opened 2015-04-19
Fix(i18nService) : Minor fr translation fixes

Oliv:
Hey there, 

here are some better translated i18n for french language

Regards, Olivier


PaulL1:
Very good, thanks.

PR #3321, opened 2015-04-20
Add Grid.prototype.clearAllFilters() method.

amyboyd:
I had a use-case where I needed a button that clears all filters applied to all columns. This pull request adds a method to the Grid class, `clearAllFilters()`, which does that.

### DIFF
--- a/src/js/core/factories/Grid.js
+++ b/src/js/core/factories/Grid.js
@@ -417,7 +417,19 @@ angular.module('ui.grid')
      * 
      */
     self.api.registerMethod( 'core', 'notifyDataChange', this.notifyDataChange );
-    
+
+    /**
+     * @ngdoc method
+     * @name clearAllFilters
+     * @methodOf ui.grid.core.api:PublicApi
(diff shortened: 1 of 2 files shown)


PaulL1:
This looks good.  A few comments:
1.  The test has a header of sorting - copy paste?
2.  You're clearing the condition and the flags as well.  The filter should become inactive when you remove the term, perhaps people would rather retain the condition?  Should this behaviour maybe be the default or at least switchable? 
3.  It seems you'd explicitly intend to have people use this on the api, perhaps it would make sense to add:
  `    self.api.registerMethod( 'core', 'clearAllFilters', this.clearAllFilters );`
to the top of the grid.js?

amyboyd:
Thanks for the feedback @PaulL1. I've changed the method signature to `Grid.prototype.clearAllFilters = function(refreshRows = true, clearConditions = false, clearFlags = false)`, added the `registerMethod` call, and fixed the copy-paste mistake. The method has 100% unit test coverage too.

PaulL1:
Very tidy.  Thanks for the contribution.

PR #3353, opened 2015-04-23
Removing the gap between the bottom of the viewport, and the bottom of the grid

jjkovacs:
There is a 2px gap between the bottom of the viewport, and the bottom of
the grid. The high contrast colors in the following plunker shows this:

http://plnkr.co/edit/J44cYftFauaZ5LPY4aOM?p=preview

This changeset removes this gap.

[Review on Reviewable](https://reviewable.io/reviews/angular-ui/ng-grid/3353)

### DIFF
--- a/src/classes/grid.js
+++ b/src/classes/grid.js
@@ -856,7 +856,7 @@ var ngGrid = function ($scope, options, sortService, domUtilityService, $filter,
     };
 
     $scope.viewportDimHeight = function() {
-        return Math.max(0, self.rootDim.outerHeight - $scope.topPanelHeight() - $scope.footerRowHeight - 2);
+        return Math.max(0, self.rootDim.outerHeight - $scope.topPanelHeight() - $scope.footerRowHeight);
     };
     $scope.groupBy = function (col) {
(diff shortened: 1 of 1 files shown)


c0bra:
Are you sure those two pixels weren't there for a reason?

jjkovacs:
I viewed history to see when/why they might have been added, and they've been there since the early days of when the template style functions were added.

Admittedly, I don't know what the use case was for it being added. I couldn't think of a reason they should be there, except for maybe if someone added it to offset top/bottom borders they might have added to the viewport. Not sure if that justifies sizing the viewport this way though, especially considering the examples on the demo page don't have borders on the viewport.

I've done some testing with it, and it seems to still be working as expected. We can hold off on this though if the community thinks otherwise.

mportuga:
@dlgski @JLLeitschuh This seems harmless enough, do you think we should merge this?

JLLeitschuh:
This is aimed at 2.x which we haven't been maintaining.
Should we merge? Sure?

mportuga:
Closing since 2.x is not longer being maintained.

Issue #2958, opened 2015-03-09
row index


pjlasl:
I am trying to display the row index in a column and I found an example that works using ng-grid, but it produces the same number "1" for every row it adds to the grid.

Using the ng-grid, I could write the following:
{name: 'ID', field: '', displayName: 'Row Number', cellTemplate: '<div>{{row.rowIndex + 1}}</div>'}

The ui-grid does not produce the same result. How can I get the same behaviour with the ui-grid?

PaulL1:
When you say you are trying to display the row index, do you mean you want the actual row index, or that you want to display a sequential id in all your rows?

If the latter, then you can do it by just adding a counter column to your data:
```
  angular.forEach($scope.myData, function( row, index){
    row.sequence = index;
  });
```

If you want to show the index of the row within the grid internals, then it depends on which internal you want.  You can get the index of the row within grid.rows, which would show the row as it stands in the original rows list (not filtered nor sorted), or the index of the row within grid.renderContainers.body.visibleRowCache (filtered and sorted), or the render index of the row within the currently displayed rows (given virtualisation, this is generally a particularly useless number).

If you're OK that whenever someone sorts or filters then the numbers will change, then you could do it with a cellTemplate, which would be something like:
```
  cellTemplate: '<div>{{row.grid.renderContainers.body.visibleRowsCache.indexOf(row)}}</div>
```

pjlasl:
Thank you for the reply. I am looking to add a sequential number (1,2,3...) to a column in the grid. 
I am very new to this control so any detail you could provide would be great. I tried to use the following (in my columnDefs) but all i get is an empty cell

```javascript
{name: 'lineNbr', displayName: '',width: 30, cellTemplate: '<div>{{row.grid.renderContainers.body.visibleRowsCache.indexOf(row)}}</div>' },
```

pjlasl:
This was not meant to be closed....

PaulL1:
Try this: http://plnkr.co/edit/9QbQCqTSz8aluh9n62Hl?p=preview

pjlasl:
Thanks Paul. That worked better! Appreciate the quick responses. I really like this grid and it has been fun learning how it works!

Awakening-j:
@PaulL1 The row index working for me at cellTemplate, but I wonder will it cause performance issue? It is so complex to get index  
[...]

PR #3387, opened 2015-04-30
make rowRenderIndex available to GridRow

tx-tim:
I am looking to do something similar to issue #2958 but I want the rowIndex to be made available during the compile phase.
If there's a better way to do this I am 100% open to suggestions.

### DIFF
--- a/src/js/core/directives/ui-grid-row.js
+++ b/src/js/core/directives/ui-grid-row.js
@@ -25,6 +25,9 @@
 
             // Function for attaching the template to this scope
             var clonedElement, cloneScope;
+            Object.getPrototypeOf($scope.row).rowIndex = null;
+            $scope.row.rowIndex = $scope.rowRenderIndex;
+            
             function compileTemplate() {
               $scope.row.getRowTemplateFn.then(function (compiledElementFn) {
                 // var compiledElementFn = $scope.row.compiledElementFn;


PaulL1:
I haven't looked all the way through the code, but I've seen lots of problems in the past with render indexes because of row virtualisation.  Are you across that, and are you sure renderIndex is the right thing to have access to v's the index in the visibleRowsCache or something else?

tx-tim:
Well, another way would be to pass $index as part of the isolate scope from the renderedRows repeater in uiGridViewport.html.  I don't really see what that would buy you because it's only iterating over renderedRows.  Something like so:

    <div ng-repeat="(rowRenderIndex, row) in rowContainer.renderedRows track by $index" class="ui-grid-row" ng-style="Viewport.rowStyle(rowRenderIndex)">
      <div ui-grid-row="row" row-render-index="rowRenderIndex" row-actual-index=$index></div>
    </div>

What do you think?

PaulL1:
It depends what you're trying to achieve.  As per the issue you reference, and the FAQ, there are a few different things an index can be:

 - a number we allocate to the data that never changes even when we sort/filter.  In that case, we want the position in the data array
 - a number we display on the screen sort of like the row number in Excel - when we resort it changes, but when we filter we get gaps in the number.  This we don't really have - we don't have a "all rows sorted but not filtered" array anywhere
 - a number that is just the number in the currently visible rows, in the current order they're showing on the screen.  This would be the position in the visibleRows cache
 - a number that is the render index of the currently rendered DOM elements (so if we have 1000 rows and we're sort of half way through, then our renderIndex might be 7 or 8 if we were the 4th row down the screen - there's usually 4 rows rendered off-screen).  In other words - this number is usually completely useless for anything

If we're going to push it into the base we need to be really clear what the renderIndex actually is, and what it can safely be used for.  I've seen a number of people tripped up by putting one in and then discovering it never goes higher than about 20, because we usually don't render more than about 20 rows.

tx-tim:
You are so right!  Which is probably why the functionality doesn't already
exist.  Ok - thanks for talking this over with me - I'll look at the code
some more.

Issue #3397, opened 2015-04-30
For Save/Restore State, what do we really want to save/restore?


ravishivt:
This is more of a discussion that hopefully will lead to some code changes.  I'm using save / restore state features to save various table states to a users profile.  The user can reload the app and still access and reload their saved states.  This leads to a problem because the saveColumns function simply does an angular.copy() on each column's "filters" object, which will attempt to save the filter's condition function.  This works fine until you reload the app.  On page reload, the closures for those functions are no longer valid so restoring that table state leads to a broken column filter.

I was thinking of creating a pull request for just saving and restoring each column filter's "term" but I think that may be too naive.  Are their other things that should be saved for each filter?  My thought is that we only save "term" since that is generally the only thing the user can modify during runtime and is not purely a config option.

PaulL1:
The editors are customisable, and some editors that people might right would use things other than the term (for example, one of my editors creates a regEx and puts it in the condition, which lets me do an OR - it aggregates items selected from a multi-select dropdown).

I'm not quite clear on why on page reload anything would remain - surely the entire grid gets rebuilt, and anything from the past is overwritten?

ravishivt:
To better explain the problem, the current implementation is to do an angular.copy() on the entire filters object for each column.  This copies the condition function for each filter with all the closures and object references it uses at that runtime instance.  When the page reloads, those references are no longer available.  So initially the table gets rebuilt and everything works fine as you stated.  But when the user restores their previously saved table state, the filter's condition function gets overwritten from the saved state and it breaks the filtering for that column.

I just created a pull request for my original thought.  After reading your comment about using other filter properties, maybe it would be better to remove only the condition function instead of only saving the "term" property.

tbaeg:
I realized this is a closed issue, but I personally don't feel this is the correct solution. If you're saving state, you should be persisting it in some form. It should not be the libraries duty to handle that sort of situation. If I attach a property to an object, I expect it to be there when I call for it. It doesn't make sense to remove the condition functions because of this one edge case. It completely breaks expected results from an API perspective. 

I came across this myself when writing custom filters and found it to be extremely frustrating to deal with when others are actually trying to persist conditions.

https://github.com/angular-ui/ui-grid/issues/5250

Above is another issue explaining the same issue.

dKab:
Omg, why would you leave condition? I use custom templates for filtering and user can chose not only term but also a condition. Now I have to reimplement the whole saving of filters...

PR #3398, opened 2015-04-30
For save/restore state, only save/restore each filter's term prop.

ravishivt:
See #3397 for more info.


PaulL1:
I think we want to store more than just the term - we allow custom filters which could add other properties to the filter object.

PR #3407, opened 2015-05-01
Fixed to handle undefined field when grouping

norsang:
Fix: Commit # 24155ad breaks group by field if the field is missing in any of the row. This is because of val.toString() which doesn't work on undefined.

http://plnkr.co/edit/9TaFbiAIsVSAObPpcgO4?p=preview

Should be changed to:

<pre>val = (val === "" || val === null || val === undefined) ? 'null' : val.toString();</pre>


PaulL1:
Thanks

Issue #3394, opened 2015-04-30
Misalignment between the grid content and the select checkbox column


clalag:
![schermata 2015-04-30 alle 13 31 00](https://cloud.githubusercontent.com/assets/7940912/7412687/2ed0d1c6-ef3d-11e4-9e78-9e0f5c9a1e8e.png)

you can see the live demo here http://ui-grid.info/docs/#/tutorial/401_AllFeatures

AgDude:
I just ran into this also, but updating to RC21 solved it.

JayHuang:
I have this issue also. Only surfaced since updating to RC21, and the demo has the same issue; it's using RC21 too.

albrethandoyo:
how can fix this?

AgDude:
I tried reproducing this in a plunkr, and cannot do so consistently. I got the issue to appear once, but can't get it to show up again after refreshing. Will someone please provide a provide a plunkr which demonstrates the problem (please specify your browser version and platform also).

JayHuang:
This seems to be the same bug detailed in #3412

AgDude:
@JayHuang, I believe this is not the same bug as #3412. I have not looked in more detail at where this is coming generated, but I just observed this again in the grouping tutorial page.

The problem has to do with the css which is generated and inserted into a style element. This is the offending line:

    grid1430839489202 .ui-grid-render-container-left .ui-grid-header-canvas { height: inherit; }

On a grid which is displaying properly, this selector has height set to a pixel value.

anwarhamr:
Would this be related?  I have a grid when we resize or sort any of the colums the Select all checkbox collapses moving up the other checkboxs so they are not aligned correctly w/ the row.  Only seems to be happening when enableSelectAll set to false.

Grid config:
$scope.gridOptions = {
        data: 'discoveredMachines',
        columnDefs: [
            {
                field: 'MachineMacAddress',
                name: 'Mac Address',
                displayName: 'MAC Address',
                width: "*"
            },
            {
                field: 'IpV4Address',
                name: 'Ip Address',
                displayName: 'IP Address',
                width: "*"
            }
        ],
        headerRowHeight: 36,
        enableColumnResizing: true,
        enableHeaderRowSelection: true,
        enableSelectAll: false,
        multiSelect: false,
        onRegisterApi: function (gridApi) {
            //set gridApi on scope
            $scope.gridApi = gridApi;

            //allow the click of a row to select the row.
[...]

Issue #3409, opened 2015-05-03
ui-grid Grouping is broken


deepfriedbrain:
In the tutorial example for Grouping for ui-grid, I noticed that the grouping is broken. The +/- signs are placed one row above their correct position after expansion. See the 2 screenshots of original un-expanded state and expanded state.

![image](https://cloud.githubusercontent.com/assets/1531669/7444159/98005872-f1ab-11e4-8277-f8ec7f6d381c.png)

![image](https://cloud.githubusercontent.com/assets/1531669/7444161/a4dabd94-f1ab-11e4-8d47-4c75edc42766.png)

AgDude:
This is the same problem described in #3394. Thanks for the details, on when it occurred.

PaulL1:
Closing as duplicate.

springzen:
Build used: ui-grid - v3.1.1 - 2016-02-09
I know this is now closed, but I am not sure if this is the same issue or a new one. When I group, I am seeing the same issue as described originally. In my case this is related to my css. I was trying to add word wrap and I took the tip from https://github.com/angular-ui/ui-grid/issues/317.
```
.ui-grid-viewport .ui-grid-cell-contents {
    word-wrap: break-word;
    white-space: normal !important;
}

.ui-grid-row, .ui-grid-cell {
    height: auto !important;
}

.ui-grid-row div[role=row] {
    display: flex;
    align-content: stretch;
}
```

If I use this css, only the headers wrap.
```
.ui-grid-header-cell .ui-grid-cell-contents {
    /*height: 48px;*/
    white-space: normal;
    -ms-text-overflow: clip;
    -o-text-overflow: clip;
    text-overflow: clip;
    overflow: visible;
    word-wrap: break-word;
}
```

![image](https://cloud.githubusercontent.com/assets/2808006/13235795/cae574e6-d974-11e5-8430-4fc15f6ff410.png)

Issue #3382, opened 2015-04-29
ui-grid-header-canvas bad height in v3.0.0-rc.21 on TreeView


jibap:
Hi, 

I use ui-grid in TreeView, in v3.0.0-rc.21-43f63ac
When i expand a node, the tree view row header height become bad (except first time).

![image](https://cloud.githubusercontent.com/assets/3915029/7390637/6e9a318a-ee7a-11e4-8156-24bf2740c568.png)

I didn't have this issue in my previous version (Apr 24, 2015).

I exploit this issue to ask you another : i use externals filters and i want to edit filters dynamically.
I am able to edit them (visually) but they are not watched by filterChanged event ( i read a lot of issue talking about it, and i anderstand the reason for no watching all - explained in tutorial 316 ). But how to rebuild(redraw, refresh ??) the grid with my new options ? 

![image](https://cloud.githubusercontent.com/assets/3915029/7390876/5aa84b24-ee7c-11e4-9f03-be4abb733444.png)

About first issue, you can look in second picture when i add a filter, i use "notifyDataChange", and the tree view row header is recalculated good. So i tried to call it "on.rowExpanded" but it is not working....

Sorry for my english. 

Thanks,

meyerds:
I'm experiencing the same bug with grouping. I believe it was caused by this commit: 43f63ac9f62a133b51d7af2f4a966318eae8e9ac - it works in the rc21 "stable" release but not in the latest unstable.

AgDude:
I think this is a duplicate of #3394

PaulL1:
I think the height portion of this is resolved, can you please confirm @jibap?

On the filters, it looks like your filters are using the ui-grid native filter logic, but you just want to apply the filtering using externalFiltering?  If this is correct, then you should be getting a filterChanged event when the filter is edited - there is a watch on any filter box that ui-grid displays.  The area we don't watch is if you programatically change a filter that is on a hidden column.

Do the filters that you use have a custom directive?  Or are they entirely native?  Could you provide a plunker and some instructions to recreate the issue?

jibap:
Hi @PaulL1, 

I try with v3.0.0-rc.21-869be69, but the issue still bad...

On the filters, sorry i forgot to precise, indeed i want to change filters programatically. The concept is having a "+" button which create a new filter row like this :
![image](https://cloud.githubusercontent.com/assets/3915029/7665618/cf7f47da-fbc0-11e4-81e3-c29e0c81a36f.png)
[...]

PR #3424, opened 2015-05-05
fix(uiGridHeader): ensure that styles are rebuilt on explicit height

AgDude:
Refactor setting of header heights to be more DRY and ensure that rebuildStyles
is set when an explicit height is set. Do not set existing explicit heights to
 null before recalculating height.

 Fixes: #3394, #3382, #3409

### DIFF
--- a/src/js/core/factories/Grid.js
+++ b/src/js/core/factories/Grid.js
@@ -2065,8 +2065,8 @@ angular.module('ui.grid')
         }
 
         if (container.header || container.headerCanvas) {
-          container.explicitHeaderHeight = null;
-          container.explicitHeaderCanvasHeight = null;
+          container.explicitHeaderHeight = container.explicitHeaderHeight || null;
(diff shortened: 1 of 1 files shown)


AgDude:
Note that this has not been validated in this case (as described by @c0bra): "If you have two render containers and one had filters that were on that you then turn off, they other header beds to know to shrink"

Is there a tutorial or other example which already has this scenario?

AgDude:
This does appears to work with toggling filters. Here is a demo: http://plnkr.co/edit/SmDX8lepvY2ItQOQvH4o?p=info

c0bra:
I will try to take a look when I can but I'm out of town till late next week

AgDude:
No problem we were just having some discussion on gitter about this issue.

c0bra:
Looks great, @AgDude. Sorry for taking such a long time to get this merged.

Issue #3395, opened 2015-04-30
when ui.grid.moveColumn and ui.grid.selection are used together, column header is flying too fast until end, same issue for pinning


moguzalp:
When column move and row selection are used together, when I try to move column, column header starts to flying from left to right too fast:
http://plnkr.co/edit/iU6weT394sSMVHO03tYT?p=preview

the same for pinning: 
http://plnkr.co/edit/LtMNx3zkOEgAJ9Oq3SD4?p=preview

moguzalp:
It seems issue is resolved with the "ui-grid - v3.0.0-rc.21-2ad8f7f - 2015-06-03" version. Thanks

moguzalp:
I closed issue. It works ok for colums however it has still problem when horizontal scrolling and after scroll right while moving those columns.

AgDude:
@moguzalp I am closing this issue as I think the original problem of "flying to the right" has been resolved. There is currently a bug with moving a column after scrolling to the right #3631, but I am not able to reproduce the "flying to right".  Please re-open or comment if I am incorrect.

bvsbharat:
hi is this issue is fixed with ui-grid.js file or we need any manual coding

PR #3426, opened 2015-05-05
fix(moveColumn): account for width of left container

### LINKED ISSUE #3417: column moving and selection don't work together (RC21)
Here is a plunker:
http://plnkr.co/edit/HIFDqPgfi2fftTV0ekrd?p=preview

The problem seems to be in src\features\move-columns\js\column-movable.js in the ```moveElement``` function. The ```newElementLeft``` is not calculated properly because the 'grid-render-container'  now has a margin of 30px to leave space for the selection checkboxes.

AgDude:
account for width of left container when positioning moving column header.

 Fixes #3417, #3395


PaulL1:
Thanks @AgDude, there's still a few other move column issues out there if you want to take a look?

PR #3429, opened 2015-05-06
fix(clearThisRowInvisible) not using parameter reason

valepu:
Not sure if intended but clearThisRowInvisible is not using the parameter "reason" when deleting the "invisible reason" but it always deletes the hardcoded parameter "user". 
This shouldn't change any other behavior since ui-grid internal functions call this function only with the parameter "user" but it will fix any unexpected behavior when using this function with a custom invisible reason

Haven't put a check if the value exists since "delete" fails silently in that case


PaulL1:
Thanks

PR #3446, opened 2015-05-07
pass context to getCellValue function expressions, and fix #3435

### LINKED ISSUE #3435: [Grouping] Need support for multiple instances of column
We should be able to apply a different aggregate function to different instances of the column. It seems that the aggregation function from the last column is applied to all column using that field.

See this plunk: http://plnkr.co/edit/MahC67?p=info

AgDude:
functions expressions parsed by the getCellValue method will now receive an argument which is an object with both the current row and column This should not be breaking, since previously it was impossible to pass arguments to those functions.

The fix to grouping features is dependent on the new getCellValue method.

### DIFF
--- a/misc/tutorial/106_binding.ngdoc
+++ b/misc/tutorial/106_binding.ngdoc
@@ -5,12 +5,9 @@
 UI-Grid supports complex object binding in the colDef.field.
 
 This tutorial shows two-way binding to properties with special characters, array element, nested property, and function.
+Angular expressions will have 'row' and 'col' objects available on scope within the grid cell.
 
-Note that a function can not be edited. Functions will receive a `context` object, with access to both the current row and column, as the only argument.
-    <pre>
-      {row: GridRow,
-      column: GridColumn}
-    </pre>
+Note that a function can not be edited.
 
 In your custom cellTemplates, you can use:
 <br/>
@@ -41,8 +38,8 @@ MODEL\_COL\_FIELD which will be replaced with row.entity.field.  Use MODEL\_COL\
                                "friends": ["friend0"],
                                "address": {street:"301 Dove Ave", city:"Laurel", zip:"39565"},
                                "getZip" : function() {return this.address.zip;},
-                               "getTown" : function(context) {
-                                 return context.col.displayName + ': '+ this.address.city + ', ' + country;
+                               "getTown" : function(col) {
+                                 return col.displayName + ': '+ this.address.city + ', ' + country;
                                }
                            }
                        ]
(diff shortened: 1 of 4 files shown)


AgDude:
Put this on hold please. I have discovered it is causing problems if the grouping directive is applied but disabled. I should have a fix soon.

swalters:
on a side note, GetCellValue is called a bizillion times.  I'd rather see this done without creating 2 new javascript objects on each call:
{context: {row: row, col:col}}

Can we just call with:
return col.cellValueGetterCache(row, row, col);

AgDude:
Unfortunately we cannot pass the mutliple args as in your example. As far as I can tell from my experiments and the angularjs docs $parsed functions take only two arguments: ```context```, and ```locals```, which must be an object.

We could pass the column object only, and say that if you want access to the row use ```this```.

AgDude:
I did a little reading of the angularjs code this morning, and the second argument when a $parsed function is called must be an object, and it is treated like a $scope (the provided args must match the object properties). Given this, I can easily make this one new object instead of two.

If we could determine when this was a function, and when it was an object property, we could only pass the extra context to functions, but I don't see an easy way to do that. $parse adds a ```constant``` property to identify constants, but doesn't have anything for differentiating objects and functions. We might be able to use some sort of hack like "ends with ```)```".

PaulL1:
I think there may be a different way to do this, and I intend to look at it.  getCellValue is a huge proportion of our cost on every digest cycle, and I'm very keen to avoid any change in it's performance at all - the most I'd be OK with really is a single if (and anything new hidden inside that if)

AgDude:
I just pushed the version I had locally with only a single object created. I like your idea of a conditional, Paul. I had played around with that in trying to determine when the context was needed in an efficient way, and didn't come up with a good solution. It didn't occur to me until reading your comment that an option saved on either the row or the column could accomplish the same thing.

My current thought is to put a flag on the row indicating we need the context. Then in grouping we only get that extra cost on the header rows. We could provide a top-level option to always put that on the row if someone wanted that context and didn't care about the performance hit.

You understand the core issue, so if you have a completely different approach that works, great!

PaulL1:
I think I'll handle this a different way in the code - I'm suggesting adding new values to the entity, inside an object named `entity.$$col.uid`.  getCellValue will check for existence of this object and use it if it's there - so effectively this masks the field value on the entity.  This also allows for people to write code to get at the object, and the individual fields on the object - so they can get at:

```
row.entity['$$' + col.uid] = {
  value: xx,
  label: 'sum: ',
  rendered: 'sum: xx',
  < could be custom values if using a custom aggregator>
}
```

PR #3458, opened 2015-05-08
Update gridTestUtils.spec.js

shabesoglu:
I am suggesting to pass browser instance to each function because it is useful when you fork a new browser instance and you don't want to use the global browser variable.


PaulL1:
Two comments:

1.  This looks like extra work for every user of the test utils in order to support a reasonably obscure use case.  Is there some way to do this that doesn't require the browser instance to be passed into every single function.

2.  The code isn't passing the style checking - did you run the tests?

PaulL1:
I think a different PR recently provided a different way to include the grid id in a standard way - check the latest E2E tutorial.

Issue #3340, opened 2015-04-22
Scroll up not propagating to parent when at top of "short" grid


GauravButola:
I've created this plunk[1] to show the bug.

Scroll down while keeping your mouse inside the grid, keep your mouse in grid and, now try to scroll up.
This bug doesn't seem to appear in the 3.0.0-rc.19 release.

Is there a solution to this in the latest 3.0.0-rc.20 release?

[1] http://plnkr.co/edit/ke8jCXG7qsOzWRqOsrw1?p=preview

PaulL1:
This is working on the tutorial: http://ui-grid.info/docs/#/tutorial/101_intro, I assume it's a difference between rc20 and unstable, and is therefore already fixed in unstable.

GauravButola:
Thanks for the quick response @PaulL1. So, I guess I should downgrade to 3.0.0-rc.19 until new version is released.

Thanks for the awesome work :)

PaulL1:
I'd be more inclined to upgrade to unstable, there's quite a bit in there.  I can't recall what other bugs you're inheriting by going back to rc19.

GauravButola:
Ah! then it'll make sense to use the unstable. What would you say is the best way, to get the unstable version? Right now I'm getting the rc20 from bower.

PaulL1:
Easiest to get it from ui-grid.info/release. Check the tutorials for links. 

You can also clone the repo and build, but that's a bit fiddlier.

GauravButola:
Great! thanks for your help.

Kidsisker:
I am implementing the latest version and I am also experiencing this issue. I have several small grids on a page (<= 10 rows) and the user is unable to scroll up while mouse is over the grid. The vertical scrollbar is disabled. I do also notice that this issue can be reproduced on the tutorial that you sited above. Any updates? I have just finished a massive refactor so I am not overly concerned about his issue, but my users might. Thanks so much!

Jacquelin:
Hi, I am also experiencing this issue. And it doesn't work on http://ui-grid.info/docs/#/tutorial/101_intro . 
This problem can be reproduced on a grid that has no scroll bar activated.

I try to debug on the last unstable version (RC21). I add some comment, if it helps you to find what's wrong, I don't want to try to presume what the problem ...

```javascript
if (event.deltaY !== 0) {
    // On scroll to top, scrollYAmount < 0 (-53 for me). On scroll to bottom, scrollYAmount > 0 
[...]

PR #3531, opened 2015-05-19
Fix scroll up not propagating to parent

Jacquelin:
Hello,

This pull request is to fix, what I consider a bug, the issue #3340

This 'bug' happen for a grid that displays less than its height (3 line, 3 * 30px, for a height of 300px for example). When this happening, the scroll up is not propagated to the parent.

I assumed that if the VerticalScrollLength AND the scrollTop + scrollYAmount is negative, then the scrollYPercentage is set to 0. 

It's working for a grid with the vertical scroll avalaible, or not.

### DIFF
--- a/src/js/core/directives/ui-grid-render-container.js
+++ b/src/js/core/directives/ui-grid-render-container.js
@@ -74,10 +74,13 @@
                 var scrollYAmount = event.deltaY * -1 * event.deltaFactor;
 
                 scrollTop = containerCtrl.viewport[0].scrollTop;
+
+                var scrollCount = scrollTop + scrollYAmount;
                 // Get the scroll percentage
-                var scrollYPercentage = (scrollTop + scrollYAmount) / rowContainer.getVerticalScrollLength();
+                var scrollYPercentage = scrollCount / rowContainer.getVerticalScrollLength();
 
                 // Keep scrollPercentage within the range 0-1.
+                if (scrollCount < 0 && rowContainer.getVerticalScrollLength() < 0) { scrollYPercentage = 0; }
                 if (scrollYPercentage < 0) { scrollYPercentage = 0; }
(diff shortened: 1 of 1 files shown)


PaulL1:
@swalters: any chance you'd take a quick look at this one too?  It looks sort of OK in concept, but I know there are gremlins hiding in scrolling code.

Jacquelin:
Hi, 

I think of another possibility would be to consider rowContainer.getVerticalScrollLength (); should be an absolute value (never negative!). It solve the bug too. If you think it's a better solution, you can just reject the PR and I will make an another PR with the absolute value

thx

rupampatel2006:
@Jacquelin Your fix works well. @PaulL1 Are you going to merge this fix?

PaulL1:
@rupampatel2006 : not without @swalters reviewing it.

Jacquelin:
Hi,

no news from @swalters ?

tuhoojabotti:
I would really like this to get fixed, not being able to scroll up at all on my page with a full width grid.

Jacquelin:
Sorry but my pull request does not fix the scroll issue ...

I make several test : 
- With no scroll needed 
 - For this case, my fix seems enough
- With scroll but VerticalScrollLength < 0 (http://plnkr.co/edit/Xoianpl2wWSKuQVtuxOt?p=preview)
 - Does not work !
- With scroll and VerticalScrollLength > 0
 - Work fine

This PR should be rejected, I'm currently working on a better fix ... sorry :-/

tuhoojabotti:
I think the main issue for me is that, even though I have set `enableVerticalScrollbar: uiGridConstants.scrollbars.NEVER` and I resize the grid to the contents it still eats the scroll event.

Jacquelin:
@tuhoojabotti : It's what I was trying to fix in first place (my fix work for this !), but still ... incomplete for a true bug fix. You can discuss about this at #3340

PR #3544, opened 2015-05-20
Fix for issue 3528. When exiting the grid the focusedCells list was n…

willcodeforbeer:
This is a fix for issue 3528. The idea is the focusedCells list wasn't getting cleared when navigating outside the grid and so the book keeping wasn't being done when focus was returned. There's a side effect here that I'm not sure about... With this fix tabbing out of the grid causes the focus highlight in the grid to disappear. I think it's probably valid but up to you! I included a unit test for the fix.


PaulL1:
This looks good to me, but I'd want @swalters to cast an eye over it before merging - just in case it messes with something he knows about and I don't.  Thanks for the contribution.

PR #3564, opened 2015-05-22
Add 2 options: In CellNav, add option to highlight the entire row and in Selection add option to show/hide highlight on selection

mloutraris-ge:
We have had to customize a lot in the grid to achieve the enhancements
requested. We use the grid in a way that makes the row the item which is
actionable, not an individual cell. It would help us a lot if the
following option were added:

1. In the cellNav feature, add an option that will highlight the entire
row rather than the just the cell on the row that has focus. This will
act as a highlight bar that will indicate the row that we will act upon.

2. In The selection feature, add an option that does not apply the row
highlight on selection. This will help us by not causing confusion when
used in conjunction with the highlight bar. The check box in the header
will signal the row is selected.

Both changes are included in this pull request.


mloutraris-ge:
Note: the extra reverts were done by accident, Git hub client was acting up on me.

AgDude:
Could you please revert all whitespace changes and squash into 1 or 2 commits, so we can easily see what changes are being made.

@PaulL1, it looks like you are helping out with this PR. I would actually like some similar functionality to the "select row" described here. I haven't spent enough time working with cellNav to understand the current limitations, but my thought was to create a flexible API for capturing keyboard and mouse events. I point that out here only so we keep in goal in mind before we get too much options bloat in to the cellNav area.

mloutraris-ge:
I am desperately trying to…. I am close this pull request because I have to refork and clean up, I am past the point of no return. There changes are small but I agree difficult to see. I am working on that, hope to clear it up and resubmit a cleaner version.

mloutraris-ge:
Closing this, going to resubmit cleaner version based on current code.

mloutraris-ge:
Hello,

I have cleaned up the pull request, the new pull request is 3746

https://github.com/angular-ui/ng-grid/pull/3746

Thanks,
Mike

Issue #3453, opened 2015-05-07
error on adding column with displayName and without name


ndudenhoeffer:
When adding columns to the grid, the name property is generated if not already defined. The method which is generating the name is using the ```displayName``` to check for duplicates, not the actual name. So adding multiple of the same displayName fails.

I haven't dug into it any deeper, but you can have two instances.  When you try to add the third it fails to generate a name, and leaves it undefined (which causes problems later). 

http://plnkr.co/edit/0D7qgs?p=preview

As a side note, there are quite a few comments in the code about allowing a blank "name" for backwards compatibility. I think this also actually an important feature to have. Consider the case of grouping and showing different aggregations for different columns, where the user is allowed to chose which columns to add and remove.

AgDude:
Here's a quick work around, to set the name before adding a column:

        var columnDef = {...}
          counter = 1;
        columnDef.name = columnDef.field;

        while ( gridApi.grid.getColumn(columnDef.name) ){
          columnDef.name += counter.toString();
          counter++;
        }
        gridOptions.columnDefs.push(columnDef);

wesleycho:
Closing, as this issue looks resolved

PR #3571, opened 2015-05-23
fix(uiGridColumns): Fix auto-incrementing of column names

AgDude:
When generating unique column names, do not consider the displayName.
Increment column name property only, using the field as the base.
Increment displayName property only if it is auto-generated,
displaying any provided displayName unchanged.

Fixes: #3453


PaulL1:
Much tidier than the previous code, thanks.

Issue #3572, opened 2015-05-23
[Grouping] treeRowHeaderAlwaysVisible not working as expected


AgDude:
My expectation with this option is that when set to ```false```,  the "plus" column will be removed/added based on whether or not grouping is active. This does not appear to be working.

AgDude:
@PaulL1  I have been working a bit on this. The core problem is solved by resetting ```grid.treeBase.numberLevels``` to 0 at the beginning of the ```createTree``` method. But then we run into the problem that we are considering a single grouping to have 0 ```numberLevels```, which is also the default without any grouping.

I am going to change this so ```numLevels``` becomes 1 if there is a single grouped column. Alternatively we could make it ```undefined``` if there are no grouped columns.

PaulL1:
Sounds reasonable.  I don't think I did it deliberately, I recall noticing it had ended up zero based then just going with it.

PR #3573, opened 2015-05-23
fix(treeBase): Change calculation of number of levels in tree

AgDude:
Change the treeBase.numberLevels variable to consider a single level as 1 instead of 0
Previously there was no way to use this variable to differentiate between a single tree level
and no grouping at all. This does not impact row.treeLevel, which is still zero-based.

Reset treeBase.numberLevels each time the tree is created.

Remove unneeded dependency for groupingService in treeBase.uiGridViewport directive

Add treeRowHeaderAlwaysVisible=false option to complex grouping tutorial (320).

Fixes: #3572


PaulL1:
Looks very tidy too.  Thanks @AgDude

PR #3610, opened 2015-05-27
Create ta.js

jeraldfdo:
Tamil lanuguage

[Review on Reviewable](https://reviewable.io/reviews/angular-ui/ng-grid/3610)

### DIFF
--- a/src/js/i18n/ta.js
+++ b/src/js/i18n/ta.js
@@ -1,0 +1,83 @@
+/**
+ * Created by Tim on 2/1/14.
+ */
+(function () {
+  angular.module('ui.grid').config(['$provide', function($provide) {
+    $provide.decorator('i18nService', ['$delegate', function($delegate) {
+      $delegate.add('en', {
+        aggregate: {
+          label: 'உருப்படிகள்'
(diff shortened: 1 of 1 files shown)


PaulL1:
This appears to change en to have tamil decodes - I think inside the file you needed to set the language to ta?  That in turn means it breaks the unit tests.

When you fix, could you please squash the commit again into a single commit?

c0bra:
@jeraldfdo Any change you could check out Paul's comment?

src/js/i18n/ta.js
```diff
@@ -0,0 +1,83 @@
+/**
+ * Created by Tim on 2/1/14.
+ */
+(function () {
+  angular.module('ui.grid').config(['$provide', function($provide) {
+    $provide.decorator('i18nService', ['$delegate', function($delegate) {
+      $delegate.add('en', {
```
JLLeitschuh: This is the only line that needs to be fixed. Please change this to `$delegate.add('ta', {` and we can merge this in.

JLLeitschuh:
Closed by #4120

Issue #2549, opened 2015-01-14
Need to make ExpandableRowHeight dynamic


flea19:
Need way to dynamically set height on the ExapndableRow. This is needed for a Feature in the Open Source tool OSEE.

PaulL1:
This has been discussed a number of times, and there are some practical limitations on the ability to do that.  There is nobody currently working on it, however we are more than happy to accept a pull request adding this feature: https://github.com/angular-ui/ng-grid/blob/master/FIRST_TIMER.md and https://github.com/angular-ui/ng-grid/blob/master/CONTRIBUTING.md

kevin-sage:
I have been trying to work around this for a while.  I am going to go ahead and start working on this enhancement.

PaulL1:
I think we probably should be clear exactly what we're doing.  I've seen a few people asking for this in different ways, and I think @jpuri is of the opinion that it's not possible.  So I'd get a few eyes on it before you start coding.

I think there are a few things to consider in here:

1. The grid doesn't really deal with dynamic row heights.  The scroll height calculates based on an assumption about the row height.  However, the expandable does some extra tricks, because the height of expandable rows isn't automatic.  I suspect that it works based on `visibleRows * rowHeight + expandedRows * expandedRowHeight`
2. If all the expanded rows must have the same height, then it gives some grief when the expanded row has only one item in the subgrid, or has 100 items.  Reasonably they might each be different.
3. I think the request is that we be able to somehow show this, particularly if the subgrid has more rows than fits in the expandable height.  Since you need to know the expandable height up front, and you may not know how many rows there are in an expanded row until you click on it, you can get into a situation where you have more rows in your subgrid than can be displayed
4. I think the most logical answer is to permit scrolling in the subgrid.  If we permit instead dynamic subgrid heights, then calculation of scroll height will require us to sum the height of each subgrid, including subgrids that aren't currently displayed
5. I've sort of thought of another way, in which we store against each row the "height up to here".  We'd recalculate all the "height up to here"s whenever someone changes a height - changes the height of a row, expands a subgrid, sorts or filters the grid etc.  But we wouldn't have to calculate it on every scroll.  I'm not sure how much work this would turn into, but @swalters has been refactoring this scrolling code recently so it may be easier than it used to be.

kevin-sage:
[...]

PR #3615, opened 2015-05-28
Added dynamic height for subgrids

JayHuang:
This PR adds a dynamic height for subgrids instead of setting one fixed height for all subgrids (which easily becomes very frustrating to use when there are large discrepancies in subgrids, resulting in large amounts of whitespace). Addresses #2549 along with many other issues I've noticed when first looking for a solution.

The change has been used for about ~2 months on my own application and worked perfectly for my use cases. Hasn't been tested extensively otherwise.

### DIFF
--- a/src/features/expandable/js/expandable.js
+++ b/src/features/expandable/js/expandable.js
@@ -23,7 +23,7 @@
   module.service('uiGridExpandableService', ['gridUtil', '$compile', function (gridUtil, $compile) {
     var service = {
       initializeGrid: function (grid) {
-        
+
         grid.expandable = {};
         grid.expandable.expandedAll = false;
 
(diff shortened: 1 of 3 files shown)


bierdok:
+1

AgDude:
@JayHuang Thanks for the contribution, however it appears that there are JShint errors. The default grunt task should run jshint. We would also appreciate it if you can revert all of the whitespace changes, as that makes the actual changes much easier to see.

Please also be sure when you update this to rebase against the current master and squash to keep it a single commit.

swalters:
Thanks a lot for the PR.  Unfortunately, there is an existing PR for this issue that we have chosen to merge.  If you see that it doesn't fix your exact issue, you can then submit another PR after that.  The duplicate is #3201 .

PR #3621, opened 2015-05-28
Enh(grouping): BREAKING change aggregation options and improve custom…

### LINKED ISSUE #3302: [Grouping] Custom aggregation functions
Hi

I've tried the new grouping feature and find out that it is impossible to provide a function that calculates a group value. For example, if I want to group by a column that contains DateTime values, now the Grid creates a group for each distinct value in the column. Instead it would be better to have groups like Today, Yesterday, This Week, This Month and so on. It can be done by adding a new field like valueFormatter or valueBuilder in the columnDef.grouping object. The type of the field is function that accept a cell value and return a formatted group value.

AgDude:
… aggregation support

Add gridOption treeAggregationFns, to add custom aggregations with support for both the column menu and saveState.
Refactor native aggregations to follow the same api as custom aggregations.

Remove columnDef option treeAggregation and replace with separate options treeAggregationType and treeAggregationLabel

Remove "custom" aggregation type, as it is redundent with specifying a custom
aggregation function.

Allow custom aggregation finalizers to omit setting the rendered string. Render the value (and label if provided)
if a rendered string is not provided.

Fix groupingShowCounts option

Breaking changes are:
- options previously specified on the treeAggregation object are now  separate options (see docs)
- aggregation entity type property may be undefined in the case of a unnamed custom aggregation function
- aggregation entity previously set the initial value to 0, it now begins as undefined

Closes #3302

### DIFF
--- a/misc/tutorial/209_grouping.ngdoc
+++ b/misc/tutorial/209_grouping.ngdoc
@@ -81,8 +81,7 @@ could potentially interact with people's custom templates.
 
 Tuning the way aggregations work can be done through defining a columnsProcessor that runs with higher (later)
 priority than the groupingColumnProcessor (so higher than 400), and that looks for grouped or aggregated columns
-and changes things like the treeAggregation.type, or the customTreeAggregationFinalizerFn.  A 'complex' tutorial
-may be forthcoming that provides examples of this.
+and changes things like the treeAggregationFn, or the customTreeAggregationFinalizerFn.  See tutorial 320 for an example.
 
 
 @example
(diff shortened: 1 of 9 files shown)


AgDude:
@PaulL1 I have butchered one of your tests in treeBase (some aggregations over two parents). It looks like you were testing multiple aggregations on one column (some native and one custom). I don't think that is supported the way I re-implemented aggregations. You can have multiple aggregations on the same field with different columns, or set different values on the aggregation entity using custom aggregation functions. 

If that doesn't work for you, could you please describe the use case?

PaulL1:
That was just laziness on my testing - easier to do lots of aggregations on the same column, and that subset of the code at that time didn't particularly mind.  The full code base didn't support multiple aggregations on the same column.

This overall looks really good @AgDude, the codes clean and better than it was before, which is always hard when touching someone else's code.

The only glitch I see is a commented out block in src/features/tree-base/test/tree-base.spec.js that could probably be deleted, but we can do that sometime.  I've merged.

PR #3640, opened 2015-05-29
add translate to i18n 'es' on pagination directive.

bpena:
hi. 
I need this translations for pagination.

### DIFF
--- a/src/features/pagination/js/pagination.js
+++ b/src/features/pagination/js/pagination.js
@@ -131,7 +131,7 @@
 
           grid.api.registerEventsFromObject(publicApi.events);
           grid.api.registerMethodsFromObject(publicApi.methods);
-          
+
           var processPagination = function( renderableRows ){
             if (grid.options.useExternalPagination || !grid.options.enablePagination) {
               return renderableRows;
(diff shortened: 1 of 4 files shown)


PaulL1:
I think you also need to change the paginationJs to create the value `translationOf` if you want this to work.

bpena:
If you're right. I forgot to add the change of paginationJs

PaulL1:
I still can't see the pagination.js changes, and could you possibly squash all the commits once you're done submitting that?

src/features/pagination/js/pagination.js
```diff
@@ -343,20 +343,21 @@
           $scope.paginationApi = uiGridCtrl.grid.api.pagination;
           $scope.sizesLabel = i18nService.getSafeText('pagination.sizes');
           $scope.totalItemsLabel = i18nService.getSafeText('pagination.totalItems');
-          
+          $scope.paginationOf = i18nService.getSafeText('pagination.of');
```
JLLeitschuh: @PaulL1 I think its right here.

PaulL1:
Ah yes.  So it's just the squash and then we can merge?

JLLeitschuh:
Taken care of. This also closes #3070

JLLeitschuh:
Merged in 1338b6c901b7030ab4d20a5e7df497294e5a00f9

bpena:
thank you for merged.
I did not have internet, so I have not answered before.

PR #3666, opened 2015-06-02
Fix bug scroll (up and down) does not propagating to parent 

Jacquelin:
Hello,

This pull request is to fix the issue #3340.

My first pull request #3531 was not good enough.

Now, I only check if the Vertical scroll length is negative. If it is, then I let the parent the parent container scroll.

Sorry for this double Pull request ...

### DIFF
--- a/src/js/core/directives/ui-grid-render-container.js
+++ b/src/js/core/directives/ui-grid-render-container.js
@@ -74,6 +74,7 @@
                 var scrollYAmount = event.deltaY * -1 * event.deltaFactor;
 
                 scrollTop = containerCtrl.viewport[0].scrollTop;
+
                 // Get the scroll percentage
                 var scrollYPercentage = (scrollTop + scrollYAmount) / rowContainer.getVerticalScrollLength();
 
@@ -98,7 +99,8 @@
               }
 
               // Let the parent container scroll if the grid is already at the top/bottom
-              if ((event.deltaY !== 0 && (scrollEvent.atTop(scrollTop) || scrollEvent.atBottom(scrollTop))) ||
+              if (rowContainer.getVerticalScrollLength() < 0 ||
+                  (event.deltaY !== 0 && (scrollEvent.atTop(scrollTop) || scrollEvent.atBottom(scrollTop))) ||
                   (event.deltaX !== 0 && (scrollEvent.atLeft(scrollLeft) || scrollEvent.atRight(scrollLeft)))) {
                 //parent controller scrolls
               }


PaulL1:
Thanks for the contribution.

My concern is that I'm not sure that getVerticalScrollLength should be returning < 0, so there's a risk that someone will fix that, and thereby break this again.  

Also, if we were to merge it we'd need the three commits squashed into one.

@swalters, do you have an opinion?

Jacquelin:
PaulL1 said :
> My concern is that I'm not sure that getVerticalScrollLength should be returning < 0, so there's a risk  that someone will fix that, and thereby break this again.

You are right. So I test some code fix with getVerticalScrollLength as an absolute value. This fix some scroll but not all. I try with getVerticalScrollLength === 0, but same, this fix some scroll but not all ... Waiting for @swalters response


> Also, if we were to merge it we'd need the three commits squashed into one.

For this I don't know how I can do it ...

> Thanks for the contribution.

it's a pleasure ;-)

PaulL1:
@Jacquelin: thanks for the input, we'll wait for @swalters to see what he can tell us.

On the squash, the basic process is to issue `git rebase -i upstream/master`, this will take you into an interactive process in which you can tell git to squash all your commits into a single commit, and rewrite the commit message that goes with that.  You then `git push -f` to force the remote to take the rewritten commit log.

Jacquelin:
Thank you for the squash !!!!!! ;-)

swalters:
@Jacquelin I have to agree with @PaulL1 . What should be fixed is the ScrollEvent.AtTop and .AtBottom.  Those should both return true if the grid is truly at top and/or bottom row.

Jacquelin:
@swalters Oki ! 

But how can I fix it ? 

When rowContainer.getVerticalScrollLength() < 0, So the 'y.percentage' is mixed (or reverse). and this make :
```
// When on top : this.y.percentage === 1, then atTop return false.
ScrollEvent.prototype.atTop = function(scrollTop) {
  return (this.y && this.y.percentage === 0 && scrollTop === 0);
};
```

```
// When on bottom : this.y.percentage === 0 so atBottom return false
ScrollEvent.prototype.atBottom = function(scrollTop) {
  return (this.y && this.y.percentage === 1 && scrollTop > 0);
};
```

What do you think if I set the scrollYpercentage before call atTop/Bottom ?
```
// Keep scrollPercentage within the range 0-1.
if (rowContainer.getVerticalScrollLength() < 0) {
  if (scrollTop > 0) { scrollYPercentage = 1; }
  else { scrollYPercentage = 0; }
}
else if (scrollYPercentage < 0) { scrollYPercentage = 0; }
else if (scrollYPercentage > 1) { scrollYPercentage = 1; }
```

swalters:
I made a comment in #3690  I think we'll go that direction for the fix.  Thanks for the PR

PR #3674, opened 2015-06-02
Hide filter cancel button if filter input is an empty string

CharlesShen:
The filter cancellation button is still visible if I type some characters into a filter textbox and subsequently delete all the characters.  The "ng-show" on the filter cancellation button currently does not test for empty strings.

### DIFF
--- a/src/templates/ui-grid/ui-grid-filter.html
+++ b/src/templates/ui-grid/ui-grid-filter.html
@@ -13,7 +13,7 @@
     </select>
 
     <div class="ui-grid-filter-button-select" ng-click="colFilter.term = null" ng-if="!colFilter.disableCancelFilterButton">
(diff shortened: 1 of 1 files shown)


PaulL1:
In master this appears to work already - look at the 208 tutorial as an example.  Are you perhaps fixing something that is already fixed, or is there a particular case that is not working?

CharlesShen:
Hi PaulL1,

Thanks for the response.  I have confirmed that the tutorial works properly and have also confirmed that the problem exists on my end.  I'll review differences between the two setups and see if I can isolate the problem.

Thanks!

CharlesShen:
Hi PaulL1,

I was using RC21 and it looks like you have committed a fix to master here
https://github.com/angular-ui/ng-grid/commit/22e6acd81ece42f1133712199edc0142df920ed1

Sorry, I made some rush assumptions and that definitely caused some misjudgments on my end.

Thanks!

PR #3680, opened 2015-06-03
Fix add row to add to hash map and Fix modify row to update entity if already exists

mikenorgate:
Fix for row values not being updated on modify rows
Fix for rows not being added to hash map on row add

[Review on Reviewable](https://reviewable.io/reviews/angular-ui/ng-grid/3680)

### DIFF
--- a/src/js/core/factories/Grid.js
+++ b/src/js/core/factories/Grid.js
@@ -1083,6 +1083,8 @@ angular.module('ui.grid')
       } else {
         // otherwise, manually search the oldRows to see if we can find this row
         newRow = self.getRow(newEntity, oldRows);
+      } else {
+          newRow.entity = newEntity;
       }
 
       // if we didn't find the row, it must be new, so create it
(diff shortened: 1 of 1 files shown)


PaulL1:
It's not clear to me what issue we're trying to solve here.  I think it works when we add rows to the data at the moment?

swalters:
@mikenorgate can you please respond with an explanation of this bug?  We would like to know how it is affecting your application.

mikenorgate:
Sorry for the delayed reply. This was to fix two bugs. The first was that when using `modifyRows`, if the row exists it was not updating the entity. I think the merge has put the else on the wrong line. 

The second fix was when using `addRows` the added row was not being added to the hash map, meaning it wasn't found when trying to do an operation on it. This presented itself when we added a row and then tried to select it.

src/js/core/factories/Grid.js
```diff
@@ -1083,6 +1083,8 @@ angular.module('ui.grid')
       } else {
         // otherwise, manually search the oldRows to see if we can find this row
         newRow = self.getRow(newEntity, oldRows);
+      } else {
```
JLLeitschuh: This needs to be fixed before we can accept the PR

PR #3690, opened 2015-06-04
Fixed scrolling when verticalScrollHeight is 0

500tech-user:
Open the basic example:
http://ui-grid.info/docs/#/tutorial/101_intro
Hover above the grid and try to scroll up - it doesn't scroll the page.
Scrolling down, however, works.

This fixes it

### DIFF
--- a/src/js/core/directives/ui-grid-render-container.js
+++ b/src/js/core/directives/ui-grid-render-container.js
@@ -75,7 +75,10 @@
 
                 scrollTop = containerCtrl.viewport[0].scrollTop;
                 // Get the scroll percentage
-                var scrollYPercentage = (scrollTop + scrollYAmount) / rowContainer.getVerticalScrollLength();
+                var verticalScrollLength = rowContainer.getVerticalScrollLength();
+                var scrollYPercentage;
+                if (verticalScrollLength === 0) { scrollYPercentage = 0; }
+                else { scrollYPercentage = (scrollTop + scrollYAmount) / verticalScrollLength; }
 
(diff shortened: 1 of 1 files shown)


tuhoojabotti:
Duplicates: #3666 #3531 #3340 (for reference)

swalters:
@500tech-user Can you tell me is this still an issue in master?  I think this was fixed in some other commits.

swalters:
ok, I see that mousewheel is still broken for it

swalters:
Thanks for PR.

Rather than putting the logic where you have it, we would rather see the verticalScrollLength property added to scrollEvent and then the atBottom, atTop, atLeft, and atRight can use that number to determine if it is at bottom, top, etc.

I'll look at making this change and will close this PR when complete.

swalters:
Should be fixed with
https://github.com/angular-ui/ng-grid/commit/e582174a826bb232ddbb4fda8001b64c3273df0d .  Thanks for the PR

500tech-user:
Awesome
BTW, I also noticed that when zooming in or out the calculation is wrong.
Did you ever encounter that?

-- 
- Adam Klein <http://il.linkedin.com/in/adamklein500>
- CTO @ 500Tech.com <http://500tech.com/> - Web and Mobile development
- hackademy.co.il <http://hackademy.co.il/angularjs/about> - AngularJS,
Ruby on Rails training
- (+972) 52 747 5633

rhrasna:
The problem persists in v.3.0.1.

PR #3697, opened 2015-06-05
Changed from offsetParent to parentElement. Use of offsetParent is a …

helgean:
…major performance degradation (in IE, any use of offsetParent, offsetWidth or offsetHeight should be used with extreme caution) , it causes document reflow/layout in the browser.

I can't see why this have to be offsetParent, it looks like it is the nearest parent of the element that is the target. If changed to parentElement this would have a positive effect on the initial rendering performance. If all calls to offsetWidth, offsetHeight and offsetParent are eliminated from the initial rendering, the initial rendering time drops by almost 40% in our application.


PaulL1:
@c0bra: any thoughts on this?

c0bra:
LGTM

PR #3703, opened 2015-06-05
Handle date strings in date sort function

dabedur:
Add a check to ensure that parameters passed into sorting function are indeed `Date` objects prior to calling `getTime()` on them


PaulL1:
Nice, thanks.

dabedur:
Not a problem. Thank you for making a kick-ass grid!

PR #3748, opened 2015-06-11
Remove moot `version` property from bower.json

kkirsche:
Per bower/bower.json-spec@a325da3

[Review on Reviewable](https://reviewable.io/reviews/angular-ui/ng-grid/3748)


c0bra:
@kkirsche hey thanks for the PR. The bower.json file is automatically built on Travis. I can look into removing the version property.

Issue #3459, opened 2015-05-08
Pinning and navigation issue


leelakrish:
When you pin a column in ui-grid the cell navigation selection is not visible after 4th column. 

http://plnkr.co/edit/so8HIWpvrnqdXyr668km

PaulL1:
Are you saying that when you use cell nav and the arrow keys to go to the right, that we lose the cellNav when the grid scrolls?  If so, I don't think that's unique to pinning, it's a current defect with cellNav.

leelakrish:
The First column is pinned to the left, now select the second cell in the first row. Then press right arrow key to navigate to the next column. When you keep navigating to other cells, the cell selection is not visible. the cell selection is not visible only when I pin the column.

c0bra:
@leelakrish Can you try with RC22 and see if it's any better?

leelakrish:
It is still behaving in the same manner how It was working with the older version.

mportuga:
No longer reproducible
http://plnkr.co/edit/WnZPj99xztlZQ4repBMK?p=preview

PR #3758, opened 2015-06-11
Pinning and navigation issue #3459

wcTalons:
When using arrow keys to nav and having columns pinned the table would
not scroll correctly.

[Review on Reviewable](https://reviewable.io/reviews/angular-ui/ng-grid/3758)


c0bra:
There's a couple things here:

1. Can you squash these commits into one,please? http://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git
2. Your build is failing so you'll need to fix the unit tests
3. I'm concerned about your change to the render container that removes the margins. Does that not prevent proper movement of the rendered data within the canvas?

Could you go over your thought process on fixing this bug? Thanks very much.

PR #3775, opened 2015-06-13
add i18n ja setting to 2.x

biblichor:
I added i18n ja setting to 2.x. Plz merge them.

[Review on Reviewable](https://reviewable.io/reviews/angular-ui/ng-grid/3775)


c0bra:
Hey thanks for the PR, but it looks like you've got some commits that don't change any files and some commits that change the concatenated JS files, which should be left alone.

Can you please squash and clean up your commits? If so I'll get this merged. Thanks again!

PR #3788, opened 2015-06-16
Add gridObjectTestUtils.

amyboyd:
Before:

    var gridTestUtils = require('lib/gridTestUtils.spec.js');
    gridTestUtils.expectHeaderColumnCount('users-table', 6);
    gridTestUtils.expectHeaderCellValueMatch('users-table', 0, 'ID');
    gridTestUtils.expectHeaderCellValueMatch('users-table', 1, 'Name');
    gridTestUtils.expectHeaderCellValueMatch('users-table', 2, 'Email');

After:

    var GridObjectTest = require('lib/gridObjectTestUtils.spec.js');
    var usersGrid = new GridObjectTest('users-table');
    usersGrid.expectHeaderColumnCount(6);
    usersGrid.expectHeaderCellValueMatch(0, 'ID');
    usersGrid.expectHeaderCellValueMatch(1, 'Name');
    usersGrid.expectHeaderCellValueMatch(2, 'Email');

### DIFF
--- a/test/e2e/gridObjectTestUtils.spec.js
+++ b/test/e2e/gridObjectTestUtils.spec.js
@@ -1,0 +1,30 @@
+/* globals element, by */
+
+var gridTestUtils = require('./gridTestUtils.spec.js');
+
+function getProxyToRealMethod(gridId, method) {
+    return function() {
+        var callArgs = [gridId].concat(Array.prototype.slice.call(arguments));
+        gridTestUtils[method].apply(gridTestUtils, callArgs);
+    };
+}
+
+/**
+ * @ngdoc overview
+ * @name ui.grid.e2eTestLibrary
+ * @description
+ * End to end test functions.
+ */
+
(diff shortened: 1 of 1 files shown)


amyboyd:
Sorry for closing -- thought there was a bug in this -- it was actually a bug in my real tests!

PaulL1:
That's really elegant.  And it looks like it's backwards compatible?  (i.e. we can choose to change the existing tests or not change them?)

amyboyd:
@PaulL1 Yep, it's fully backwards compatible. Thanks for the merge. :)

PaulL1:
Further, I've just updated the 101 and 102 tutorials to use this format.  I noted that @c0bra (I think) has needed to use promises to deal with Firefox, and that code relies on the return value from some of the functions.  I updated the object to cascade the return value from each function, which seems to allow that code to keep working as well.

Issue #2840, opened 2015-02-24
Keyboard-enable all features for section 508 compliance and accessibility


ygyg70:
Column sorting, hiding etc. doesn't seem to respect focus and keyboard interaction.
Is this a planned capability?

PaulL1:
It is something we'd like, but not currently in the roadmap.

ygoraly:
Section 508 compliance is a requirement for many software projects. As of today all ExtJS components support keyboard interaction. Would be nice to have an open source, modern, high quality alternative.

PaulL1:
Yeah, we'd really love to have it too :-).  I don't think it's overly hard, it's just a bunch of time to invest to work through all the permutations and handle them.  We'd really welcome some assistance - the gitter channel is quite active, and there's a good first time contributors guide.  Do you have to time to help out with it?

ygoraly:
This is not my personal project. Will include that option when selecting framework and components.

c0bra:
@ygoraly ExtJS has the benefit of being backed by a company so they can afford to tackle things like this. A small open source project really can only do what users are willing to contribute, and that's usually only the things that the users themselves need.

If enough people want the grid to be 508 compliant then someone will probably write it. The other option would be paying for it by funding for the issue on bountysource or something like that.

rajulkansara:
I would like to give my best shot. Let me know.

PaulL1:
We'd welcome any PRs on this topic.

I'd suggest by starting small to get the feel of the process.  Take a look at the first time contributors guide:  https://github.com/angular-ui/ng-grid/blob/master/FIRST_TIMER.md

JLLeitschuh:
Assuming that there has been no further progress on this. Assuming that this
I have yet to actually peruse the repository thoroughly however I understand the general idea. Should accessibility be a default or something that is implemented as a feature?

PaulL1:
1.  It should be default.  I see no reason that anybody would want to turn accessibility ff
2.  I wouldn't deal with some of the features if I were you.  I'd skip expandable and leave it for someone who actually uses it.  My feel is that treeView and grouping have replaced a lot of the space that expandable was filling, and that expandable has some kruft built up in it that may make things hard
[...]

PR #3850, opened 2015-06-24
Adds Voice Over Accessibility support and Keyboard Nav

JLLeitschuh:
# Accessibility
- Adds semantically correct role tags to the grid according to the [W3 Aria Standards](http://www.w3.org/TR/2011/CR-wai-aria-20110118/). This allows a screen reader to accurately describe the content on the screen auditorily.
- Adds focus management and controls on all menus and inputs. When opening a menu the focus automatically moves into that menu. This means that most of the core grid is fully keyboard accessible.
- Pagination button controls now show the focus cursor on OSX with Chrome. They also move the focus when they are disabled when hitting the page limits. The pagination controls are also semantically labeled as a menu for Screen Readers.
- All aria labels use i18n.
- Adds focus helpers to the `gridUtil` to make setting focus on elements easier inside of directives.

## Requirements
In order to take full advantage of the accessibility and keyboard support angular aria must be included as a dependency to the app. Aria adds the tabindex to dom elements that need them and adds keypress handlers so that the enter key triggers the ng-click action.

## Review Specifics
Before merging it would be nice if you were to review the following.
- Headers:
  * The header is now a button that can be clicked using the OSX Voice Over cursor (because it simulates a click. However, it can not be clicked by using the enter key (because the header doesn't use an ng-click directive). Instead it uses complex logic to determine if a click sort is happening or if a click and drag is happening.
  * When you hide a column the focus is automatically moved to the gird menu button. However, if the grid menu is not enabled then focus is lost and a warning is printed to the console. I don't know where else to set the focus when you remove a column using the menus.
- Pagination:
  * Does the new CSS on the buttons work for the rest of the grid?
    * Does this break on any other browsers? eg. not an OSX browser?
  * Any edge cases for focus movement that I didn't think about causing the document focus to be lost?
- General focus:
 * The focus highlight stands out quite a bit. You may get complaints from users about this. However there are many people in the accessibility world that say to [stop messing with the browser default](http://tjvantoll.com/2013/01/28/stop-messing-with-the-browsers-default-focus-outline/). Do we want to cater to these users that insist upon hiding the focus highlight or just tell them to work around it themselves.
 * Angular Aria [automatically adds keypress events](https://docs.angularjs.org/guide/accessibility#ngclick) to buttons with ng-click. This means that without aria you will get the focus highlight on elements but you can't use the keyboard to click them. I don't think this is a problem however I think this could be discussed.

*Related* #2840

[Review on Reviewable](https://reviewable.io/reviews/angular-ui/ng-grid/3850)


c0bra:
Looking really good! A couple notes: 

1. I personally don't like the focus highlight, if by that you mean the outline on the elements that were changed to buttons. It makes the UI feel strange to me and I would prefer to prefer to remove it. Users can always set it back to normal if they want that.
2. The pagination controls are overlapping the bottom of the grid now:

JLLeitschuh:
Response to
 1. I would argue the opposite, there are countless accessibility professionals out there who say, don't remove the focus highlight. Also, by doing so, you are making the entire grid unusable for a very large population of web users. Alternatively, I would suggest [styling it](http://a11yproject.com/posts/never-remove-css-outlines/). If you really want an option that can be toggled I would prefer it to be something that users are forced to turn off. At least then it forces devs to think about the entire demographic of their users.
 2. I experienced that problem once. I did a `grunt clean build` and never saw it again. However, I've also seen the same problem on unstable so this may be a bug outside of what I've changed. I'll have a look into seeing if I can fix it though.

c0bra:
1. Ah; I didn't realize the outline was actually required for focus+aria to work properly. Styling it is fine with me. The only *real* problem I have is that it's a bit clunky blue outline in Chrome and it looks rather bad.

2. - [ ] I did do a clean build but I can try it again and will let you know.

c0bra:
@JLLeitschuh OK it must have been some weird caching thing in Chrome. Looks good to me now!

c0bra:
I've tried setting up outline CSS to make it more stylistically consistent, but it doesn't seem to want to work in Chrome. When I set `outline: 1px solid blue` instead of the border going around the entire element it's just a single side, or only the left and right sides. Strange.

JLLeitschuh:
Outline on Chrome in OSX is being incredibly buggy.
 - [My stack overflow post](http://stackoverflow.com/questions/30787767/focus-border-not-appearing-on-html-button-or-select-in-google-chrome-only-osx-1)
 - [My google product fourms post](https://productforums.google.com/forum/?utm_campaign=q_and_a_en&utm_medium=en&utm_source=en#!topic/chrome/XccA4yP37EA;context-place=mydiscussions)

c0bra:
OK, so... what's our way forward?

JLLeitschuh:
What I'm in the process of finishing up:
- [x] Screen reader support to cell nav.
- [x] Full screen reader support to safari (fixed finally thanks to the accessibility inspector built into XCode)
- [ ] Small Close button at the top of the menus (focusable only with tab key) so you can close from the top and bottom of the menus.
- [ ] Hiding menus moves focus to neighbor cell header instead of onto the grid men (because the grid menu may not exist).
- [ ] Update docs to reflect these changes.

JLLeitschuh:
#3987 Created a conflict that needs to be resolved.
I'm running out of time to work on this project. Hopefully this will be the last time that I need to rebase.

c0bra:
LGTM

Issue #3795, opened 2015-06-17
[saveState | externalSorting] external sorting listener invoked without arguments on state restore


lpand:
Hi, external sorting and saveState plugin don't play nice together. I made a [demo](http://plnkr.co/edit/Ua02w1ltytbtWPoHo22v?p=preview) out of the saveState plugin tutorial.

1. change sorting of a column
2. save state
3. change sorting again
4. restore state

The sorting listener has been called without arguments while the column appears with the little triangle on it.

P.S. thanks for the great job

lpand:
This seems to be the guilty line:

https://github.com/angular-ui/ng-grid/blob/master/src/features/saveState/js/saveState.js#L552

I'll create a pull request if you agree.

PaulL1:
So the problem is that it didn't provide the sort column that changed in the call?  If that's what you're fixing then yeah, happily would accept a PR.

PR #3857, opened 2015-06-24
 #3795 raise sortChanged with proper arguments when grid state is restored

lpand:
This request is related to the issue numbered #3795.
`restoreColumns()` raises `sortChanged()` with proper arguments as the event doc says.


PaulL1:
Looks great thanks, merging.

Issue #3791, opened 2015-06-17
How to filter by display value instead of actual value?


senpost:
I have a gender column in my grid and its actual value is a integer and it gets transformed into strings such as Male, Female etc... Filtering won't work if I try to filter it by the text, filtering by actual integer value works fine.

How can I filter it by the text that is displayed rather than the value behind the text?

Here is the Plunker http://plnkr.co/edit/JgjKWsVWCV7v4hA8mak3?p=preview

Thank you.

PaulL1:
You can't filter without using a custom filter widget.  I'd recommend using the dropdown widget, which is shown in the http://ui-grid.info/docs/#/tutorial/306_custom_filters tutorial (for gender in fact).

senpost:
Thanks @PaulL1

mc0re:
Dropdowns are fine (although require some work if you want them translated and dynamic - pick only the values that exist in the data). But a full-text search on the values would also be really nice.

Could I suggest, that UI Grid adds the translated value to a helper column and uses it for standard text filtering, _instead_ of the actual value?

PR #3905, opened 2015-07-01
Enh(filter and sort) Support filter and sort by displayed value

### LINKED ISSUE #2839: enableFiltering with CellFilter
Hi,

More of a question than an issue perhaps,
but I'm working with a ui-grid where we want filtering on as a default with the default filtering to be contains (This was an easy fix, maybe something that could be configurable in the future?).

The problem we are currently facing are cellFilters defined on the grid with underlying Id's, 
causing the filtering to be applied to the underlying value and not the displayed value.

This can be illustrated by enabling filtering on 
201 - Edit feature
as in this plunker http://plnkr.co/edit/KnlHaCJX7aFvIINREkRq
where it is only possible to filter on 1 / 2, not male or female

I've currently solved this with a perhaps naive approach, and added a piece of code in the runColumnFilter-function that overwrites the entered value with a $eval'd value using the filter,
requiring the $filter-service in the appScope.

This naturally felt right quite a hack, so I'm wondering if there is any easier solution that I've overlooked,
or if this is something that could be considered as a future feature?

Br
/ Petter

AgDude:
Add getCellDisplayValue to grid core. This returns the cell value after any `cellFilter` has been applied.

Add `columnDef` options `filterCellFiltered` and `sortCellFiltered`, which apply the sort and filter respectively using `grid.getCellDisplayValue` instead of `grid.getCellValue`.

Closes #2839 #3791

[Review on Reviewable](https://reviewable.io/reviews/angular-ui/ng-grid/3905)

### DIFF
--- a/misc/tutorial/102_sorting.ngdoc
+++ b/misc/tutorial/102_sorting.ngdoc
@@ -30,6 +30,9 @@ asynchronously after the columns it will often decide all your columns are strin
 column def using `type='number'`.  Valid types are documented in {@link api/ui.grid.class:GridOptions.columnDef columnDef}, and
 include `string`, `number`, `numberStr` and `date`.  If you use date be aware the code expects a javascript date object.
 
+By default the sorting algorithm will be applied to the row value before any `cellFilters` are applied. The {@link api/ui.grid.class:GridOptions.columnDef#sortCellFiltered sortCellFiltered}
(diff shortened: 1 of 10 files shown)


PaulL1:
I don't have time to pull and test, but I've done a quick review and it looks good to me.  Basically you're using a $parse to get the displayed value so long as the user has used the standard cell template and filters setup.  Which is reasonable, but I wonder about the performance impact.  The fact this is an optional setup means that people can make their own decisions about that.

I'd merge happily given the unit tests you've provided, but I'll leave in case someone is inclined to pull and test rather than blindly merge as I would do.

PR #3987, opened 2015-07-15
add classes to pagination buttons

jbarrus:
Not sure if the class names follow whatever convention has been established, but it would be nice to have these available for both styling and e2e tests.

[Review on Reviewable](https://reviewable.io/reviews/angular-ui/ng-grid/3987)


k1ngdr3w:
Makes automation easier, thanks!

Issue #3960, opened 2015-07-10
Stable sorting should reverse


AgDude:
The sorting processor remembers the previous sort order, and preserves ordering each time it runs when two rows evaluate as equal. When the sort order is reversed, this preserved stable order should reverse.

For example, look at the first grid on the 102 tutorial: http://ui-grid.info/docs/#/tutorial/102_sorting

Sort the gender column so female is first. You should see that Ethel Price is first, followed by Claudine Neal. Now reverse the sort on the gender column. The expected behavior is that Ethel Price would be last, and Claudine Neal second-to-last. What happens instead is that Ethel Price is the first female, and Claudine Neal is the second female.

PR #3993, opened 2015-07-16
Sorting to asc from desc doesn't work when suppressRemoveSort is set to false

eeiswerth:
When clicking on a column to sort it (not via the menu dropdown), it sorts in asc order correctly. I then can click it again and it will sort in desc order correctly. If I then attempt to sort it asc again by clicking the column a 3rd time, nothing happens.  If I then set suppressRemoveSort = true for my column in the columnDef, then the sort works as expected.

I traced this back to the code in question. Removing the check for suppressRemoveSort fixes the issue for my particular case.  I do not fully understand what the purpose of this check is so I fully admit that this may regress some obscure state in the grid that I don't understand.  

In any case, I my change is correct and if the suppressRemoveSort check needs to be re-instated it should probably outside of this particular if/else block and should stand on its own.  

Definitely open to discussing, but the sorting does seem to function as expected with this change.

Thanks!

[Review on Reviewable](https://reviewable.io/reviews/angular-ui/ng-grid/3993)

### DIFF
--- a/src/js/core/factories/Grid.js
+++ b/src/js/core/factories/Grid.js
@@ -1957,9 +1957,9 @@ angular.module('ui.grid')
         column.sort.direction = uiGridConstants.DESC;
       }
       else if (column.sort.direction && column.sort.direction === uiGridConstants.DESC) {
-        if ( column.colDef && column.suppressRemoveSort ){
-          column.sort.direction = uiGridConstants.ASC;
-        } else {
+        column.sort.direction = uiGridConstants.ASC;
+
+        if (!(column.colDef && column.suppressRemoveSort)) {
           column.sort = {};
         }
       }


eeiswerth:
Caused a build failure.  I thought I ran the tests, but must have forgot. I'll close the PR for now, but I still think there is something strange here.

eeiswerth:
Looks related to issue #3960

eeiswerth:
This workaround worked for me:
```
$.each($scope.gridOptions.columnDefs, function(i, columnDef) {
      columnDef.suppressRemoveSort = true;
});
```

eeiswerth:
Updated the PR to fix the build.  Again, just my best guess as to how to address the issue. Thanks.

JLLeitschuh:
I think this needs to be rebased and squashed into one commit.

PaulL1:
I'm not certain that this is targeting the right behaviour.  At the moment sorting toggles through asc, desc, default sort.  This sounds like it removes default sort as an option (aka unsorted).  If so, I think that it would be a breaking change for many users.

eeiswerth:
Perhaps this is a misunderstanding on my part. For some reason I didn't expect it toggle 3 states. I expected it to toggle 2 states (asc/desc) once sorting was used. This could be based on my experience with previous grids with other frameworks.

Perhaps the issue I am encountering is only with infinite scroll.  I am happy to close this one and investigate a bit further. Something is definitely fishy somewhere, but you're right. I may not be targeting the correct behavior.  I'll close the PR in a few days if it's still open without comment.

PR #4002, opened 2015-07-19
Issue 3828 - footerCellClass function gets called only once

alam123:


c0bra:
You should only address one issue in a PR but it seems like you have two commits, each of which addresses a different issue.

And also tests are not passing. You can click the red Xs to see the output on Travis.

PR #4013, opened 2015-07-21
Issue 3828 - footerCellClass function gets called only once

alam123:


alam123:
Any chance of this getting merged?

Issue #3893, opened 2015-06-30
Follow DHTML Style Guide Working Group for Grid Keyboard Shortcuts


JLLeitschuh:
Would it be better if cell nav were to follow the following guidelines?
http://access.aol.com/dhtml-style-guide-working-group/#grid

I think it would fix some of the problems with being unable to use the arrow keys inside of a cell in a grid to manipulate dropdowns and input boxes.

This is an example of a grid that uses this keyboard setup. http://oaa-accessibility.org/example/18/

swalters:
We modeled the keyboard navigation on excel and google spreadsheet.  Enter in spreadsheets go to next cell which isn't listed in this doc.

bialikover:
Are there any standard in grouping and other features? I've found some issues when implementing grouping and keyboard navigation together

mportuga:
Note: Add help wanted label

PR #4020, opened 2015-07-21
fix(cellNav): keyboard navigation while editing cell value

robinbolton:
Allow keyboard navigation using Up/Down/Left/Right within a form field when the cellNavigation plugin is being used, as described in Tutorial 202.

BREAKING CHANGE: Function signature of uiGridCellNavService.getDirection changes from getDirection(evt) to getDirection(grid, evt) in order to test for status of grid.api.edit feature.

### DIFF
--- a/src/features/cellnav/js/cellnav.js
+++ b/src/features/cellnav/js/cellnav.js
@@ -454,7 +454,17 @@
          * @description  determines which direction to for a given keyDown event
          * @returns {uiGridCellNavConstants.direction} direction
          */
-        getDirection: function (evt) {
+        getDirection: function (grid, evt) {
+          // Limit navigation during edit operations
+          if (grid.api.edit) {
+            var el = document.activeElement;
+            var tagName = el.tagName.toLowerCase();
+            if ((tagName === 'input' || tagName === 'select' || tagName === 'textarea') &&
+              (evt.keyCode !== uiGridConstants.keymap.ENTER)) {
+                return null;
+            }
+          }
+
           if (evt.keyCode === uiGridConstants.keymap.LEFT ||
             (evt.keyCode === uiGridConstants.keymap.TAB && evt.shiftKey)) {
             return uiGridCellNavConstants.direction.LEFT;
@@ -687,8 +697,9 @@
               };
 
               uiGridCtrl.cellNav.handleKeyDown = function (evt) {
-                var direction = uiGridCellNavService.getDirection(evt);
-                if (direction === null) {
+                var grid = uiGridCtrl.grid;
+                var direction = uiGridCellNavService.getDirection(grid, evt);
+                 if (direction === null) {
                   return null;
                 }
 
--- a/src/features/cellnav/test/uiGridCellNavService.spec.js
+++ b/src/features/cellnav/test/uiGridCellNavService.spec.js
@@ -77,7 +77,7 @@ describe('ui.grid.edit uiGridCellNavService', function () {
     });
   });
 
-  describe('getDirection(evt)', function () {
+  describe('getDirection(grid, evt)', function () {
     beforeEach(function(){
       grid.registerColumnBuilder(uiGridCellNavService.cellNavColumnBuilder);
       grid.buildColumns();
@@ -87,7 +87,7 @@ describe('ui.grid.edit uiGridCellNavService', function () {
       evt.keyCode = uiGridConstants.keymap.TAB;
       var colDef = grid.options.columnDefs[0];
       var col = grid.columns[0];
-      var direction = uiGridCellNavService.getDirection(evt);
+      var direction = uiGridCellNavService.getDirection(grid, evt);
       expect(direction).toBe(uiGridCellNavConstants.direction.RIGHT);
     });
 
@@ -96,7 +96,7 @@ describe('ui.grid.edit uiGridCellNavService', function () {
       evt.keyCode = uiGridConstants.keymap.RIGHT;
       var colDef = grid.options.columnDefs[0];
       var col = grid.columns[0];
-      var direction = uiGridCellNavService.getDirection(evt);
+      var direction = uiGridCellNavService.getDirection(grid, evt);
       expect(direction).toBe(uiGridCellNavConstants.direction.RIGHT);
     });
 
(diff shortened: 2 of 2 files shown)


PaulL1:
In concept this looks fine, but I'm aware that the navigation within a cell whilst editing is both fragile and has reasonably complex semantics.  I think we're better to have someone pull this and test it to verify behaviour than to merge it and then raise defects after.

JLLeitschuh:
Any good way we could create a regression test for this?

swalters:
I'm assuming this breaks our current excel-like editing experience.  
type text, enter, next cell, type text, enter......

If it breaks this behavior, then we will need another approach.  maybe an option to enable this behavior you are wanting.

What I do when I want to ignore Enter in my editor is stack another ui-grid-editor directive and modify the default bahavior.

JLLeitschuh:
@swalters What do you think about my suggestion regarding a new set of cell nav keys?
#3893

robinbolton:
According to the Tutorial: 202 Cell Navigation documentation - "When combined with the editable feature, the left-right arrow keys will be subsumed when in "deep edit" mode, allowing you to move around within the text you're editing. The tab key and shift-tab keys continue to function."

However my experience was that this is not correct as using the arrow keys while editing a cell would simply navigate to another cell NOT within the cell currently being edited as one should expect.  This was extremely frustrating and didn't suit my nor my users' expectations or requirements.  The PR was merely a naive attempt at resolving the issue and having the grid act in the manner described in your documentation.

If the implementation is unacceptable then please either develop an officially recognized solution or alter the documentation to describe the actual functionality as exists.


Robin

swalters:
@robinbolton deep edit is only invoked when you click inside the editor.  Are you not seeing this behavior?  It is working correctly for me on 401 tutorial.

This behavior is how Excel and Google Sheets manage the arrow keys which is what we used to design editing behavior.

JLLeitschuh:
@swalters That isn't accessible to someone who is only able to use a keyboard.

swalters:
Excel allows use of arrow keys if you initiate editing via F2.  What if we invoked deep edit automatically on f2?

robinbolton:
I don't have a current version of Excel installed, but using LibreOffice as my test platform, which I believe works the same as Excel in this manner, the following occurs:

When pressing F2 to edit a cell:
- The cursor is placed at the end of the cell in edit mode.
- Using the left/right arrow keys navigates left/right within the current cell.
- Using the up/down arrow keys does nothing.
- Typing a letter or number inserts at the point of the cursor.
- Double clicking highlights all text within the cell and remains in edit mode.
- Pressing enter exits edit mode and moves down to the cell immediately below.
- Pressing tab exits edit mode and moves to the cell immediately right.
- Pressing shift-tab exits edit mode and moves to the cell immediately left.

When double clicking to edit a cell:
- The cursor is placed at the point within the text where the double-click occurred.
- Using the left/right arrow keys navigates left/right within the current cell.
- Using the up/down arrow keys does nothing.
- Typing a letter or number inserts at the point of the cursor.
- Double clicking highlights all text within the cell and remains in edit mode.
- Pressing enter exits edit mode and moves down to the cell immediately below.
- Pressing tab exits edit mode and moves to the cell immediately right.
- Pressing shift-tab exits edit mode and moves to the cell immediately left.

Using Google Docs:

When pressing F2 to edit a cell:
- The cursor is placed at the end of the cell in edit mode.
- Using the left/right arrow keys navigates left/right within the current cell.
- Using the up/down arrow keys jump to the beginning/end of the current cell.
- Typing a letter or number inserts at the point of the cursor.
- Double clicking highlights all text within the cell and remains in edit mode.
- Pressing enter exits edit mode and moves down to the cell immediately below.
- Pressing tab exits edit mode and moves to the cell immediately right.
- Pressing shift-tab exits edit mode and moves to the cell immediately left.

When double clicking to edit a cell:
- The cursor is placed at the end of the cell in edit mode.
- Using the left/right arrow keys navigates left/right within the current cell.
- Using the up/down arrow keys jump to the beginning/end of the current cell.
- Typing a letter or number inserts at the point of the cursor.
- Double clicking highlights all text within the cell and remains in edit mode.
- Pressing enter exits edit mode and moves down to the cell immediately below.
- Pressing tab exits edit mode and moves to the cell immediately right.
- Pressing shift-tab exits edit mode and moves to the cell immediately left.


Using Tutorial 401 in Google Chrome 44 on OSX 10.10.4 the following occurs:

When pressing F2 to edit a cell:
- All text within the cell is selected.
- Using the left/right arrow keys navigates to the next cell, left or right. 
- Using the up/down arrow keys navigates to the next cell, up or down. 
- Typing a letter or number replaces all text within the cell with the typed character.
- Double clicking highlights all text within the cell and remains in edit mode.
- Pressing enter exits edit mode and moves down to the cell immediately below.
- Pressing tab exits edit mode and moves to the cell immediately right.
- Pressing shift-tab exits edit mode and moves to the cell immediately left.

When double clicking to edit a cell:
- All text within the cell is selected.
- Using the left/right arrow keys navigates to the next cell, left or right. 
- Using the up/down arrow keys navigates to the next cell, up or down. 
- Typing a letter or number replaces all text within the cell with the typed character.
- Double clicking highlights all text within the cell and remains in edit mode.
- Pressing enter exits edit mode and moves down to the cell immediately below.
- Pressing tab exits edit mode and moves to the cell immediately right.
- Pressing shift-tab exits edit mode and moves to the cell immediately left.

It appears that the problem lies with the fact that all text is automatically selected when entering edit mode and the UI Grid and keyboard navigation does not function properly until the text is replaced, which is not how LibreOffice/Excel/Google Docs function.  


Robin

swalters:
I've created #4088 to implement these changes. If that works for you, we can continue the conversation in that issue.

robinbolton:
Perfect, I'll just keep using my own patched version until this is officially resolved.


Cheers,
Robin

jesseolsen:
@chzmnky Can you share your patched version?  I'd like the user to go directly to deep edit mode when they double click on a cell...

adamebra:
Issue #4088 has been staled. Is anyone planning on picking up this issue? Why is this issue not gaining more attention? It seems to be a basic feature that does not agree with what the documentation says.

nandoflorestan:
This is the single most important usability issue in the entire grid and nobody has done anything in 5 years!!!

PR #4021, opened 2015-07-21
paginationPageSizes as label if singular

kencaron:
Seems like a given that there shouldn't be a dropdown if paginationPageSizes only contains a single option. This renders as a span if `grid.options.paginationPageSizes.length <= 1` and only renders a dropdown if `> 1`.


JLLeitschuh:
I think this makes sense.
Devs: If you've seen this PR and don't see any problem with it leave a LGTM comment.

PaulL1:
Looks reasonable.  Merging.

PR #4028, opened 2015-07-22
fixed removeViewportAdjuster bug

littledice:
- renamed wrongly named 'registerViewportAdjuster' fn to 'removeViewportAdjuster'
- when checking if a function is in the viewportAdjusters array need to check that its index is > -1


JLLeitschuh:
Can you rebase your changes on to the top of master so that they are only one commit.

PR #4038, opened 2015-07-23
Wrong url in README.md

lukasz-madon:
Grunt build uses pkg.name which is `ui-grid`. Documentation uses . as a separator. 

- Change url in README.md


JLLeitschuh:
Good catch. Thank you!

Issue #3181, opened 2015-04-01
custom Editable Cell Template with ng-model-options


dardino:
hi,
I have a problem with my custom Cell Template:
```html
<div>
   <form name="inputForm">
      <input ng-model-options="{ updateOn: 'blur' }" type="INPUT_TYPE" ui-grid-editor ng-class="'colt' + col.uid" ng-model="MODEL_COL_FIELD" />
   </form>
</div>
```
this template differ with default only for ng-model-options but when I leave from filed (blur) the model is not updated.

is it a bug or the binding on blur not supported?

Thanks a lot!
Gabriele

dardino:
Hi, anyone has this problem?

bencameron:
Yes, I've run into this problem too.  Before finding the issue you created, I tried the same approach (custom cell template) and ran into the same problem.  I got rid of the custom template and am just allowing the model to be updated from the view after each keystroke for now.  That's definitely not ideal for my application, but I've got other priorities for now and was hoping this might get resolved by the time I circle back to it.

Have you had any luck?

dardino:
Hi! Thanks for the reply.
For now I have placed the event raise instruction into a $timeout callback with a delay of 10 msec.
Doing this, angular can commit the value before the end cell edit destroys the DOM element and its directive.
This is not a good choice but for now is the fastest way to get around the problem.

PaulL1:
The recommended way to handle update on every keypress is to do a debounce, rather than trying to change things so that there isn't an update on every keypress.  I'm not familiar with the option that you're setting to cause it to update only on blur, but I do recall discussion of a problem where angular by default doesn't have a blur event, perhaps that is the problem.

PaulL1:
Looks like original raiser has found a solution, no further updates for some time, closing.

dardino:
Good! thanks!
I'll try it as soon as possibile

aiguofer:
I was having this issue. Unfortunately the linked solution didn't work for me. I have submitted a PR #4698 with a solution that's working for me.

PR #4058, opened 2015-07-27
Fixed ng-model-options="{ updateOn: 'blur'}" on custom edit templates

mikebm:
I have re-ordered the endEdit() code to trigger the grid focus before the scope destruction and edit control removal. This will allow for the blur event to fire and update the scope.

This will allow ng-model-options="{ updateOn: 'blur'}" on custom edit templates to work.

### DIFF
--- a/src/features/edit/js/gridEdit.js
+++ b/src/features/edit/js/gridEdit.js
@@ -782,6 +782,14 @@
               if (!inEdit) {
                 return;
               }
+              
+              //sometimes the events can't keep up with the keyboard and grid focus is lost, so always focus
+              //back to grid here. The focus call needs to be before the $destroy and removal of the control,
+              //otherwise ng-model-options of UpdateOn: 'blur' will not work.
+              if (uiGridCtrl && uiGridCtrl.grid.api.cellNav) {
+                uiGridCtrl.focus();
+              }
+              
               var gridCellContentsEl = angular.element($elm.children()[0]);
               //remove edit element
               editCellScope.$destroy();
@@ -790,11 +798,6 @@
               inEdit = false;
               registerBeginEditEvents();
               $scope.grid.api.core.notifyDataChange( uiGridConstants.dataChange.EDIT );
-              //sometimes the events can't keep up with the keyboard and grid focus is lost, so always focus
-              //back to grid here
-              if (uiGridCtrl && uiGridCtrl.grid.api.cellNav) {
-                uiGridCtrl.focus();
-              }
             }
 
             function cancelEdit() {


mikebm:
This will also fixed the closed #3181

JLLeitschuh:
This makes sense to me. I don't see any problem merging it.

swalters:
Please rebase your commits into one commit and we'll merge.  this help keep our commit history clean. thanks.

mikebm:
Updated. Thanks.

JLLeitschuh:
@swalters Looks good to me. Feel free to merge.

framundo:
@mikebm this is not working for me, blur event doesn't seem to be called. *Angular 1.4.6* and *ui-grid 3.0.6*. Any suggestions? Is there another workaround?

mikebm:
I haven't looked, but I am assuming 3.0.6 has my merged changed.

In my case, I was using a custom template for my controls. This was working
in my case:

 $templateCache.put('stdGridTextBox',
          "<div><form name=\"inputForm\"><input type=\"INPUT_TYPE\"
ng-class=\"'colt' + col.uid\" ui-grid-editor ng-model=\"MODEL_COL_FIELD\"
ng-model-options=\"{ updateOn: 'blur' }\"></form></div>"
        );

On my column definition, i would do the following:
columnDef.editableCellTemplate = $templateCache.get(control.GridTemplate);
columnDef.allowCellFocus = true;
columnDef.allowCellEdit = true;


I believe one of the key things that you need to inject as well is
ui.grid.cellNav.

The dependecies I was using were:

        'ui.grid',
        'ui.grid.resizeColumns',
        'ui.grid.edit',
        'ui.grid.cellNav',

Without this change, editing a field and changing the data, then tabbing to
the next field would revert my changes.

Hope this helps.

framundo:
Yes, it has your change. The template is the same also.
I correct myself: blur is called if I click outside the editing cell, but If i hit enter or tab, it doesn't.

mikebm:
Right, that is the issue i had without the change. Clicking worked, tabbing
did not.

You have the cellNav dependency right? Wonder if something broke this in
recent changes. I haven't tested it with the latest yet. The issue was the
control was getting destroyed before the focus left it, which caused the
values to never get pushed. Are you able to debug the source and verify the
focus is getting called?

aiguofer:
So I was having this same problem and I found a workaround by calling `editCellScope.inputForm.$commitViewValue()` if it's not a `cancelEdit()`:

```
               function endEdit(cancel) {
                  $scope.grid.disableScrolling = false;
                  if (!inEdit) {
                     return;
                  }

                  //sometimes the events can't keep up with the keyboard and grid focus is lost, so always focus
                  //back to grid here. The focus call needs to be before the $destroy and removal of the control,
                  //otherwise ng-model-options of UpdateOn: 'blur' will not work.
                  if (uiGridCtrl && uiGridCtrl.grid.api.cellNav) {
                     uiGridCtrl.focus();
                  }

                  var gridCellContentsEl = angular.element($elm.children()[0]);
                  //remove edit element

                  if (!cancel)
                     editCellScope.inputForm.$commitViewValue();

                  editCellScope.$destroy();
                  angular.element($elm.children()[1]).remove();
                  gridCellContentsEl.removeClass('ui-grid-cell-contents-hidden');
                  inEdit = false;
                  registerBeginEditEvents();
                  $scope.grid.api.core.notifyDataChange(uiGridConstants.dataChange.EDIT);
               }

               function cancelEdit() {
                  $scope.grid.disableScrolling = false;
                  if (!inEdit) {
                     return;
                  }
                  cellModel.assign($scope, origCellValue);
                  $scope.$apply();

                  $scope.grid.api.edit.raise.cancelCellEdit($scope.row.entity, $scope.col.colDef);
                  endEdit(true);
               }
```

I'm not familiar enough with the rest of the code to determine if this has any side effects, but it seems to be working for me. If this is a good solution I can put in a PR.

PR #4060, opened 2015-07-28
Makes AllTemplates Multiline

JLLeitschuh:
Makes all of the templates multiline to reduce the number of merge conflicts. This also makes git blame easier to use as each tag change will be line dependent.
This makes the entire code base much cleaner.


c0bra:
LGTM

Issue #4010, opened 2015-07-20
E2E Tests: Failure + Safari 5 Incompatibility


JLLeitschuh:
The e2e tests fail because 103 has one more column then the test expects.
The e2e tests aren't compatible with Safari 5.
See my comments on this commit:
https://github.com/angular-ui/ui-grid/commit/f71f9dc2f19925641585eadb1de4ff2839afe886

AgDude:
A while back we decided to drop support for old (non evergreen) browsers. Closing as won't fix.

PR #4094, opened 2015-07-31
Fix #4046 - Last column disappearing upon edit

### LINKED ISSUE #4046: Last column disappearing upon edit
### Issue
Using a table with editing and cellnav with many columns (in this case it's 25) if I scroll to the last column and edit a field in the table, the last column disappears.

To reproduce, you need to have the browser in full screen (if using Chrome, Firefox is not working even when windowed).

Furthermore, it's working correctly if you run it from inside pluker edit mode.

I'm going on instinct here since we just started looking at ui-grid, but it seems to me that some measurement is failing and it believes the rightmost columns should be falling outside of the visible part of the table, so it doesn't render them.

### Environment
I tried with Firefox (38.0.5) and Chrome (44.0.2403.89 m) on Windows 7 Enterprise, it happens on both. Tried also with clients desktop using Chrome (don't know their version but quite up to date) on Windows XP, it was also affected.

Tried with both angularjs 1.4.3 and 1.3.0, I saw no difference.

### Plunker
http://run.plnkr.co/plunks/PU8B6By5kZfy1rAqfXml/

### Images
image
image

imbalind:
I ran the tests and I got:
```
Failures:

  1) ui-grid tutorial/103_filtering first grid on the page, filtered by male by default grid should have seven visible columns
   Message:
     Expected 8 to equal 7.
```
which I have even without my changes.

Do I need to provide anything else?

### DIFF
--- a/src/js/core/factories/GridRenderContainer.js
+++ b/src/js/core/factories/GridRenderContainer.js
@@ -485,9 +485,10 @@ angular.module('ui.grid')
     var columnCache = self.visibleColumnCache;
     var maxColumnIndex = columnCache.length - minCols;
 
-    // Calculate the scroll percentage according to the scrollTop location, if no percentage was provided
+    // Calculate the scroll percentage according to the scrollLeft location, if no percentage was provided
     if ((typeof(scrollPercentage) === 'undefined' || scrollPercentage === null) && scrollLeft) {
-      scrollPercentage = scrollLeft / self.getCanvasWidth();
+      var horizScrollLength = (self.getCanvasWidth() - self.getViewportWidth());
+      scrollPercentage = scrollLeft / horizScrollLength;
     }
 
     var colIndex = Math.ceil(Math.min(maxColumnIndex, maxColumnIndex * scrollPercentage));


JLLeitschuh:
I don't understand what this is doing. I do know that that is a known failing test #4010 so you are all good on that front.

imbalind:
What do you mean by:
> I don't understand what this is doing.

?

JLLeitschuh:
Basically I don't understand the code that you've modified. One of the other devs should take a look at this before merging. I don't feel like I understand that bit of the code well enough to add a LGTM.

swalters:
This is being recorded in our application as well.  I'm doing some research as to why this started happening. The code this addresses hasn't been touched in a very long time.  Either this bug was always with us or something has changed elsewhere to cause this issue.

swalters:
@imbalind thanks for the detailed issue and PR.  I'm sure it would have taken me forever to even figure out how to recreate this bug.

JLLeitschuh:
Closes #4046

PR #4097, opened 2015-07-31
feat(core): Add Clear all filters to grid menu

mboriani:
I add a new item on menu that clear all the filters. It is available only if you enabledFiltering option.
To used you have to go to menu and you will see it at the top


swalters:
duplicate

PR #4132, opened 2015-08-06
feat(exporter): add exporterPdfPrefix and exporterPdfPostfix options

### LINKED ISSUE #4073: PDF export prefix and postfix
I'm trying to display several lines of text above the exported table. Everything displayed in header is cut by the height of this header, and if I'm not mistaken I can set its size only for ALL pages, using code such as

exporterPdfCustomFormatter: function (docDefinition) {
                  docDefinition.pageMargins = [40, 200, 40, 40];
                  return docDefinition;
                },

But in fact I don't need such a huge margin for all pages, only for the first one where I display report title and parameters. And it's not correct even to regard it as "margin" or "header", it's not header, it's grid export prefix, it is not intended to be displayed in page header, it's a regular part of a document.

Using  exporterPdfTableStyle: {margin: [40, 200, 40, 40]}, it is possible to introduce an empty space before the grid, but it will remain empty, because there's no way to put anything in it.

Looks like the only way to handle is problem is by introducing two additional gridOptions, exporterPdfPrefix and exporterPdfPostfix. Is it a valid feature request or there's a better way to handle this problem? This situation is quite common (for example, when you need to display several report parameters at the beginning of the report), and support for this will add your pdf exporting a lot of flexibility.

avitenberg:
Adds two grid options, exporterPdfPrefix and exporterPdfPostfix,
allowing to display custom content in pdf export
before and after the exported grid

Resolves #4073

### DIFF
--- a/misc/tutorial/206_exporting_data.ngdoc
+++ b/misc/tutorial/206_exporting_data.ngdoc
@@ -72,6 +72,19 @@ In this example we use the native grid menu buttons, and we show both the pdf an
         exporterPdfOrientation: 'portrait',
         exporterPdfPageSize: 'LETTER',
         exporterPdfMaxGridWidth: 500,
+        exporterPdfPrefix : [
+          {
+            text: 'My prefix - custom title of pdf export',
+            style: "headerStyle"
+          },
+          {
+            text: 'Prefix can contain several differently formatted sections',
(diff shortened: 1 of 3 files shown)


avitenberg:
closed it myself, see comments in #4073

PR #4144, opened 2015-08-08
Support function templates

mirik123:
ui-grid fails when template is defined as a function, for example:
````
$templateCache.put('ui-grid/uiGridCell', function () {
   return "<div class=\"ui-grid-cell-contents\" title=\"TOOLTIP\">{{COL_FIELD CUSTOM_FILTERS}}</div>"
});
````

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+<a name="v3.0.4"></a>
+### v3.0.4 (2015-08-13)
+
+
+#### Bug Fixes
+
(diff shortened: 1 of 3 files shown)


JLLeitschuh:
Looks great! I like it!
Can you squash it into one commit though?
Also, if you have the time adding a test case would be really great too.

mirik123:
I'm sorry, I've been totally messed with git, so I better create a new pull request.

mirik123:
opened new pull request already squashed #4191.

PR #4150, opened 2015-08-10
extended gridedit api to allow invoking cell editor by row/column

justin-romano:
new to pull requests, hope im doing it right.
I have not put much here  because im hoping the documentation on the methods is mostly enough.
My main concern is around security. Im mostly used to working in Dojo which does not hide much and was hoping that adding an array of column scopes column directive to the row object was in fitting with the way things are done here.
```
            if (!$scope.row.cols) {
              $scope.row.cols = [];
            }
            $scope.row.cols[$scope.colRenderIndex] = { "$scope": $scope, "$elm": $elm };
```

### DIFF
--- a/src/features/edit/js/gridEdit.js
+++ b/src/features/edit/js/gridEdit.js
@@ -114,8 +114,9 @@
             methods: {
               edit: {
                 /**
-                 * @ngDoc method
+                 * @ngDoc function
                  * @name editCell
+                 * @methodOf ui.grid.edit.api:PublicApi
                  * @description invokes a cells editor and optionally allows the edit to persist
                  * <pre>
                  *      gridApi.edit.editCell(row, col, true);
@@ -131,8 +132,9 @@
                   this.rows[row].cols[col].$scope.$broadcast('editCell');
                 },
                 /**
-                 * @ngDoc method
+                 * @ngDoc function
                  * @name editEditCell
+                 * @methodOf ui.grid.edit.api:PublicApi
                  * @description cancels a cells editor if persisted by editCell(...)
                  * <pre>
                  *      gridApi.edit.endEditCell(row, col);
@@ -519,7 +521,7 @@
             if (!$scope.row.cols) {
               $scope.row.cols = [];
             }
-            $scope.row.cols[$scope.colRenderIndex] = { "$scope": $scope, "$elm": $elm };
+            $scope.row.cols.push({ "$scope": $scope, "$elm": $elm });
 
             // Bind to keydown events in the render container
             if (uiGridCtrl && uiGridCtrl.grid.api.cellNav) {
--- a/src/features/edit/js/gridEdit.js
+++ b/src/features/edit/js/gridEdit.js
@@ -126,10 +126,11 @@
                  * @param {boolean} persist the cell in edit mode. Prevent cancel events such as scroll from removing the editor
                  */
                 editCell: function (row, col, persist) {
+                  var cellScope = this.element.find("div[id*=" + row.toString() + "-" + this.columns[col].uid + "]").scope();
                   if (persist) {
-                    this.rows[row].cols[col].$scope.persistEdit = true;
+                    cellScope.persistEdit = true;
                   }
-                  this.rows[row].cols[col].$scope.$broadcast('editCell');
+                  cellScope.$broadcast('editCell');
                 },
                 /**
                  * @ngDoc function
@@ -143,8 +144,9 @@
                  * @param {integer} col of the cell to edit
                  */
                 endEditCell: function (row, col) {
-                  this.rows[row].cols[col].$scope.persistEdit = false;
-                  this.rows[row].cols[col].$scope.$broadcast('endEditCell', false);
+                  var cellScope = this.element.find("div[id*=" + row.toString() + "-" + this.columns[col].uid + "]").scope();
+                  cellScope.persistEdit = false;
+                  cellScope.$broadcast('endEditCell', false);
                 }
 
               }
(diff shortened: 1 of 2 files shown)


justin-romano:
So what happens now?

src/features/edit/js/gridEdit.js
```diff
@@ -112,12 +112,45 @@
               }
             },
             methods: {
-              edit: { }
+              edit: {
+                /**
+                 * @ngDoc method
+                 * @name editCell
```
JLLeitschuh: You need a @methodOf param here.
Example:
https://github.com/angular-ui/ui-grid/blob/master/src/features/cellnav/js/cellnav.js#L418

src/features/edit/js/gridEdit.js
```diff
@@ -112,12 +112,45 @@
...
+                 */
+                editCell: function (row, col, persist) {
+                  if (persist) {
+                    this.rows[row].cols[col].$scope.persistEdit = true;
+                  }
+                  this.rows[row].cols[col].$scope.$broadcast('editCell');
+                },
+                /**
+                 * @ngDoc method
+                 * @name editEditCell
```
JLLeitschuh: You need a @methodOf param here.
Example:
https://github.com/angular-ui/ui-grid/blob/master/src/features/cellnav/js/cellnav.js#L418

src/features/edit/js/gridEdit.js
```diff
@@ -112,12 +112,45 @@
               }
             },
             methods: {
-              edit: { }
+              edit: {
+                /**
+                 * @ngDoc method
```
JLLeitschuh: This should be function not method

src/features/edit/js/gridEdit.js
```diff
@@ -112,12 +112,45 @@
...
+                 * @param {boolean} persist the cell in edit mode. Prevent cancel events such as scroll from removing the editor
+                 */
+                editCell: function (row, col, persist) {
+                  if (persist) {
+                    this.rows[row].cols[col].$scope.persistEdit = true;
+                  }
+                  this.rows[row].cols[col].$scope.$broadcast('editCell');
+                },
+                /**
+                 * @ngDoc method
```
JLLeitschuh: This needs to be `function` not `method`

JLLeitschuh:
If it's not too much to ask it would be really great if you could add a unit and/or e2e test for this.

justin-romano:
I was going to wait to see if you guys would accept it before a put time
into that. Will do ASAP

src/features/edit/js/gridEdit.js
justin-romano: Sorry nube to this framework. so in the context of this framework what is the definition of a method?
In my understanding a function is not a property of an object.

justin-romano: e.g.
```
/**
* @ngdoc function
* @name angular.merge
* @module ng
* @kind function
*
* @description
* Deeply extends the destination object `dst` by copying own enumerable properties from the `src` object(s)
* to `dst`. You can specify multiple `src` objects. If you want to preserve original objects, you can do so
* by passing an empty object as the target: `var object = angular.merge({}, object1, object2)`.
*
* Unlike {@link angular.extend extend()}, `merge()` recursively descends into object properties of source
* objects, performing a deep copy.
*
* @param {Object} dst Destination object.
* @param {...Object} src Source object(s).
* @returns {Object} Reference to `dst`.
*/
function merge(dst) {
  return baseExtend(dst, slice.call(arguments, 1), true);
}
```

justin-romano: vs...
```javascript
 /**
         * @ngdoc method
         * @name $cookies#put
         *
         * @description
         * Sets a value for given cookie key
         *
         * @param {string} key Id for the `value`.
         * @param {string} value Raw value to be stored.
         * @param {Object=} options Options object.
         *    See {@link ngCookies.$cookiesProvider#defaults $cookiesProvider.defaults}
         */
        put: function(key, value, options) {
          $$cookieWriter(key, value, calcOptions(options));
        },
```

swalters:
I don't think we would ever assign anything column related to the row. It feels very wrong to me.  The marriage of Rows and Columns is done through a nested repeat, so you have $scope.row and $scope.col on every cell directive.

What are you trying to accomplish?

justin-romano:
Thank Shane. I know exactly what you mean. Feels dirty to me but given the code structure I could not easily think of a way to get the scope for a column so I could then apply it to the cell editor. For my project I want to invoke the editor of a cell so when I add a new row I can place a cell in edit mode for convenience of ui. Let me know if you can think of a better way

justin-romano:
so how about this approach?
```javascript
                editCell: function (row, col, persist) {
                  var cellScope = this.element.find("div[id*=" + row.toString() + "-" + this.columns[col].uid + "]").scope();
                  if (persist) {
                    cellScope.persistEdit = true;
                  }
                  cellScope.$broadcast('editCell');
```

JLLeitschuh:
Make sure the the element you are trying to grab isn't a directive that has `replace=true`. Don't know if this is at all related but you may just want to check that.

justin-romano:
Thanks. Its set in the uiGrid service but only after rendering complete. I may need to mock it the tests if thats allowed

JLLeitschuh:
Thats today in 2 hours

swalters:
@justin-romano this looks much better to me.  You have some failing jshint's.  also, please rebase to squash the commits. thanks

JLLeitschuh:
Looks good. Take care of @swalters requests and we'll get this merged up.

swalters:
@justin-romano will this work when angular debug info is turned off?
https://docs.angularjs.org/guide/production

mportuga:
Closing this due to inactivity.

PR #4164, opened 2015-08-12
Fix homepage tutorial link taking you to localhost

stoneG:
Updated the link to match the tutorial href on the nav bar


JLLeitschuh:
Seems to work for me locally. Don't think this will break the site somehow. Merging.

PR #4172, opened 2015-08-12
feat(core): Add Clear all filters to grid menu

mboriani:
I add a new menu label action call "Clear all filters". This action clear all existing filters in the grid. To enable you have to have enableFiltering = true.

### DIFF
--- a/src/features/cellnav/test/uiGridCellNavService.spec.js
+++ b/src/features/cellnav/test/uiGridCellNavService.spec.js
@@ -162,20 +162,20 @@ describe('ui.grid.edit uiGridCellNavService', function () {
 
   describe('scrollTo', function () {
     /*
-     * We have 11 rows (10 visible) and 11 columns (10 visible).  The column widths are 
+     * We have 11 rows (10 visible) and 11 columns (10 visible).  The column widths are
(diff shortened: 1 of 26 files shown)


swalters:
Please don't commit files with only whitespace changes

JLLeitschuh:
Sometimes you don't really have a choice.
Please fix your code to make sure that the tests don't fail.

mboriani:
Look like when i run grunt karmangular --browsers=Chrome it is failing that test but on the opposite way.

JLLeitschuh:
Ah. I see what @swalters was referring to. Your commit has whitespace changes in files you didn't otherwise modify. Can you please remove those files from your PR and squash them into one?
Thanks!

src/features/cellnav/test/uiGridCellNavService.spec.js
```diff
@@ -290,8 +290,8 @@ describe('ui.grid.edit uiGridCellNavService', function () {
...
       });
       $timeout.flush();
-      
+
       expect(args).toEqual( null );
     });
   });
-});
\ No newline at end of file
+});
```
JLLeitschuh: The only changes to this file are whitespace changes. Please remove from your PR.

PR #4179, opened 2015-08-13
Fix ttf path by removing '?'.  Resolves webpack issue.

### LINKED ISSUE #4113: In the css file:  in@font-face the '?' in url('ui-grid.ttf?') confuses webpack loader
Webpack throws an exception because it does not find the path of 'ttf?'.
When the '?' is removed it works fine

Please remove it if not needed
Thanks

mtraynham:
This fixes #4113. Simple change.

### DIFF
--- a/src/less/icons.less
+++ b/src/less/icons.less
@@ -3,7 +3,7 @@
   src: url('@{font-path}ui-grid.eot');
   src: url('@{font-path}ui-grid.eot#iefix') format('embedded-opentype'),
        url('@{font-path}ui-grid.woff') format('woff'),
-       url('@{font-path}ui-grid.ttf?') format('truetype'),
+       url('@{font-path}ui-grid.ttf') format('truetype'),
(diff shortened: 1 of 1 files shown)


JLLeitschuh:
Can you ensure that this doesn't break on some browser?
http://www.paulirish.com/2009/bulletproof-font-face-implementation-syntax/

mtraynham:
Seems the initial commit had it:
https://github.com/angular-ui/ui-grid/commit/ce330978fe566e9695c30acef04e55221e520960#diff-46456a3b2c603089887395d6c827b689R6

It's just a path reference. Usually you would have query parameters after the `?` to cache bust, like so: `url('ui-grid.ttf?v=20')`

mtraynham:
I'm actually able to resolve this with webpack, but it was a slight hiccup and felt the question mark is unnecessary...

JLLeitschuh:
I'll trust it. I asked some other people with more CSS skills than my own and they approve.

mtraynham:
thanks! :smile:

Issue #3925, opened 2015-07-03
Bower.json dependency resolves to Angular 1.2.28 instead of 1.3.16


danielstankiewicz:
I use `"angular-ui-grid": "3.0.0-rc.22"` in my `bower.json` file while I'm using `"angular": "~1.3.0"` (currently resolved to 1.3.16) as the base for my app.

This causes angular-ui-grid not to install, because there's conflict. Adding `"resolutions"` helps to fix it and reveals the true problem: 

```
Please note that,
    angular-ui-grid#3.0.0-rc.22 depends on angular#>=1.2.16 <=1.3.x which resolved to angular#1.2.28
```

Maybe you'd better define dependency as `"angular": ">=1.2.16 <1.4"` or something like that?

c0bra:
Do you have any idea why the version spec that we use: `>=1.2.16 <=1.3.16` doesn't work? Shouldn't that allow any version between 1.2.16 and 1.3.16? I have confirmed this with manual testing using semver.js: http://plnkr.co/edit/dHvHpgK53mtEcgI2goDQ?p=info  but for some reason it doesn't work with Bower.

danielstankiewicz:
Yes, I've got an idea. Bower uses it's own slightly modified version of `semver`, called `bower-semver` https://github.com/bower/semver . I tried to install and run it from node:

`semver.maxSatisfying(['1.3.16', '1.2.18'], '>=1.2.16 <=1.3.16')`

and the result was `1.2.18`, while in common `semver` it results with `1.3.16`.

c0bra:
Well that's kind of terrible, isn't it?

PR #4185, opened 2015-08-13
chore(build): Update release to support angular 1.4.x

JLLeitschuh:


JLLeitschuh:
Closes: #3925

mikemccaughan:
What versions of angular are supported by ui-grid? I can probably come up with a fix if I know that...

PR #4189, opened 2015-08-14
Fixing filter display when used with expandable/select column

gdigumarthy:


JLLeitschuh:
We don't support JQuery as a required dependency so you can't have jquery in core.
Also, what are you trying to fix. Do you have a specific issue?

gdigumarthy:
When we use filtering with selectable row or expandable row , it is not
enabling the filters.
The filters are hidden below the first row.
The root cause for this issue is the left pinned container header is not
sizing when the filters are enabled.
Somehow if I put some break points and delay the execution , the filters
appear.
I have tried http://ui-grid.info/docs/#/tutorial/401_AllFeatures , but no
luck.

PR #4191, opened 2015-08-14
Support function templates

mirik123:
ui-grid fails when template is defined as a function, for example:
````
$templateCache.put('ui-grid/uiGridCell', function (elem, attrs) {
 return "<div class=\"ui-grid-cell-contents\" title=\"TOOLTIP\">{{COL_FIELD CUSTOM_FILTERS}}</div>"
});
````

### DIFF
--- a/src/js/core/services/gridClassFactory.js
+++ b/src/js/core/services/gridClassFactory.js
@@ -112,6 +112,7 @@
              templateGetPromises.push(gridUtil.getTemplate(col[providedType])
                 .then(
                 function (template) {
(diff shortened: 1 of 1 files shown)


swalters:
can we move this logic to gridUtil.getTemplate(col[providedType])?  That way, any template anywhere in ui-grid can be a function.

mirik123:
yes, sure, no problem.
I have 7 more bugs in ui-grid-expandable on the road...

mirik123:
This pull request id a squashed copy of #4144

PR #4202, opened 2015-08-15
fix(columns): don't reset resized columns on data reset by default

### LINKED ISSUE #4005: Column width reset when updating grid.data
In Version 3.01, I have the following problem:

When you manually change column widths, and then reset grid.data, the column widths reset to the values defined in code.

I didn't have this problem in v3.0.0-rc.21.

See the following plunker: http://plnkr.co/edit/R2JWg0YO75RKBklwU9vN?p=preview

Change a column width, and then click on the "Set data a" button, and the column width resets to the default values.

What am i  missing?

EDIT:

The same issue can be seen in the tutorial 401: http://ui-grid.info/docs/#/tutorial/401_AllFeatures

Change a column width and click on "Refresh Data": the column width resets.

SomeKittens:
Previously, if a user resized a column and the data were reset, the column size would also be reset.  Instead, disable this behavior by default and add a flag (`allowCustomWidthOverride`) to allow it.  Code originally taken from https://github.com/mage-eag/ui-grid/commit/966c4d98bd7d0ab09e04936d564870cbeaa5c2a9

Fixes #4005

### DIFF
--- a/src/features/resize-columns/js/ui-grid-column-resizer.js
+++ b/src/features/resize-columns/js/ui-grid-column-resizer.js
@@ -430,6 +430,7 @@
 
           // check we're not outside the allowable bounds for this column
           col.width = constrainWidth(col, newWidth);
+          col.hasCustomWidth = true;
 
           refreshCanvas(xDiff);
 
@@ -542,6 +543,7 @@
 
           // check we're not outside the allowable bounds for this column
           col.width = constrainWidth(col, maxWidth);
+          col.hasCustomWidth = true;
 
           refreshCanvas(xDiff);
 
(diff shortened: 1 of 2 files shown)


swalters:
@c0bra @PaulL1 would you guys also take a look at this.  It looks good to me, other than we probably need to document this new flag.

cedejic:
Hi,

I may be wrong, but I believe you should also set column's hasCustomWidth property in uiGridSaveStateService::restoreColumns. So when you load state from e.g. database columns would set up properly.

```
              if ( grid.options.saveWidths ){
                currentCol.width = columnState.width;
                currentCol.hasCustomWidth = true;
              }
```

PR #4223, opened 2015-08-19
Fix misleading references to ng-grid and a typo

benheymink:
I've fixed up some old references to ng-grid in FIRST_TIMER.md. It was instructing people to do things like set ng-grid.git as the upstream project (which would have worked, but I think this is now less confusing). Oh, and fixed a typo too. Small change I know, but I wanted to get to grips with your build process.


JLLeitschuh:
Thank you very much!

PR #4225, opened 2015-08-19
fix(core): fix #4180 by adding validity check for minWidth and maxWidth

### LINKED ISSUE #4180: Columns not displaying properly
Hi,

when I am setting minWidth property for the columns they are not displayed properly.Below is the plunker to replicate this issue : 

http://plnkr.co/edit/OraYwr4YyWsDNcFOqRZ7?p=preview

Please look into this

Thanks,
Navneet

merlinchen:
- String and number are both allowed.

- Valid String value will parse into number properly.

- Invalid String value will throw an error.

### DIFF
--- a/src/js/core/factories/GridColumn.js
+++ b/src/js/core/factories/GridColumn.js
@@ -464,8 +464,23 @@ angular.module('ui.grid')
       }
     }
 
-    self.minWidth = !colDef.minWidth ? 30 : colDef.minWidth;
-    self.maxWidth = !colDef.maxWidth ? 9000 : colDef.maxWidth;
+    ['minWidth', 'maxWidth'].forEach(function (name) {
+      var minOrMaxWidth = colDef[name];
+      var parseErrorMsg = "Cannot parse column " + name + " '" + minOrMaxWidth + "' for column named '" + colDef.name + "'";
+
+      if (!angular.isString(minOrMaxWidth) && !angular.isNumber(minOrMaxWidth)) {
+        //Sets default minWidth and maxWidth values
+        self[name] = ((name === 'minWidth') ? 30 : 9000);
+      } else if (angular.isString(minOrMaxWidth)) {
+        if (minOrMaxWidth.match(/^(\d+)$/)) {
+          self[name] = parseInt(minOrMaxWidth.match(/^(\d+)$/)[1], 10);
+        } else {
+          throw new Error(parseErrorMsg);
+        }
+      } else {
+        self[name] = minOrMaxWidth;
+      }
+    });
 
     //use field if it is defined; name if it is not
     self.field = (colDef.field === undefined) ? colDef.name : colDef.field;
--- a/test/unit/core/factories/GridColumn.spec.js
+++ b/test/unit/core/factories/GridColumn.spec.js
@@ -560,6 +560,52 @@ describe('GridColumn factory', function () {
       expect(updateCol(colDef.width)).toThrow();
     });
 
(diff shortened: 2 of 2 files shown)


src/js/core/factories/GridColumn.js
```diff
@@ -464,8 +464,22 @@ angular.module('ui.grid')
...
-    self.minWidth = !colDef.minWidth ? 30 : colDef.minWidth;
-    self.maxWidth = !colDef.maxWidth ? 9000 : colDef.maxWidth;
+    ['minWidth', 'maxWidth'].forEach(function (name) {
+      var minOrMaxWidth = colDef[name];
+      var parseErrorMsg = "Cannot parse column " + name + " '" + minOrMaxWidth + "' for column named '" + colDef.name + "'";
+
+      if (!angular.isString(minOrMaxWidth) && !angular.isNumber(minOrMaxWidth)) {
+        self[name] = ((name === 'minWidth') ? 30 : 9000);
+      } else if (angular.isString(minOrMaxWidth)) {
+        if (minOrMaxWidth.match(/^(\d+)$/)) {
```
JLLeitschuh: Can you add a comment describing this regex?

src/js/core/factories/GridColumn.js
```diff
@@ -464,8 +464,22 @@ angular.module('ui.grid')
...
+          self[name] = parseInt(minOrMaxWidth.match(/^(\d+)$/)[1], 10);
```
JLLeitschuh: Can you add a comment describing this regex?

src/js/core/factories/GridColumn.js
```diff
@@ -464,8 +464,22 @@ angular.module('ui.grid')
       }
     }
 
-    self.minWidth = !colDef.minWidth ? 30 : colDef.minWidth;
-    self.maxWidth = !colDef.maxWidth ? 9000 : colDef.maxWidth;
+    ['minWidth', 'maxWidth'].forEach(function (name) {
```
JLLeitschuh: Should we also check the `width` variable too?

src/js/core/factories/GridColumn.js
```diff
@@ -464,8 +464,22 @@ angular.module('ui.grid')
...
+      var minOrMaxWidth = colDef[name];
+      var parseErrorMsg = "Cannot parse column " + name + " '" + minOrMaxWidth + "' for column named '" + colDef.name + "'";
+
+      if (!angular.isString(minOrMaxWidth) && !angular.isNumber(minOrMaxWidth)) {
+        self[name] = ((name === 'minWidth') ? 30 : 9000);
```
JLLeitschuh: Can you add a comment line describing this logic?
For example:
Sets default minwidth and maxwidth values

src/js/core/factories/GridColumn.js
merlinchen: The regex is a copy paste from the `width` variable checking logic.

src/js/core/factories/GridColumn.js
merlinchen: The `width` variable is already checked. 0_0
[https://github.com/merlinchen/ui-grid/blob/col-min-and-max-width-validity-check/src/js/core/factories/GridColumn.js#L435](https://github.com/merlinchen/ui-grid/blob/col-min-and-max-width-validity-check/src/js/core/factories/GridColumn.js#L435)

src/js/core/factories/GridColumn.js
merlinchen: Sure.

JLLeitschuh:
This looks good. Squash it into one commit and I'd have no problem merging it in.

merlinchen:
Squashed.

PR #4226, opened 2015-08-19
Partially fix jumpy infiniteScroll scrolling

reupen:
The current implementation causes the scroll position to jump around when new data is loaded and/or when data is removed from the grid. This is largely due to not accounting for the size of the viewport correctly when calculating new scroll percentages, but also because the getViewportHeight function does not appear to correctly account for the space taken up by filters if enabled and the horizontal scrollbar in the example below.

This pull request significantly reduces the jumping. The difference can be seen when scrolling in the two Plunkers below:

**Without changes:**
http://plnkr.co/edit/DfXkANDYkBOmjauF1HmF?p=preview

**With changes:**
http://plnkr.co/edit/pZwFdlK2iHWJqLEwDTrq?p=preview

This pull request does not fix the flickering that occurs when data is removed from the grid. I think fixing that that would need much larger changes. There also appear to be some other glitches remaining in Chrome.

I did not directly fix the getViewportHeight function (as regards the filter height problem) but just put some workarounds in the infiniteScroll code for now. I assumed changing getViewportHeight could potentially break other stuff, but obviously it's preferable to fix it.

### DIFF
--- a/src/features/infinite-scroll/js/infinite-scroll.js
+++ b/src/features/infinite-scroll/js/infinite-scroll.js
@@ -142,8 +142,8 @@
                * used if you're subsequently going to call `dataRemovedTop` or `dataRemovedBottom`
                */
               saveScrollPercentage: function() {
-                grid.infiniteScroll.prevScrolltopPercentage = grid.renderContainers.body.prevScrolltopPercentage;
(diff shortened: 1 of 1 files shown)


JLLeitschuh:
I had a look at your two demos and I can see that you did improve the overall scroll performance of infinite scroll.

Issue #4230, opened 2015-08-19
Exporting in IE only offers the Save and Cancel options


killahwave:
The exporting process uses the call msSaveBlob to create the csv file for export in > IE 9. However, this method suppresses the option to "open" the file when export is triggered. The user is left with Save and Cancel, and causing an extra click for the user. This will not work for my firm's heavy IE user base.

The issue can be demonstrated using IE 10 or greater with this plunker:
http://plnkr.co/edit/89ZVlPZcQbHYzgX5l4yq?p=preview

The fix is to change this method call to use msSaveOrOpenBlob. It seems like this would be a beneficial change to IE users. I'd be happy to submit this as a pull request.

PaulL1:
Please go ahead and submit the PR, I'm happy with that.

My main problem is that I don't have IE at all to test with.  I'd love it if you could test IE10 and 11, for both CSV export and PDF export, and fix or discuss anything broken.  Unfortunately we typically get reports of things that don't work, but it's hard to find out things that do work.  So it's currently not clear whether or not PDF export works on IE.

killahwave:
This change affects the CSV export, only. It has been tested on IE 10 and 11, and works as expected.

We aren't currently implementing the PDF export, but will try to test it out when we have some time.

PaulL1:
Thanks for the PR, would love to hear how pdf export is working for you.

AgDude:
Looks like this should have been closed by the reference PR.

PR #4235, opened 2015-08-20
Fix bug in IE 10+ so that file export has options to save, cancel and…

killahwave:
#4230 

Current functionality only offers save and cancel. This change allows open, save and cancel to appear as options when exporting in IE 10+.


PaulL1:
Thanks, looks good.

Issue #4066, opened 2015-07-28
Incorrect css link in the example


Leo-g:
I was trying the tutorial http://ui-grid.info/docs/#/tutorial/101_intro and I found the link to the css style sheet is not correct in the example
     
    <script src="/release/ui-grid.css"></script> 

Shouldn't it be ?
     
      <link rel="styleSheet" href="node_modules/angular-ui-grid/ui-grid.min.css"/>

Also the main.css file was not included in index.html

Leo-g:
Also on the main site  http://ui-grid.info/ when I click on the tutorial button I get redirected to http://127.0.0.1:9003/docs/#/tutorial

JLLeitschuh:
That is a known bug that should be fixed.

ingshtrom:
1. It looks like this is prevalent across a lot of the tutorials. So a PR should fix all of those.
2. the link to the tutorials on the homepage links to http://ui-grid.info/docs/#/tutorial as expected. That must have been fixed in another PR.
3. I want to get started with UI-Grid and contributing (even the basics of PR's in GitHub), so I'll take this.

ingshtrom:
One note: PR #4245 does not fix the css being loaded via script tags in all of the tutorial's source code.  I think I would need to create a PR on https://github.com/c0bra/grunt-ngdocs to fix that.

c0bra:
`<script src="/release/ui-grid.css"></script>` is correct, it's a relative url to http://ui-grid.info/release/ui-grid.css

PR #4245, opened 2015-08-23
#4066 - Fix path to ui-grid.min.css in tutorial/101_intro

ingshtrom:
- Issue #4066 
- this is my first PR on GitHub. ever. Let me know of any incorrect styles or procedures I did not complete. 
- this does NOT fix the script tags that point to css files in all of the tutorial’s source code--that may need to be verified by @c0bra 

Thanks!

### DIFF
--- a/misc/tutorial/101_intro.ngdoc
+++ b/misc/tutorial/101_intro.ngdoc
@@ -23,7 +23,7 @@ Steps:
 <ul>
  <li>Include uiGrid in your script and css
   <pre>
-      <link rel="styleSheet" href="release/ui-grid.min.css"/>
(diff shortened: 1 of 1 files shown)


swalters:
@c0bra can you verify that this is an ok change?

c0bra:
`node_modules` doesn't get published anywhere, so no I don't think this is correct.

c0bra:
`<script src="/release/ui-grid.css"></script>` is correct, it's a relative url to http://ui-grid.info/release/ui-grid.css

ingshtrom:
@c0bra Alright, I guess I misread that issue then. I was thinking we were documenting it as if the user brought in UIGrid using NPM, since this isn't actually loading the script for the tutorial page.  So then #4066 should be closed since there isn't an issue there?

Issue #3992, opened 2015-07-16
Combine menu and sort icon in header


AgDude:
I am opening this ticket for some feedback, The problem is for narrow columns the column menu icon and the sort icon take up lots of space, and often the menu icon ends up overlapping the display name.

One idea is to combine the two icons. Here is a proof of concept (no grid, just icons):

http://plnkr.co/U3t2mJ9Khok3p4iW5eNL

The basic idea comes from Excel when "auto-filter" is enabled.

Clicking on this icon would open the menu, which would have sort options, clicking the header text could also apply sorting.

PaulL1:
Works OK for me - it's basically still showing both icons, just a bit smaller.  The semantics, as you note, work fine by clicking on the header, so there's no real need for the sort icon to be separately clickable.

lebolo:
:+1:

PR #4257, opened 2015-08-25
feat(core): Adds sort priority number to header

JLLeitschuh:
The headers now show the sort priority of the column.
This is done using a subscript tag that is pulled closer to the sort
direction icon.
New Sort Icons

### DIFF
--- a/src/less/header.less
+++ b/src/less/header.less
@@ -79,6 +79,11 @@
   .sortable {
     cursor: pointer;
   }
+
+  // Moves the sort priority number closer to the icon
+  .ui-grid-sort-priority-number {
+    margin-left: -8px;
+  }
 }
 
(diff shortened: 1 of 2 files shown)


swalters:
I like this a lot.  @c0bra if you're cool with it, please merge.  (I seem to remember you having some comments about this feature in 2.x)

AgDude:
@JLLeitschuh I like the numbers.  

I started to put the sort icon and the column menu together, but never found the time to get it ready for a PR. There is a plunker linked in #3992. Would you be up for combing that into your feature? I have implemented it in my app, but it is a simplified because we always have a column menu.

JLLeitschuh:
@AgDude I think thats outside the scope of this PR and outside of the time constraints that I have at the moment.

PR #4263, opened 2015-08-25
fix(cellNav): Error when field defined with number

### LINKED ISSUE #4258: Error is thrown when focusing a cell
We have identified that the error below is thrown when using dot notation in col.field e.g "1.Blabla". This occurs once you focus the cell.
screen shot 2015-08-25 at 09 39 44

The function below is what causing this error and was added in 9532de2bdb083a0bb40a6099924e98893ab354e7
```javascript 
RowCol.prototype.getIntersectionValueRaw = function(){
        var getter = $parse(this.col.field);
        var context = this.row.entity;
       return getter(context);
      };
```
9532de2bdb083a0bb40a6099924e98893ab354e7

JLLeitschuh:
Cell nav was throwing an error when you had a field defined like `'1.b'`
Resolved by using the `getQualifiedColField` method on the gridRow.

Closes #4258

### DIFF
--- a/src/features/cellnav/js/cellnav.js
+++ b/src/features/cellnav/js/cellnav.js
@@ -67,8 +67,8 @@
        *          If the column has a cellFilter this will NOT return the filtered value.
        */
       RowCol.prototype.getIntersectionValueRaw = function(){
-        var getter = $parse(this.col.field);
-        var context = this.row.entity;
+        var getter = $parse(this.row.getEntityQualifiedColField(this.col));
+        var context = this.row;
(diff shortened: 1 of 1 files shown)


swalters:
the change looks good.

One question I have is why is this code in cellNav?  it doesn't look like a cellNav specific logic.  Should it be moved to core?

JLLeitschuh:
Sure? I suppose. I put it here because that's where it was originally defined before I made it an angular module. I can move it to core in another PR.

PR #4273, opened 2015-08-26
Fix row width being shorter than header

ayatsai:
example:
when the grid's total width is 150px, and we set column 1 to 25% and column 2 to 75%, column 1 will get rounded to 37px and column 2 will get rounded to 112px, leaving a 1px difference between the rows and the header. after fix, the column 1 will be 37.5px, column 2 will be 112.5px


swalters:
@PaulL1 can you review?  you had this comment in that file:
- we use parseInt quite a bit, as we try to make all our column widths integers

imbalind:
@PaulL1 do you know if there is a problem in having pixels not rounded?

PR #4290, opened 2015-08-30
perform simple shallow watch of data

david2tm:
when using same array for data when 1) replacing array items 2) number of items is the same as before, grid will fail to update. So using $watchCollection will also perform simple check of first level of inner members.

is in following example:

``` javascript
$scope.data = [{id:"foo1"}, {id:"foo2"}, {id:"foo3"}];

// after change like bellow, grid will not refresh data
$scope.data.length = 0;
$scope.data.push({id:"foo4"});
$scope.data.push({id:"foo5"});
$scope.data.push({id:"foo6"});
// as it watches fore reference (unchanged) and .length (unchanged too)
```

### DIFF
--- a/src/directives/ng-grid.js
+++ b/src/directives/ng-grid.js
@@ -95,7 +95,7 @@
                                 }
                                 $scope.$emit("ngGridEventData", grid.gridId);
                             };
-                            $scope.$on('$destroy', $scope.$parent.$watch(options.data, dataWatcher));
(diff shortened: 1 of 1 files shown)


mportuga:
@JLLeitschuh Do you still think this is a good change?

JLLeitschuh:
We haven't been maintaining the 2.x branch.

mportuga:
We haven't been maintaining the 2.x branch.

Issue #4242, opened 2015-08-21
Enable a default sort direction in columnDefs 


jonathanboswell:
Looks like this feature existed in 2.0
https://github.com/angular-ui/ui-grid/pull/1864

I've gone through the javascript and maybe I'm missing it, but I can't find anyway of doing this.

Use case:
Financials almost always make sense to have the first click be DESC, we do a lot of sorting on various currency fields and currently have to click twice to get it to DESC

Thank you for all the great work!

PaulL1:
http://ui-grid.info/docs/#/tutorial/102_sorting or http://ui-grid.info/docs/#/api/ui.grid.class:GridOptions.columnDef

You set the direction on the columnDef, in the sorting object.

jonathanboswell:
Correct me if I'm wrong, but that sets the sort on the load, not the first click on a column header.

For example, let's say we have a grid with two columns "Clients" and "Revenue" on page load we want it to load clients alphabetically ascending. Like so:

$scope.gridOptions = {
  enableSorting: true,
  data: clientFactory.revenues
  columnDefs: [
            { field: 'clients', sort: { direction: 'asc', priority: 0 } },
            { field: 'revenue', cellFilter: 'currency' }
  ]
};

All well and good. The issue is that whenever we want to sort on "Revenue" by clicking on the header it will sort 'asc' with all the zero revenue clients first when really what we wanna see is all the highest revenue clients first. Sure we can click on the header again, but in a financial grid with 7 currency fields constantly having to double click the columns is not ideal. And if we add the sort object to the revenue field:
  columnDefs: [
            { field: 'clients', sort: { direction: 'asc', priority: 0 } },
            { field: 'revenue', cellFilter: 'currency', sort: { direction: 'desc' } }
  ]

...the page loads with all the clients sorted alphabetically and then revenue sorted with any clients that have the same name, which is not the solution.

This functionality appears to be in 2.x: https://github.com/angular-ui/ui-grid/pull/1864

And seems to be missing from 3.x

Again, hopefully I'm wrong about this, but I've thoroughly gone through the documentation and code and it seems to not exist anymore.

PaulL1:
Ah, OK, I see.  This you can do by listening to the sort changed event, and altering the sort manually.  A bit clunky, but it will work.

benheymink:
[...]

PR #4338, opened 2015-09-08
add preferredSortDirection column option

gmaslov:
This implements the "default sort direction" feature requested in #4242. The columnDef now accepts a new option called "preferredSortDirection" which controls the initial sorting direction when clicking on an unsorted column. Previously this would always be uiGridConstants.ASC.

Please let me know if there's anything I could improve about this PR :-)

### DIFF
--- a/src/js/core/factories/Grid.js
+++ b/src/js/core/factories/Grid.js
@@ -1921,7 +1921,8 @@ angular.module('ui.grid')
    * Emits the sortChanged event whenever the sort criteria are changed.
    * @param {GridColumn} column Column to set the sorting on
    * @param {uiGridConstants.ASC|uiGridConstants.DESC} [direction] Direction to sort by, either descending or ascending.
-   *   If not provided, the column will iterate through the sort directions: ascending, descending, unsorted.
+   *   If not provided, the column will iterate through the sort directions: preferred, not-preferred, unsorted.
+   *   The default preferredSortDirection is ascending.
    * @param {boolean} [add] Add this column to the sorting. If not provided or set to `false`, the Grid will reset any existing sorting and sort
    *   by this column only
    * @returns {Promise} A resolved promise that supplies the column.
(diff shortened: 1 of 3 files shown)


JLLeitschuh:
If you want to change your commit message to follow the following format it will automatically be added to the changelog when we do the next release:
https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#-git-commit-guidelines

gmaslov:
OK, done. Thanks!

swalters:
@gmaslov great PR.  

Another complaint that we have had is that we have a third state - unsorted.  Do you think your approach could be extended to allow developers to eliminate the unsorted state?

maybe something like colDef.Sorts = "DESC,ASC".  The default would be "ASC,DESC,UNSORTED".  If you think you could get something working like this that would also work for your scenario, it would be great.   

If not, I'll merge what you have and we'll address the other later.

gmaslov:
I believe the existing suppressRemoveSort option covers that use case, unless I misunderstand. I like your suggestion though. I think it would probably simplify things to give a list like that instead of a couple of flags. If you don't mind waiting a week or two I could submit another PR taking that approach, replacing this new preferredSortDirection option and deprecating the existing suppressRemoveSort flag. On the other hand maybe it'd be better to keep this as it is to avoid having to add backward compatibility logic. Let me know what you think.

benheymink:
@gmaslov @swalters I think the issue with suppressRemoveSort is that it prevents sorting being removed from a column when you subsequently sort on another column. We have a use case to only allow sorting on one column at a time - if a column is currently being sorted, but the user clicks on another column, then sorting should be removed on the first column. However, we don't want the ability to remove sorting on a column manually, if that makes sense?

swalters:
@gmaslov If you are willing and have the time, I'd like to keep this PR open and get the more comprehensive approach.

gmaslov:
@swalters OK, I'll look into it this week and let you know. Thanks.

swalters:
merged #4366 which uses different solution

PR #4343, opened 2015-09-09
Dynamically expanded row height

Namek:
I have rebased+squashed PR #3201 of @yccteam.


Namek:
I have removed event part of commit from `expandable.js` because it didn't pass tests:
```js
toggleRowExpansion: function (rowEntity) {
   event.stopPropagation();   // <-- these two lines
   event.preventDefault();    //
```

Issue #4363, opened 2015-09-14
infiniteScroll: no more data is loaded if initial load didn't bring enough data to make scroll bar show up


DmitryEfimenko:
Please [see Plunkr](http://plnkr.co/edit/O8ea7ja659ogov34pAfV?p=preview) that shows that issue.

JLLeitschuh:
I have no idea how we would work around this though.
Do you have an idea?

PR #4365, opened 2015-09-14
fix(infinite-scroll): load more data if needed

DmitryEfimenko:
detect if more data needs to be loaded after the initial load. Before if
not enough data was loaded to enable scrolling, no more data would be
loaded. Resolves issue [4363](https://github.com/angular-ui/ui-grid/issues/4363)

### DIFF
--- a/src/features/infinite-scroll/js/infinite-scroll.js
+++ b/src/features/infinite-scroll/js/infinite-scroll.js
@@ -368,6 +368,12 @@
 
           newVisibleRows = grid.getVisibleRowCount();
 
+          // in case not enough data is loaded to enable scroller - load more data
+          var canvasHeight = rowHeight * newVisibleRows;
+          if (grid.infiniteScroll.scrollDown && (viewportHeight > canvasHeight)) {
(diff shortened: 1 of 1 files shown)


src/features/infinite-scroll/js/infinite-scroll.js
```diff
@@ -368,6 +368,11 @@
 
           newVisibleRows = grid.getVisibleRowCount();
 
+          var canvasHeight = rowHeight * newVisibleRows;
+          if (grid.infiniteScroll.scrollDown && (viewportHeight > canvasHeight)) {
+            grid.api.infiniteScroll.raise.needLoadMoreData();
+          }
+
```
JLLeitschuh: Can you add a comment explicitly describing what is going on here and why. Without the context of the commit message future developers will have no idea why this chunk of code is here.

JLLeitschuh:
This needs to be squashed into one commit before it will be merged.

DmitryEfimenko:
took me a while, but finally squashed

JLLeitschuh:
@angular-ui/grid This is ready to go if someone wants to merge it.

PR #4366, opened 2015-09-14
feat(core): add sortDirectionCycle column option

gmaslov:
Alternative to #4338, addressing #4242 in a different way.

This option allows you to specify how a column's sort direction should
be cycled as the heading is repeatedly clicked. The default value,
[null,ASC,DESC], preserves the existing behavior.

I decided to disallow repeating values like [null,ASC,ASC,ASC,DESC] since that would require keeping track of the index explicitly, and I didn't want to add any more fiddly internal state variables.

@benheymink I think that this logic should handle your use case, if I understand correctly. Setting the cycle to [ASC,DESC] on all columns should result in sorting by one column at a time (the most recently clicked one), but the sort won't be removed by repeated clicks on the same column.


benheymink:
I'd still prefer an option to simply disable removing sorting on a column (that works when you sort on another column) - but this looks like it would fix my case. Would also allow us to change the default sort direction on a column, which is a bonus.

swalters:
Looks great! Thanks for the PR.

Issue #4411, opened 2015-09-21
Keybind Up de-editing issue when using Edit WIth Cell Nav (scrollToIfNecessary bug)


dpinho17:
Hi,
After trying to create our grid using the Edit With Cell Nav feature which is exactly what we need. We noticed that if you move up to something that is new in the viewport it does not correctly scroll which causes that Cell to not be in Edit Mode. This seemed to work fine for going below the bottom bound however going over the top bound did not work. After looking at the UI-Grid code I found the issue.

Reproductions Steps:
http://ui-grid.info/docs/#/tutorial/309_editable_with_cellnav

If you scroll down to bottom in the demo app and then select a cell it will go to edit mode. Now continue to press the up key until you go past the top bound, in the old non fixed code it would incorrectly take you out of edit mode because of the scroll issue.

My Solution to the Problem:
The issue lies within the scrollToIfNecessary method. The calculation for scrolling up is not exactly correct. There were a few minor issues making the scrolling not consistent and one major issue which caused the up scrolling not to work. 

The main issue was that the variable pixelsToSeeRow was calculated assuming you always needed an extra row. Here is the line of code.

var pixelsToSeeRow = ((seekRowIndex + 1) * self.options.rowHeight);

As you can see it adds one row to the seekRowIndex. This works perfectly fine when going down as you always move to the next row however for going up this should not include the +1 and should simply be:

var pixelsToSeeRow = (seekRowIndex  * self.options.rowHeight);

Doing this fixes the major issue however there was also a few small calculation errors in the scrollLength and topBound variables.

For the scrollLength it is missing the scrollBarHeight which is included in other calculations for scrollLength throughout the code. Here is the before and after:

Code before scrollLength fix:
var scrollLength = (self.renderContainers.body.getCanvasHeight() - self.renderContainers.body.getViewportHeight());

Code after scrollLength fix:
var scrollLength = (self.renderContainers.body.getCanvasHeight() - self.renderContainers.body.getViewportHeight() + self.renderContainers.body.grid.scrollbarHeight);

This fix allows the scrollToIfNecessary method to more accurately scroll to where it needs to scroll.

The last change was in TopBound. In the code the TopBound calculation includes HeaderHeight however from my testing it doesn't seem like headerheight should be included as this throws off the topBound which makes the scrollToIfNecessary scroll prematurely. Here is the before and after:

Code before topBound fix:
var topBound = self.renderContainers.body.prevScrollTop + self.headerHeight;

[...]

PR #4412, opened 2015-09-21
Fix (scrollToIfNecessary) - Moving Up Past Top Bound Issues - #4411

dpinho17:
Issue #4411 - Navigating up past the Top Bound of the grid would sometimes not put the cell into edit mode when using Edit on Cell Nav Feature. This was due to scrollToIfNecessary having a faulty calculation on where to scroll to.


dpinho17:
I'm not sure why this is failing, this is my first pull request so I may have done something wrong. Please le me know if you need me to do something to fix this.

Issue #4386, opened 2015-09-17
buildColumns / preCompileCellTemplates called each time grid data is changed


oriondean:
Hi,

I noticed that ui-grid dataWatchFunction rebuilds all columns and pre-compiles cell templates each time data is changed (see: [ui-grid.js:109](https://github.com/angular-ui/ui-grid/blob/caf8ae5e89c2e13dc242db23c9f2fc3435fefc4b/src/js/core/directives/ui-grid.js#L109)).

My understanding is that after the columns/cell templates have been built/compiled once then they do not need to be built/compiled again unless the column definitions change in any way.
There's already a watcher on the column definitions (see: [columnDefsWatchFunction](https://github.com/angular-ui/ui-grid/blob/caf8ae5e89c2e13dc242db23c9f2fc3435fefc4b/src/js/core/directives/ui-grid.js#L68)) that does this job already.

From my experiences, not building/compiling columns/cell-template each time data changes can improve performance significantly for grids that update data often.

Question: Is there any possible reason why columns need to be built each time data changes?
If not, i'm more than happy to submit a pull request to add this improvement!

Thanks

badri77:
Hi, in converse to this issue, i am binding variable length data, so i am binding columns dynamically and also i have conditional styles and icons are being rendered on based the cell data. When, i rebind the data (which is different), the data is changing but not cell styles and icons. please let me know how to resolve this issue.

PR #4428, opened 2015-09-23
perf(core): Don't rebuild columns and precompile cell templates each time data is changed

oriondean:
Resolves https://github.com/angular-ui/ui-grid/issues/4386

This is quite a nice performance improvement in grids that have rapidly updating data

### DIFF
--- a/src/js/core/directives/ui-grid.js
+++ b/src/js/core/directives/ui-grid.js
@@ -91,9 +91,12 @@
         }
         
         if (newData) {
+          // columns length is greater than the number of row header columns, which don't count because they're created automatically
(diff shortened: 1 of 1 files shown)


JLLeitschuh:
Thank you for your pr. I'm going to defer this to the rest of the @angular-ui/grid to review. But I'll add my looks good to me.
If you change your commit message to match these guidelines then it will show up in our changelog.
https://github.com/angular-ui/ui-grid/blob/master/CONTRIBUTING.md#-git-commit-guidelines
If you do choose to change your commit message simply amend your commit locally and then do `git push [branch name] --force`

oriondean:
I've updated the commit message to match the guidelines, let me know if it's still incompatible. 

It makes good sense to defer this to the rest of the team, I'm still unsure whether anything relies on the column set being rebuilt each time the data changes!

PR #4455, opened 2015-10-01
feat(sort): Give more information to the sort functions

Rhathe:
Passing rowA and rowB to the sort functions allows the sorting functions to consider other fields when the sorting occurs. Passing direction is useful for such cases such as always having null values sorted at the bottom of the grid.


swalters:
Great idea.  Can you also update the documentation on the sort function?  it should be an embedded JSDoc somewhere for the sortFn.  Also, check if any of the tutorials can be update with the info. Don't forget to squash your commits into a single commit. Thanks.

Rhathe:
I've updated the pull request with the extra documentation

chmirenko:
@Rhathe, @swalters 
How can you get access the sorted column from inside the custom algorithm?
I posted on stack overflow and gitter:
http://stackoverflow.com/questions/39649712/ui-grid-get-the-current-column-field-name-from-inside-a-custom-sorting-algorit

PR #4458, opened 2015-10-01
Clean up format of modifyRows function

nkoterba:
Hopefully this displays the pseudo-code as intended.  Not sure if Markdown code block start and end syntax is supported in the help guide renderer.


JLLeitschuh:
Thanks!

PR #4461, opened 2015-10-02
fix(selection): properly update selectAll flag based on current selection

ivanvoznyakovsky:
This fixes the following case:
* select all rows w/ "Select all" button
* deselect one of the selected rows
* "Select all" becomes unchecked
* select back unselected row

Problem: "Select all" expected to become checked again as all rows are selected but it's not.

### DIFF
--- a/src/features/selection/js/selection.js
+++ b/src/features/selection/js/selection.js
@@ -464,10 +464,11 @@
             return;
           }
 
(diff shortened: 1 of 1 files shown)


JLLeitschuh:
This makes sense.
Thanks for the fix.

PR #4519, opened 2015-10-14
fix(infinite-scroll): Remove returns of adjustInfiniteScrollPosition.

nkoterba:
Update uiGridInfiniteScrollService methods so they do **not** return promise of `adjustInfiniteScrollPosition` when `adjustInfiniteScrollPosition` does not itself return a promise.  Also updated documentation to match code changes.

Seems to be discrepancies between scroll events returning promises and not returning promises.  I think it's not a bad idea to return promises but `adjustInfiniteScrollPosition` calls `grid.scrollContainers` which doesn't return a promise so the promise chain doesn't actually exist.

### DIFF
--- a/src/features/infinite-scroll/js/infinite-scroll.js
+++ b/src/features/infinite-scroll/js/infinite-scroll.js
@@ -125,12 +125,11 @@
                * infinite scroll events upward
                * @param {boolean} scrollDown flag that there are pages downwards, so
                * fire infinite scroll events downward
-               * @returns {promise} promise that is resolved when the scroll reset is complete
                */
               resetScroll: function( scrollUp, scrollDown ) {
                 service.setScrollDirections( grid, scrollUp, scrollDown);
 
-                return service.adjustInfiniteScrollPosition(grid, 0);
+                service.adjustInfiniteScrollPosition(grid, 0);
               },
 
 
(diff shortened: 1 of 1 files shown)


JLLeitschuh:
Would it be better to create a PR that actually adds the promise?

nkoterba:
@JLLeitschuh So initially, I went in and added `return` statements and updated the documentation to use/return promises only to realize that everything comes down to `Grid.scrollContainers`: https://github.com/angular-ui/ui-grid/blob/e89598a588791a6fc18a040c00a805dc1b6c5d25/src/js/core/factories/Grid.js#L1717.

`scrollContainers` does **not** use promise as it calls each scroll callback handler synchronously (I assumed this based based on how the callback handlers are named, i.e., with Sync in their name).

The only `syncScrollHandler` currently added by ui-grid source is a vertical scroll handler: https://github.com/angular-ui/ui-grid/blob/9d1a0b46e54983e09e0be8f89f9fac0ee74046e4/src/js/core/directives/ui-grid-viewport.jsL81 which does nothing more than adjust scrollTop on the elment.

I don't think any of those runs asynchronously OR provides promises so I'm not sure there's any promise that could actually be bubbled up.

I was planning to submit a PR to allow custom scroll timing and animation as this is a feature I would like. So instead of:
```javascript
$elm[0].scrollTop = newScrollTop;
```
it would instead look like:

```javascript
$elm[0].animate({
					scrollTop: newScrollTop
				}, 400);
```

The `animate` method has a callback or a promise I believe so I could indeed bubble that up.  However, the `grid.scrollContainers` would need to be re-written to **not** be synchronous and/or any Scroll Handlers added would need to be "async" based and then we'd probably need to use `$q.all()` instead of looping through each SyncScrollHandler in `scrollContainers`.

I submitted this PR partially to start a discussion on this topic and also perhaps to get some feedback from whomever wrote `scrollContainers` as to the async vs. sync nature of Scroll events and function calls.

In the meantime, however, cleaning up the documentation and code base so its consistent seemed like it could benefit others.

mportuga:
@dlgski I am not sure what to do about this one, seems harmless enough though.

nkoterba:
@mportuga I did this a while back but if you look at the code/documentation "promises" are being returned when in reality there are no actual promises or the documentation is not aligned with the actual internal code.

PR #4575, opened 2015-10-26
Remove code triggering scroll on each menu select

tirithen:
Each time toggling column visibility with the top right menu the menu is scrolled to the top. This gets impossibly annoying when having 20+ menu items. Once this code (that is mentioned as a "quick fix") is removed it stops scrolling the menu to the top when showing/hiding a column.

### DIFF
--- a/src/js/core/directives/ui-grid-menu.js
+++ b/src/js/core/directives/ui-grid-menu.js
@@ -247,13 +247,6 @@ function ($compile, $timeout, $window, $document, gridUtil, uiGridConstants, i18
 
               if ( !$scope.leaveOpen ){
                 $scope.$emit('hide-menu');
-              } else {
-                /*
-                 * XXX: Fix after column refactor
-                 * Ideally the focus would remain on the item.
-                 * However, since there are two menu items that have their 'show' property toggled instead. This is a quick fix.
-                 */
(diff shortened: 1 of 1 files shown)


JLLeitschuh:
This is my fault.
However, this isn't really a good solution because it breaks the focus management.
A better solution would be to find a CSS selector to move focus to.
A easy fix would be to add a grid option to disable this.

tirithen:
Sorry I'm new to this project so I do not have a complete overview of what that code should do, is the point visually marking the button triggered as focused?

Should we not use something like this in that case?

    gridUtil.focus.bySelector(angular.element($event.target.parentElement), 'button[type=button]', true);

Take the parent li of $event.target (that is the "clicked" button tag) and pass on this button selector to bySelector method, or maybe just directly do:

    $event.target.focus();

I did try the two methods but the focus does not trigger any visual styling changes, is there a different effect that the "quick fix" was supposed to trigger?

tirithen:
I'm very keen on having this issue solved and included in a ui-grid release as soon as possible to prevent us from having a modified copy of this script in the project that we are currently using ui-grid in.

JLLeitschuh:
That looks like it could solve the issue as well. Give it a shot and tell me if it fixes both of our problems.

mportuga:
@JLLeitschuh @tirithen  What is the status of this?

mportuga:
@JLLeitschuh ?

mportuga:
Closing due to lack of response.

PR #4577, opened 2015-10-26
feat(edit): add a function to retrieve dropdown options

imbalind:
Add a new member to ColumnDef: editDropdownOptionsFunction:
a function returning an array of values in the format
[ {id: xxx, value: xxx} ], which will be used to populate
the edit dropdown.  This can be used when the
dropdown values are dependent on the backing row entity with
some kind of algorithm.
If this property is set then both editDropdownOptionsArray
and editDropdownRowEntityOptionsArrayPath will be ignored.


swalters:
Great PR! thanks

PR #4586, opened 2015-10-28
fix(grid): adjust grid height when initial height is equal to row height

ivanvoznyakovsky:
Today had a situation when grid did not adjust it height to minimum rows to show height. It turned out that rendered grid height was equal to `rowHeight` option value and so `autoAdjustHeight` was not called.
This [change](https://github.com/angular-ui/ui-grid/compare/master...ivanvoznyakovsky:fix-autoadjust-heigth?expand=1#diff-d586707ee1c4e842337e09afc6ebd587R280) fixes this.


ivanvoznyakovsky:
`npm install` failed...

Issue #4152, opened 2015-08-10
Validation proposal


imbalind:
# Ui-grid validation

I've started working on some kind of validation.

You can find a working [proof of concept](http://plnkr.co/edit/Kpc3Ul?p=info) of what I've come up with and a short analysis of how I thought it should be.

Comments & suggestions are needed and welcome.
## Guidelines
- Validation should be an external feature, 
  just like navigation, grouping, importing, etc.
- Validation can safely rely on ui-grid.edit since validation would make little 
  sense without editing capabilities.
- The end user should be able to insert a not valid value, and see a significant
  graphical hint that warned him about the invalidity of the inserted value (for
  example a red border). 
- Given how users could interact with a grid (i.e. if doing data entry they could
  be editing many fields without even watching the screen), validation hints 
  should be kept until the field is edited againg with a valid value.
- Validation could not rely on standard angular validation because of the way
  ui-grid and angular ng-model works (please, correct me if I'm wrong):
  - If we created a standard angular validation on a text input field, and the 
    user inserted an invalid value, then the model would not update to the invalid
    value but would keep the original value.
  - The user would thus see the field showing the old value but without being
    warned about the invalid value.
  - The invalid value would be lost possibly causing loss of data.
- Validation would be based on a set of standard validator wich could be 
  extended programmatically.
- Both standard and custom validation should be asyncronous.
## How to add validation
1. Validation is added to the grid using `ui-grid-validate` directive and to the
   application injecting `ui.grid.validate`.
2. Validators are defined:
   - A standard validator can be attached to a column declaring it as part of 
     `columnDefs` inside the `gridOptions`
   - A custom validator can be created and added to the standard ones for a
     given column by calling the API method `addColumnValidator`.
3. To show the hint a custom template is needed. There are two readily available:
   - `ui-grid/cellTitleValidator`
   - `ui-grid/cellTooltipValidator`  
   
   Both of them show a red border based on a `ng-class` attribute whose job is 
   to add the `invalid` class to invalid fields.  
   Furthermore they can show a list of error messages but they do so using
   different methods: the former uses the  `title` attribute of the cell `div`, 
   while the latter depends on ui-bootstrap and uses a `tooltip`.
## How it works
1. Upon creation of the `ui-grid-validate` directive, the `initializeGrid`
   method is called wich:
   1. catches the `afterCellEdit` event from `ui.grid.edit`'s API and binds it to
      the `runValidators` internal method.
[...]

PR #4635, opened 2015-11-03
feature(validation): add validation feature

imbalind:
This feature will make validation on fields possible.
It will not use angularjs basic validation because it wouldn't work when editing is over.

Instead it adds custom properties to field considered as invalid.

It provides a basic way to display error messages using html title attribute and a cooler one that depends on bootstrap to show a popover.

Tests and docs are provided.

For more info see issue #4152


swalters:
Merging per a team discussion that agree's that we should have as part of central repository instead of a separate feature repository in 3.0

PR #4636, opened 2015-11-03
Romanian language

sMoKeSz:
Added romanian language translation .

### DIFF
--- a/src/js/i18n/ro.js
+++ b/src/js/i18n/ro.js
@@ -1,0 +1,106 @@
+(function () {
+  angular.module('ui.grid').config(['$provide', function($provide) {
+    $provide.decorator('i18nService', ['$delegate', function($delegate) {
+      $delegate.add('ro', {
+        headerCell: {
+          aria: {
+            defaultFilterLabel: 'Filtru pentru coloana',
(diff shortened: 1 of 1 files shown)


sMoKeSz:
so what should i do now ?

PaulL1:
You need to correct it so that it passes all tests, and then squash those changes so that it's a single commit.

sMoKeSz:
that's not my error i only add a new language .. maybe i should for another branch and not master? what's the newest branch?

PaulL1:
OK, I've restarted the build - it looks like it had an intermittent error on fontello.

PaulL1:
Hmm.  I'm assuming that the fontello error exists on master, and merging anyway.

sMoKeSz:
thank you  for merge, whats the next step to get this commit on angular-ui bower install, cause i need that very much :) (we need that )

PaulL1:
It won’t come through to bower until we cut a new version.  I believe these are going every few weeks, and it will just come through when that happens.

Paul Lambert
Program Director NTS
021 385 488

sMoKeSz:
hello how about next version of ui-grid ?

PR #4651, opened 2015-11-05
remove header sorts for columns with columnDef sortable false

mikeallisonJS:
simple template change, removes header sort for sortable: false.


c0bra:
@mikeallisonJS what are you trying to accomplish here? Just hide the sorting icon if the column isn't sortable? If so, why not toggle the visibility of the sorting icon `<span>` rather than adding another binding location for the column name?

Note that ng-ifs affect grid performance because it causes DOM elements to create/destroyed as the grid is scrolled.

PR #4663, opened 2015-11-09
feat (i18n): Polish translation.

przemoli:
Polish translation based off english translation.

### DIFF
--- a/src/js/i18n/pl.js
+++ b/src/js/i18n/pl.js
@@ -1,0 +1,109 @@
+/**
+ * Created by zuk38 & przemoli on 9/11/15.
+ */
+(function () {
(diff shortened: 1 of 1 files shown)


przemoli:
Most of actuall translation should be credited to zuk38 in #1951

PaulL1:
Awesome.

przemoli:
When can I expect release with this translation? Can You release it as 3.0.8? Since translations can not be added easily to min.js build of ui-grid, it's somehow pressing topic.

PaulL1:
The releases tend to go every few weeks, and it will be automatically in that release.  I'm not sure at the moment what the plan is for 3.0.8 timing.

PR #4698, opened 2015-11-15
fix(edit): emit blur event on stopEdit

aiguofer:
Using a custom template with `ng-model-options={ updateOn: 'blur' }` is not working. #4058 attempted to fix it, but the problem is that with the custom handling of key events, the blur event never gets emitted for the edit cell using TAB or ENTER for example. This seems to fix it and I don't see why it would have adverse effects (other than potentially emitting 'blur' more than once).

### DIFF
--- a/src/features/edit/js/gridEdit.js
+++ b/src/features/edit/js/gridEdit.js
@@ -974,6 +974,7 @@
                 $scope.deepEdit = false;
 
                 $scope.stopEdit = function (evt) {
+                  evt.target.blur();
                   if ($scope.inputForm && !$scope.inputForm.$valid) {
                     evt.stopPropagation();
                     $scope.$emit(uiGridEditConstants.events.CANCEL_CELL_EDIT);


imbalind:
I found an issue in your code, you can replicate following these steps:

- Open tutorial 301
- Start editing fields and passing to the next one pressing TAB
- As soon as the grid scrolls with your TAB (i.e. when you just edited the last visible column) you lose the data you just entered

Can you please check if you can see this behaviour too?

I'm using Chrome 46 on Win 8.

aiguofer:
Sorry, I just now got time to give it a try. Does it work as expected without my change? It seems to work fine on FF 42 and Chrome 46 on Fedora 23. I'm noticing some strange behavior where sometimes it loses focus after it's in switched to a new cell using TAB, but this seems to be the case with or without my change.

imbalind:
@aiguofer I remember it was working as expected without your change, anyhow I tried right now with another device and I can confirm that the current version is working correctly.

FF 43 and Chrome 47 on Win 7.

On a side note, in FF 43 I can confirm the strange behaviour you're reporting: editable cell sometimes losing focus right after being focused, but it's not related to your change.

aiguofer:
That's good to hear that it's working as expected now. Wonder if more testing needs to be done with other versions before merging?

imbalind:
I'm afraid I was not clear enough :smile: your version has still problems, master versione is still ok.

I think you could close this if you experience the same.

aiguofer:
Inbox <project.157159340.6137081@todoist.net>

mportuga:
Closing this since the discussion indicates that the issue is no longer present.

aiguofer:
Sorry for not having responded again earlier; my fork worked for me so I just let it be. I just had to update the app and tried the latest version and this issue still occurs. After a bit more digging I figured out the problem, although I'm still not sure what the "right" fix is.

I have API calls (that take a while) based on changes to the editable data; because of this, I didn't want it fetching data on each keystroke so I set:  
```
editableCellTemplate: "<div><form name=\"inputForm\"><input type=\"INPUT_TYPE\" ng-class = \"'colt' + col.uid\" ui-grid-editor ng-model=\"MODEL_COL_FIELD\" ng-model-options = \"{ updateOn: 'blur' }\"></form></div>"
```

The problem is that the `blur` event is never emitted, which is why this fix works for me. If this is enough of an edge case I can just keep my own fork on this, although I still didn't experience the problems that @imbalind pointed out with my change implemented.

aiguofer:
So I've been playing around with this a bit more... both of the following work for me to get the expected behavior when setting `ng-model-options={ updateOn: 'blur' }`, and neither changes the behavior on `Tutorial 301`.

I'm using Chrome Version 53.0.2785.143 (64-bit) and Firefox 49 on Fedora 24. Sadly I do not have a windows machine to test.

```
                $scope.stopEdit = function (evt) {
                  evt.target.blur();
                  if ($scope.inputForm && !$scope.inputForm.$valid) {
                    evt.stopPropagation();
                    $scope.$emit(uiGridEditConstants.events.CANCEL_CELL_EDIT);
                  }
                  else {
                    $scope.$emit(uiGridEditConstants.events.END_CELL_EDIT);
                  }
                  $scope.deepEdit = false;
                };
```

or 

```
                $scope.stopEdit = function (evt) {
                  if ($scope.inputForm && !$scope.inputForm.$valid) {
                    evt.stopPropagation();
                    $scope.$emit(uiGridEditConstants.events.CANCEL_CELL_EDIT);
                  }
                  else {
                    evt.target.blur();
                    $scope.$emit(uiGridEditConstants.events.END_CELL_EDIT);
                  }
                  $scope.deepEdit = false;
                };
```

Issue #4687, opened 2015-11-13
Memory leaks - grid object not getting destroyed


mloutraris-ge:
We have noticed that the grid data is not getting destroyed causing the memory to grow. You can see this by using the tutorial.

Steps to recreate: (using chrome with debugger window open)
1. Bring up the tutorial: http://ui-grid.info/docs/#/tutorial 
2. Using the profiler, take a heap snapshot. This will be the baseline.
3. Click on the 190 Large Dataset link on the left. Take another snapshot. Note that the heap size goes up.
4. Click to 098 Contributing to ui-grid.  Take another snapshot. Note that the heap size is still high.
5. Click on the 190 Large Dataset link on the left. Take another snapshot. Note that the heap size goes up even more.
6. Click to 098 Contributing to ui-grid.  Take another snapshot. Note that the heap size is still high.

Obviously you can keep this going for ever and it continues to grow.

If you look at the snapshots this is what I see:

1. Original profile:
![image](https://cloud.githubusercontent.com/assets/10615622/11137477/ef8614ca-8985-11e5-8cca-17880fefecbe.png)
2. After loading large data:
![image](https://cloud.githubusercontent.com/assets/10615622/11137499/32c79614-8986-11e5-9778-88503e56a8f2.png)
3. After going to 098 Contributing to ui-grid (grid was destroyed)
![image](https://cloud.githubusercontent.com/assets/10615622/11137523/764e1ee4-8986-11e5-94db-dccb246e24ea.png)
4. After going back to 190 Large Dataset link (data doubles):
![image](https://cloud.githubusercontent.com/assets/10615622/11137552/bbf40148-8986-11e5-92f8-4e0373c45d31.png)
5. After going to 098 Contributing to ui-grid (grid was destroyed)
![image](https://cloud.githubusercontent.com/assets/10615622/11137572/f89c5fa0-8986-11e5-99c0-05807640c52e.png)

We are seeing this in our application that uses the ui-grid. I have not been able to track down where the leak is. Has anyone seen this or knows the the issue might be?

Thanks

imbalind:
I confirm this bug and I think I found the culprit.

At [line 46 of uiGridMenu](https://github.com/angular-ui/ui-grid/blob/9d1a0b46e54983e09e0be8f89f9fac0ee74046e4/src/js/core/directives/ui-grid-menu.js#L46) we access `this` inside of an `apply` method.

[...]

PR #4704, opened 2015-11-16
4687 grid column leak

oriondean:
Fixes another leak found in issue #4687 (leak begins when switching between two pages with grids present).

Caused by GridColumn's hideColumn + unsort methods overriden prototype of all constructed GridColumns each time a GridColumn is constructed. Each of these methods kept a reference to grid via the self property in the closure scope.

### DIFF
--- a/src/js/core/factories/GridColumn.js
+++ b/src/js/core/factories/GridColumn.js
@@ -97,16 +97,6 @@ angular.module('ui.grid')
 
     self.updateColumnDef(colDef, true);
 
-    /**
-     * @ngdoc function
-     * @name hideColumn
-     * @methodOf ui.grid.class:GridColumn
-     * @description Hides the column by setting colDef.visible = false
-     */
-    GridColumn.prototype.hideColumn = function() {
-      this.colDef.visible = false;
-    };
-
     self.aggregationValue = undefined;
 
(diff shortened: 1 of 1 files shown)


src/js/core/factories/GridColumn.js
```diff
@@ -195,6 +185,16 @@ angular.module('ui.grid')
     };
   }
 
+  /**
+   * @ngdoc function
+   * @name hideColumn
+   * @methodOf ui.grid.class:GridColumn
+   * @description Hides the column by setting colDef.visible = false
+   */
+  GridColumn.prototype.hideColumn = function() {
```
oriondean: I couldn't see anywhere in the codebase where this method is actually called...

src/js/core/factories/GridColumn.js
```diff
@@ -806,15 +806,17 @@ angular.module('ui.grid')
         }
       });
     }
+  };
 
-    // Remove this column from the grid sorting, include inside build columns so has
-    // access to self - all seems a bit dodgy but doesn't work otherwise so have left
```
oriondean: Moving this method out of the build columns method seems to work. It's only called in one place and it has the correct context ($scope.col.unsort())

PR #4711, opened 2015-11-16
fix(core): correct filter detection in autoAdjustHeight

nickescallon:
defining a both a `filter` or `filters` property on a `columnDef` while alongside `options.enableFiltering === true` was resulting in one extra row's worth of height being calculated in `autoAdjustHeight`


nickescallon:
Fixes https://github.com/angular-ui/ui-grid/issues/4726

nickescallon:
@c0bra anything else I need to do to this PR? Still having to work around this bug locally in our app.

Issue #4724, opened 2015-11-18
[ui.grid.expandable] Undocumented row.isExpanded option for ui.grid.GridRow


joeskeen:
When looking at the sample code for the expandable tutorial: http://ui-grid.info/docs/#/tutorial/216_expandable_grid, I came across a check for `row.isExpanded` (app.js line 76).  This option isn't documented anywhere in the API, but I have tested it in my own project and it seems to work as expected.
I noticed that `expandable` is marked as 'Alpha', so I am wondering if this isn't in the API because there are plans to change it, or just because no one has documented it yet.  I would be happy to contribute if it is the latter, as I would also like to add it to the DefinitelyTyped definition for ui-grid as well. (see https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/ui-grid/ui-grid.d.ts#L1488)

PaulL1:
There is no current plan to change the expandable API, it's alpha because it doesn't really have any contributors that work on it, and it has a number of known defects.  We'd welcome a contribution that added documentation to the code.

PaulL1:
Thanks, I've merged.

PR #4729, opened 2015-11-18
docs(expandable): add documentation for row.isExpanded

joeskeen:
This is my first pull request for ui-grid, so hopefully I did it right.  I'm unfamiliar with your API documentation format.

This is a documentation-only change fixing issue #4724.


PaulL1:
Normally we'd also look for a squash - we prefer to have a single commit.  But since there are two bits to your commit.....I'll let it slide.  :-)  

Thanks for the contribution.

PR #4740, opened 2015-11-20
Fix cell focus in cellNav when a new row is added

ahocquet:


mportuga:
@JLLeitschuh Do you see any potential issues with this?

JLLeitschuh:
Looks good.

Issue #4607, opened 2015-10-30
Bug: when selection: true + grouping


billyaraujo:
When  selection is set to true, the grouping expand/collapse doesn't work properly.

http://plnkr.co/edit/6LlaQrXSnxhADTUOYfvJ?p=preview

If this isn't supported should give warning or something.

billyaraujo:
If this isn't supported should give warning or something.

swalters:
It works properly in this tutorial. Group on something and you can see it work.  Can you compare your gridOptions to this tutorial?  http://ui-grid.info/docs/#/tutorial/401_AllFeatures

billyaraujo:
enableRowHeaderSelection: false,

friedbizkvit:
"enableFullRowSelection: false" helps, but in my case i need full row selection to be enabled.

billyaraujo:
me too.

nickxbs:
The event is disabled by the selector ` .ui-grid-cell.ui-grid-disable-selection.ui-grid-row-header-cell{pointer-events: none;}`

imbalind:
This also happens for selection header.

billyaraujo:
Why is this closed?

PaulL1:
It's not.

billyaraujo:
Oh my bad. Sorry. Saw the close tag and thought it was.

sreekumarp:
@bgou: is #4607 (current issue) is also resolved by @4693 fix ?
If yes any idea when this will be released ?

bgou:
@sreekumarp I don't think this issue (#4607) is resolved yet. My reported issue (#4693) was closed as a duplicate to this one.

fk-mbc:
I have the same problem: http://plnkr.co/edit/TLilMmj3yWnyOaDStVU9?p=preview
When i use suggested solution with enableRowHeaderSelection: false then i can open the groups, but doesn't have the ability to select a row on click.

benjamin-wilson:
Proposed a fix: https://github.com/angular-ui/ui-grid/pull/4760

imbalind:
Merged with commit f715767

PR #4760, opened 2015-11-23
fix(Tree View) Row expansion when Tree View and Select both enabled

benjamin-wilson:
https://github.com/angular-ui/ui-grid/issues/4607
When Tree View and Select are both enabled, the Tree View expand button
did not work. The Tree View expand button was in a cell header class
who's click events are disabled. Enabling the pointer events on the
tree-header-row class resolves the issue.

### DIFF
--- a/src/features/tree-view/less/tree-view.less
+++ b/src/features/tree-view/less/tree-view.less
@@ -3,3 +3,6 @@
 .ui-grid-tree-header-row {
   font-weight: bold !important;
 }
+.ui-grid-tree-header-row .ui-grid-row-header-cell.ui-grid-disable-selection.ui-grid-cell {
+  pointer-events: all;
+}


dlgski:
I found this fix was done in the wrong order.  I found with: 

					enableRowSelection: true,
					enableFullRowSelection: true,
					enableRowHeaderSelection: false,

my cell is rendered with the following div:

    <div ng-repeat="(colRenderIndex, col) in colContainer.renderedColumns track by col.uid" ui-grid-one-bind-id-grid="rowRenderIndex + '-' + col.uid + '-cell'" class="ui-grid-cell ng-scope ui-grid-disable-selection ui-grid-coluiGrid-0007 ui-grid-row-header-cell" ng-class="{ 'ui-grid-row-header-cell': col.isRowHeader }" role="rowheader" ui-grid-cell="" tabindex="-1" id="1450099884304-0-uiGrid-0007-cell">

note, in CSS the order of the classes is important so the fix here did not match.  You need to re-order the classes like so:

    .ui-grid-tree-header-row .ui-grid-cell.ui-grid-disable-selection.ui-grid-row-header-cell {
      pointer-events: all;
    }

to catch this fix

dlgski:
also note, this fix is to override:  (in expandable.less)

     .ui-grid-cell.ui-grid-disable-selection.ui-grid-row-header-cell{
          pointer-events: none;
      }

so it makes sense to keep the order of the classes the same

PR #4807, opened 2015-12-03
fix(core): Sort Priority Zero Based

### LINKED ISSUE #4685: Sort number does not start from 0 when there is default sorted column ui-grid 3.0
When sorting is enabled by default for a column, the sorting number is 0. -> OK
But when I clicked another column, the sorting number changes to 1 while there is only one sorted column. -> It should be 0.

See this plunker:
http://plnkr.co/edit/q0Dhe9MLzgteYqdjtq6r?p=preview

JLLeitschuh:
Before sort priority was a mix of 0 and 1 based.
You could create columns with a sort priority of zero but once the user
clicked on a column to sort it its index would change to 1. There was no
way for the sort index to return to 0 from within the grid's menus.
This mix of 0 and 1 based sorting was confusing and created UI problems
when displaying the sort index in the column header.

Closes #4685
BREAKING CHANGE:
**GridOptions.columnDef.sort.priority** now expects the lowest value
to be 0.
The Grid Header will display a sort priority of 0 as 1.
Using `if(col.sort.priority)` to determine if a column is sorted is no
longer valid as `0 == false`.
Saved grid objects may be affected by this.

### DIFF
--- a/src/js/core/factories/Grid.js
+++ b/src/js/core/factories/Grid.js
@@ -397,7 +397,7 @@ angular.module('ui.grid')
      * @methodOf  ui.grid.core.api:PublicApi
      * @description The visibility of a column has changed,
      * the column itself is passed out as a parameter of the event.
-     * 
+     *
      * @param {$scope} scope The scope of the controller. This is used to deregister this event when the scope is destroyed.
      * @param {Function} callBack Will be called when the event is emited. The function passes back the GridCol that has changed.
      *
@@ -1087,17 +1087,17 @@ angular.module('ui.grid')
    *   append to the newRows and add to newHash
    *   run the processors
    * ```
-   * 
+   *
    * Rows are identified using the hashKey if configured.  If not configured, then rows
    * are identified using the gridOptions.rowEquality function
-   * 
+   *
    * This method is useful when trying to select rows immediately after loading data without
    * using a $timeout/$interval, e.g.:
-   * 
+   *
    *   $scope.gridOptions.data =  someData;
    *   $scope.gridApi.grid.modifyRows($scope.gridOptions.data);
    *   $scope.gridApi.selection.selectRow($scope.gridOptions.data[0]);
-   * 
+   *
    * OR to persist row selection after data update (e.g. rows selected, new data loaded, want
    * originally selected rows to be re-selected))
    */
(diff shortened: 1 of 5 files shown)


JLLeitschuh:
This has a breaking change. It would be good to do a version bump to 3.1.0 after this is merged.

src/js/core/factories/Grid.js
```diff
@@ -1876,12 +1876,12 @@ angular.module('ui.grid')
         p = 0;
 
     self.columns.forEach(function (col) {
-      if (col.sort && col.sort.priority && col.sort.priority > p) {
-        p = col.sort.priority;
+      if (col.sort && col.sort.priority !== 'undefined' && col.sort.priority >= p) {
```
imbalind: I think here you should compare `col.sort.priority` to `undefined` and not `'undefined'` or `typeof col.sort.priority === 'undefined'`

JLLeitschuh: Thanks! I'll change that. I saw it like this somewhere else.

swalters:
@JLLeitschuh can you fix the merge issues?  I think we need to get this merged and released as soon as we can. thanks.

JLLeitschuh:
Yea, I'll try to fix this tonight.

src/templates/ui-grid/uiGridHeaderCell.html
```diff
@@ -21,13 +21,13 @@
       aria-label="{{getSortDirectionAriaLabel()}}">
       <i
        ng-class="{ 'ui-grid-icon-up-dir': col.sort.direction == asc, 'ui-grid-icon-down-dir': col.sort.direction == desc, 'ui-grid-icon-blank': !col.sort.direction }"
-       title="{{isSortPriorityVisible() ? i18n.headerCell.priority + ' ' + col.sort.priority : null}}"
+       title="{{isSortPriorityVisible() ? i18n.headerCell.priority + ' ' + ( col.sort.priority + 1 )  : null}}"
```
dlgski: In Grid.js line 1880 aren't you bumping up the sortPriority, so adding 1 again here would be incorrect?

src/templates/ui-grid/uiGridHeaderCell.html
```diff
@@ -21,13 +21,13 @@
...
        aria-hidden="true">
      </i>
      <sub
        ui-grid-visible="isSortPriorityVisible()"
        class="ui-grid-sort-priority-number">
-       {{col.sort.priority}}
+       {{col.sort.priority + 1}}
```
dlgski: same as above

PR #4819, opened 2015-12-07
Fixing error of iterating over non own properties in render containers.

kedrzu:
This error caused grid not to render at all.


JLLeitschuh:
Can you provide more details about this problem. A screenshot or GIF would be helpful and/or a plunker demo.

PR #4828, opened 2015-12-09
- enableSorting grid default option will be overridable by column def…

idangozlan:
- enableSorting grid default option will be overridable by column def, so enableSorting could be set to specific column.


mportuga:
In the future please add an issue number to your commit.

PR #4830, opened 2015-12-09
enable to show/hide sorting priority on default grid options

idangozlan:
enable to show/hide sorting priority on default grid options

### DIFF
--- a/src/templates/ui-grid/uiGridHeaderCell.html
+++ b/src/templates/ui-grid/uiGridHeaderCell.html
@@ -25,7 +25,7 @@
        aria-hidden="true">
      </i>
      <sub
-       class="ui-grid-sort-priority-number">
+       class="ui-grid-sort-priority-number" ng-show="grid.options.enableSortingPriority">
        {{col.sort.priority}}
      </sub>
     </span>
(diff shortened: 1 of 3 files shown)


src/templates/ui-grid/uiGridHeaderCell.html
```diff
@@ -25,7 +25,7 @@
        aria-hidden="true">
      </i>
      <sub
-       class="ui-grid-sort-priority-number">
+       class="ui-grid-sort-priority-number" ng-show="grid.options.enableSortingPriority">
```
JLLeitschuh: This should be on two lines.

test/unit/core/factories/GridOptions.spec.js
```diff
@@ -40,6 +40,7 @@ describe('GridOptions factory', function () {
         wheelScrollThrottle: 70,
         scrollDebounce: 300,
         enableSorting: true,
+        enableSortingPriority: true,
```
JLLeitschuh: I think that there are more of these than this that need to be modified.

JLLeitschuh:
- [ ] The tests need to be fixed
- [ ] This should be squashed into one commit

Issue #3747, opened 2015-06-10
white space when the rows ends.


hummorsa:
Hi, 

I am kinda new using ng-grid. I have a grid that is creating a white gap inside the grid view port after the last row is rendered, so is creating extra vertical scrolling. 

Check where the yellow are is in this image: http://postimg.org/image/lmqdsu2rh/

Checking I realize that the 'ui-grid-canvas' class have a height set of 630px, removing that height solve the problem, but there is a option in the configuration to solve that ? 

The container of the grid has a set height.   

Thanks for any help.

JLLeitschuh:
To make it easier to see: ![Grid using too much space](http://s22.postimg.org/yqvy5iutd/extraspace.png)

JLLeitschuh:
There is currently a todo about this: https://github.com/angular-ui/ng-grid/blob/master/src/js/core/directives/ui-grid-render-container.js#L120

lexigren:
Is there some fix for this problem? Or what use cases with grid should be avoided to not trigger this bug?

ticorunner:
Has this been fixed already with a most recent release of ui-grid ?

irfan:
Hi team,

Any update? Should I fix and open a PR?

mportuga:
Closed by the following PR: https://github.com/angular-ui/ui-grid/pull/5710

PR #4833, opened 2015-12-09
Fix issue with space after the last table row

Nilanno:
Can't find out a reason of issue, but it's reproduced easily in our application. This code fixes it.

### DIFF
--- a/src/js/core/directives/ui-grid-viewport.js
+++ b/src/js/core/directives/ui-grid-viewport.js
@@ -114,12 +114,12 @@
 
             var styles = {};
 
-            if (index === 0 && rowContainer.currentTopRow !== 0) {
-              // The row offset-top is just the height of the rows above the current top-most row, which are no longer rendered
-              var hiddenRowWidth = (rowContainer.currentTopRow) * rowContainer.grid.options.rowHeight;
-
-              // return { 'margin-top': hiddenRowWidth + 'px' };
-              styles['margin-top'] = hiddenRowWidth + 'px';
+            if (rowContainer.currentTopRow !== 0){
+              //top offset based on hidden rows count
+              var translateY = "translateY("+ (rowContainer.currentTopRow * rowContainer.grid.options.rowHeight)  +"px)";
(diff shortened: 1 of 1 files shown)


JLLeitschuh:
Can you describe the issue in more detail? Include screenshots a GIF or a plunker demo.

Nilanno:
Here is an image - http://prntscr.com/9ehrsv

Behavior was pretty strange: if scroll had 50% of container scroll at beginning, he was smaller and smaller with scroll down so at the end of the scroll he had height near a half of initial.
So it looked like an infinity scroll without a data. All table data was preloaded. No table options about paging and scroll helped. If I set virtualizationThreshold > data.length then issue didn't appear and scroll worked nice.

Sorry, can't reproduce it in plunker.
I've found this fix in one of the forked repositories, but that repo was based on grid version 2(ngGrid)

Nilanno:
@JLLeitschuh , toggle "waiting-response" status

Havock94:
The same happens to me and it seems that the more data you load, more space is added.
The scrollbar keeps becoming smaller and smaller but the second half of the grid is empty as @Nilanno screenshot shows. Was it fixed yet?

Nilanno:
@Havock94, there were no new releases since this PR was opened. You can fix it temporarily by patching your local library with a code in this pull request.

Nilanno:
Right now I can't reproduce this issue by rolling back local code, maybe it was a browser bug.
Nevertheless, translateY is a faster than padding in browsers work.

irfan:
Hi All,

Are you planning to merge this PR? Is this PR fixes https://github.com/angular-ui/ui-grid/issues/3747 ?

mportuga:
@dlgski This will have a conflict with Gennadi's PR that changes this logic completely. I suppose we can merge this if we decide not to merge that one. Otherwise we can merge this PR.

mportuga:
@dlgski Actually, I just compared this to the grid scrolling change and they can both be pushed. No conflicts.

Issue #4811, opened 2015-12-04
Internationalization incorrect


adrien-be:
Hi guys,
I noticed that the column sorting is incorrect if I add an item with a diacritic, for instance "élodie".
Then add another item without diacritic and starting with the same letter but without diacritic, for instance "elsa".
One would expect these two to consistently be one after another. But whenever I reorder the column, their positions keep changing quite illogically.

I stumbled upon this string sorting in the past & it's a much harder task if internationalized.
You can read a simple solution on http://stackoverflow.com/questions/51165/how-do-you-do-string-comparison-in-javascript/26295229#26295229
Cheers,
Adrien

imbalind:
Could you please edit one of the plunker in the tutorials so to show this behaviour?

adrien-be:
Just tried it on Plunker, it actually looks fine: all the characters with a diacritic are consistently added in the same sort order. The sorting chosen can be considered awkward but that's another topic.

http://plnkr.co/edit/hOTVATmXHGMlViscMWDB?p=preview

I believe I missed the fact that the 3rd click puts the data back to the order in which it initially was.

imbalind:
I guess your suggestion still holds about the sorting order and the use of `localeCompare()`.

I'm closing this, feel free to open another issue for the sorting order.

adrien-be:
Not sure what sort order are implemented in each browser though. So I won't mention that, but `localeCompare()` definitely is the way forward.

PR #4849, opened 2015-12-13
Update sorting so it can support internationalization

adrien-be:
see why localCompare() is the best solution http://stackoverflow.com/a/26295229/759452

Fix as mentioned in https://github.com/angular-ui/ui-grid/issues/4811#event-490369609


JLLeitschuh:
This make sense. Thanks!

adrien-be:
You're welcome. Glad to see better internationalization spreading around :)

PR #4854, opened 2015-12-14
re-order styles for tree-header-row override

dlgski:
In fix #4760 an override was introduced to allow clicking on the tree-expansion when row selection is on.  

The generated code with the grid options:

    enableRowSelection: true,
    enableFullRowSelection: true,
    enableRowHeaderSelection: false,

render each cell with the following classes:

    class="ui-grid-cell ng-scope ui-grid-disable-selection ui-grid-coluiGrid-0007 ui-grid-row-header-cell"

so, the order of these classes needs to have `ui-grid-cell` first, followed by `ui-grid-disable-selection` then `ui-grid-row-header-cell` in order for this override to work

### DIFF
--- a/src/features/tree-view/less/tree-view.less
+++ b/src/features/tree-view/less/tree-view.less
@@ -3,6 +3,6 @@
 .ui-grid-tree-header-row {
   font-weight: bold !important;
 }
-.ui-grid-tree-header-row .ui-grid-row-header-cell.ui-grid-disable-selection.ui-grid-cell {
+.ui-grid-tree-header-row .ui-grid-cell.ui-grid-disable-selection.ui-grid-row-header-cell {
   pointer-events: all;
 }


AgDude:
Will you please provide a plunker (or tutorial page) demonstrating the fix. Was it simply a mistake that #4760 was merged, and it never worked, or did it work in some cases? I don't see any link in that PR or issue demonstrating that it works.

dlgski:
I actually have no idea if it ever worked.  It took me a while on my end to find this error, I'd be curious to hear from the developer who fixed #4760 and see if there is a difference in my implementation versus theirs.
I will put together a plunker
Thanks,-Dana 


    On Monday, December 14, 2015 11:30 AM, Nate Dudenhoeffer <notifications@github.com> wrote:
 

 Will you please provide a plunker (or tutorial page) demonstrating the fix. Was it simply a mistake that #4760 was merged, and it never worked, or did it work in some cases? I don't see any link in that PR or issue demonstrating that it works.—

dlgski:
Here is a plunker:

http://plnkr.co/edit/oLB8m4ngQFTxdQ98vlCJ?p=preview
Click on "Refresh Rows" to populate the grid.  The group by Age (or another column).  Click on the + icon on a row and it will not expand.

Now try this:
http://plnkr.co/edit/rc1HzaMnF2odGptWxj39?p=preview
This plunker has my simple CSS fix.  

Thanks,
-Dana

JLLeitschuh:
Oh cool! Thanks for the plunker.

AgDude:
Thanks for the fix, and providing the requested example.

PR #4876, opened 2015-12-17
feat(sort): change sort priority indicator to only show when multiple columns are sorted

### LINKED ISSUE #4503: Sort Priority Numbers
The new sort priority numbers are pretty sweet, but I think it would be better if they didn't show up when there's only a single sort.

jbarrus:
Sort priority number indicators were added with #4257.  These look strange when only one column is sorted.  This pull request changes the indicators to only display when multiple columns are sorted.

fixes #4503

### DIFF
--- a/src/js/core/directives/ui-grid-header-cell.js
+++ b/src/js/core/directives/ui-grid-header-cell.js
@@ -31,14 +31,20 @@
               headerCell: i18nService.getSafeText('headerCell'),
               sort: i18nService.getSafeText('sort')
             };
+            $scope.isSortPriorityVisible = function() {
+              //show sort priority if column is sorted and there is at least one other sorted column
+              return $scope.col.sort.priority && $scope.grid.columns.some(function(element, index){
+                  return element.sort.priority && element !== $scope.col;
+                });
+            };
             $scope.getSortDirectionAriaLabel = function(){
               var col = $scope.col;
               //Trying to recreate this sort of thing but it was getting messy having it in the template.
               //Sort direction {{col.sort.direction == asc ? 'ascending' : ( col.sort.direction == desc ? 'descending':'none')}}. {{col.sort.priority ? {{columnPriorityText}} {{col.sort.priority}} : ''}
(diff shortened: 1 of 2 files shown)


JLLeitschuh:
This should be toggleable with a setting. Also this will conflict with #4807 (but I guess I'll have to take care of that)

jbarrus:
What should the name of the setting be?  hideSingleColumnSortIndicator?

I'm still having a hard time seeing why anyone would turn this off.  Are there any other examples of grids that show numeric sort indicators like this when only one column is sorted?

swalters:
@JLLeitschuh I agree with the comment above. Is there a reason we need to always show a sort indicator of 1 if there is only one column sorted?

jbarrus:
So does someone need to make a decision?

dlgski:
I like this idea.  If more than 1 sort is provided, add numbers, (indexing at 1) otherwise do not show priority

JLLeitschuh:
Okay, this makes sense. I see no problem with merging this.

dlgski:
There is something off with this solution.  I just pulled 3.1.0 down to test this.

1.  I sort on 1 column  --> I get an arrow indicator with no priority number shown  (GOOD)
2.  I sort on a 2nd column -->  First sort remains with same indicator, new column gets an arrow indicatory  (BAD)  ==> expected result: once I sort on a 2nd column I should see a "1" next to my first column and a "2" next to my second
3.  I sort on a 3rd column --> First column remains with no priority indicator, second column gets a "1" indicator, and 3rd column gets a "2" indicator (BAD)  ===> expected result: I should see sort priorities "1", "2", "3"

dlgski:
The reason this is not working properly is because in ui-grid-header-cell.js you are testing for:
```
$scope.grid.columns.some(function(element, index){
                  return element.sort.priority !== undefined && element !== $scope.col;
                });
```
but if element.sort.priority === 0 then this condition is returning 0 which will result with `isSortPriorityVisible` resolving as 0 or "false" when it should be "true"

dlgski:
@JLLeitschuh where is your fix to index sort.priority at 1 instead of 0?  If this went in, then the condition above would resolve to true and there would be no bug

Issue #2053, opened 2014-11-10
using a url in cellTemplate in columnDef causes grid to malfunctions.


joshmackey:
If I include the html inline, it works fine, but as soon as its a remote file, the entire grid malfunctions having the headers be only a few pixels in height and the contents of the grid be invisible.

prathik457:
I am also facing this issue. 
I wanted auto-focus on the input element on click of filter button(external filter). So I explicitly defined the headerCellTemplate in columnDefs. I am keeping the html code in a separate template and I am trying to reference the in columnDefs.
In the console it says fetching URL and does not actually apply the template in the header.

PR #4902, opened 2015-12-23
fix(uiGrid): Fix race condition in data watcher

### LINKED ISSUE #4532: Custom column/template not being rendered after v3.0.7
After updating to v3.0.7, one of my columns (which makes use of custom `headerCellTemplate` and `cellTemplate` in the column definition) is no longer being rendered (the column shows up blank and there is no HTML for it).

The column is initially pinned to the left (`pinnedLeft: true` in column definition) and I use it as a custom selection column (one that I can control the filtering/behavior of). The templates are based on the default templates with minor differences.

The related change is #4428 (see issue #4386). Particularly [here](https://github.com/angular-ui/ui-grid/pull/4428/files#diff-d586707ee1c4e842337e09afc6ebd587L108):

###### Before
```javascript
// If we either have some columns defined, or some data defined
if (self.grid.options.columnDefs.length > 0 || newData.length > 0) {
  // Build the column set, then pre-compile the column cell templates
  promises.push(self.grid.buildColumns()
    .then(function() {
      self.grid.preCompileCellTemplates();
    }));
}
```

###### After
```javascript
// If we haven't built columns before and either have some columns defined or some data defined
if (!hasColumns && (self.grid.options.columnDefs.length > 0 || newData.length > 0)) {
  // Build the column set, then pre-compile the column cell templates
  promises.push(self.grid.buildColumns()
    .then(function() {
      self.grid.preCompileCellTemplates();
    }));
}
```

If I remove `!hasColumns` from the if statement, my column renders again. Is the logic being implemented wrong? I don't understand enough about what @oriondean was trying to do here.

oriondean:
I introduced a guard to reduce some unnecessary calls to `buildColumns` and `preCompileCellTemplates` each time data is changed. 

As a result, the `$q.all` promise in `dataWatchFunction` could resolve **before** promises created in previous calls. This is because an empty array of promises can be passed to `$q.all` the method now. This unearthed an issue where `modifyRows` could be passed outdated data from previous `dataWatchFunction` calls, rather than the most recent call.

Fixes #4532, issue introduced by #4428.

### DIFF
--- a/src/js/core/directives/ui-grid.js
+++ b/src/js/core/directives/ui-grid.js
@@ -78,6 +78,8 @@
         }
       }
 
+      var mostRecentData;
+
       function dataWatchFunction(newData) {
         // gridUtil.logDebug('dataWatch fired');
         var promises = [];
@@ -89,6 +91,8 @@
             newData = $scope.uiGrid.data;
           }
         }
+
+        mostRecentData = newData;
 
         if (newData) {
           // columns length is greater than the number of row header columns, which don't count because they're created automatically
@@ -118,7 +122,8 @@
           }
 
           $q.all(promises).then(function() {
-            self.grid.modifyRows(newData)
+            // use most recent data, rather than the potentially outdated data passed into watcher handler
+            self.grid.modifyRows(mostRecentData)
               .then(function () {
                 // if (self.viewport) {
                   self.grid.redrawInPlace(true);


imbalind:
@c0bra @swalters I'm not so deep into this, but some users have issues so I think merging this is quite high priority, can you have a look?

oriondean:
#2053 seems to be a similar issue that @c0bra fixed a year ago (https://github.com/angular-ui/ui-grid/commit/2b25f249cd133a02f44eec34ccee7f4e77adab16).

It seems that the fix then was to prevent the promises from resolving out of order. This fix was reverted by the changes made in #4428. This PR doesn't fix the promises resolving out of order but instead prevents the wrong data being set *if* they do.

PR #4918, opened 2015-12-30
Don't always disable rowHeaderSelection for fullRowSelection

Robinson7D:
I might be missing something obvious - and feel free to point it out - but it seems very strange to me that if you enable "full row selection", there is no way to click the row header anymore. 

It feels against the spirit of "full row", and also surprises many users trying to select/deselect rows. 

Regardless of that belief, it makes an explicit `enableRowHeaderSelection: true` setting not work. 

(Simpler fix would be just removing the `addClass('ui-grid-disable-selection')` line, but I expect it's there for a reason? If there's not a use-case for disabling them, I'd be happy to amend the commit to simply remove the lines.)

### DIFF
--- a/src/features/selection/js/selection.js
+++ b/src/features/selection/js/selection.js
@@ -893,7 +893,10 @@
 
             function registerRowSelectionEvents() {
               if ($scope.grid.options.enableRowSelection && $scope.grid.options.enableFullRowSelection) {
-                $elm.addClass('ui-grid-disable-selection');
+                if (!$scope.grid.options.enableRowHeaderSelection){
+                  $elm.addClass('ui-grid-disable-selection');
+                }
+
                 $elm.on('touchstart', touchStart);
                 $elm.on('touchend', touchEnd);
                 $elm.on('click', selectCells);


swalters:
@Robinson7D testing this locally and see an issue.  If that class is not present, then shift+click results in all the cells being selected by the browser which doesn't look nice.

1-11-2016 11-24-24 am

Robinson7D:
Ah, you're right. I guess I was misunderstanding the word "selection" in that class.
It looks like (based on the tutorial examples) the goal is to make sure that none of the cells can have their text highlightable. A better way might be to use `user-select: none` instead of removing all touch events?

Example with prefixes: http://stackoverflow.com/a/4407335/624590

swalters:
sure, try it and see.  

You can add 'EnableFullRowSelection' button to the first grid in the 210 tutorial so we can easily test it in action.

```

        $scope.toggleFullRowSelection = function() {
          $scope.gridOptions.enableFullRowSelection = !$scope.gridOptions.enableFullRowSelection;
          $scope.gridApi.core.notifyDataChange( uiGridConstants.dataChange.OPTIONS);
        };

<button type="button" class="btn btn-success" ng-click="toggleFullRowSelection()">Toggle FullRowSelection</button>  <strong>rowSelection:</strong> <span ng-bind="gridApi.grid.options.enableFullRowSelection"></span>
      <br/>
```

Robinson7D:
Sure, I'll do that in a few minutes after standup

Robinson7D:
Going to need to put this on hold for a day or two.

Digging deeper into the issue, it seems that the class is already using the CSS that I just suggested (in `selection.less` you have it)

However, `expandable.less` has:

```
.ui-grid-cell.ui-grid-disable-selection.ui-grid-row-header-cell{
  pointer-events: none;
}
```

Presumably, that's to prevent "expanding" from triggering a row selection (taking a different meaning from the word "selection" from the text-selection prevented before), however it seems to have caused casualties due to it being a generic selector that matches all of the headers. 

To fix this, we would need to figure out a good way to get a class on that particular header, so that expandable only affects the correct cells (or selection header ignores it).

Robinson7D:
In essence, "selection" in `.ui-grid-disable-selection` seems to mean two things by different parts of the code: 
1) rowSelection uses that class to mean disable selecting the text (via drag)
2) Expand uses the same class to mean disable selecting the row (via click)

Probably best to have separate classes for the two types of things that are attempted here.
Perhaps expand should use `.ui-grid-disable-row-selection`. Will try changing one of them later.

mportuga:
@Robinson7D What is the status on this?

Robinson7D:
It still happens and is still caused by the (Beta) Grouping feature's CSS clashing with the Selection feature. 

However, I didn't find time to fix up the PR, and instead just overrode the CSS in my application code (because my company decided I had spent enough time in there already).

Feel free to take it over if you wish.

mportuga:
@Robinson7D Thanks for the response. I will close this PR for now and will look into making the changes you are suggesting.

PR #4948, opened 2016-01-12
Add option when exporting to get the cell display value rather than the source value

ruhley:
Supersedes #4820


PaulL1:
Nice work, thanks.

PR #4961, opened 2016-01-13
Tutorial: 312 Exporting Data With Custom UI 

mbellotto:
Hi,

HTML code of this tutorial has some errors. SELECT end tag is duplicated at lines 118, 123 & 129.

A semicolon also missing at line 136, after &nbsp.

Regards


PaulL1:
There's something somewhat weird with this pull request - but ultimately it appears to be a request to merge 2.x into master, which we wouldn't do.  

The tutorial that is referred to is auto generated, so we wouldn't normally edit the HTML directly.  Unless I'm missing what the request seems to be.

mbellotto:
Hi |Paul!

a few weeks ago, Matteo Brassini asked me to make a pull request about this. To be honest I never use this option so maybe I made a mistake.

I copied the code of this tutorial from GitHub and I found there was an error.

Please, let me know if I have to make this pull request in some other place.

Thanks

swalters:
I suggest you start over with another PR.  This one can't be merged

PR #4968, opened 2016-01-14
Prevent selection triggering on plus/minus click

FLinares:
When grouping and enableRowSelection is set to true clicking on plus/minus icons selects/deselects the row. This seems counter intuitive.

The problem was fixed by detecting where the click came from; in the case of an icon, it is ignored, otherwise a select/deselect action is triggered.


JLLeitschuh:
I feel like this could be done better without explicitly depending upon the CSS class names.

PR #4987, opened 2016-01-18
The fix for 4973

YonatanKra:
Calculate the body container (the not pinned width) and only move the
selected column header when we are in its boundaries.


mportuga:
@YonatanKra Please fix your unit tests.

mportuga:
Closing due to inactivity.

PR #5000, opened 2016-01-20
fix(core): check for undefined on sort priority

dlgski:
in Issue #4876 isSortPriorityVisible() is introduced.  However it is testing the variable rather than for undefined, so when element.sort.priority == 0, it is returning as false, even though that should return true.

This update is adding the checks for undefined

@jbarrus @JLLeitschuh

### DIFF
--- a/src/js/core/directives/ui-grid-header-cell.js
+++ b/src/js/core/directives/ui-grid-header-cell.js
@@ -33,8 +33,8 @@
             };
             $scope.isSortPriorityVisible = function() {
               //show sort priority if column is sorted and there is at least one other sorted column
-              return $scope.col.sort.priority && $scope.grid.columns.some(function(element, index){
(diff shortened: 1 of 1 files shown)


dlgski:
A better fix for this would be to index sort.priority at 1 rather than 0, but I cannot find where sort.priority is getting initially set to 0.

If sort.priority was indexed at 1, then this code would have worked the way it was.  Also, having an indicator of "0" looks kind of odd.  (which this will result with)

swalters:
Would this PR fix the issue? https://github.com/angular-ui/ui-grid/pull/4807

dlgski:
I added some comments to that PR.  If that PR is indexing sortPriority at
1, then I think these checks will properly return "true" when sortPriority
exists

JLLeitschuh:
This makes sense since 0 would evaluate to false in this case. Okay, cool Thanks.

PR #5024, opened 2016-01-25
feat(expandable): Add 'expandRow' and 'getExpandedRows' 

rafaelb:
Features required for project I am currently working on.

### DIFF
--- a/src/features/expandable/js/expandable.js
+++ b/src/features/expandable/js/expandable.js
@@ -100,7 +100,7 @@
         /**
          *  @ngdoc object
          *  @name ui.grid.expandable.api:GridRow
-         * 
+         *
          *  @description Additional properties added to GridRow when using the expandable module
          */
         /**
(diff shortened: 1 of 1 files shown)


swalters:
@rafaelb please rebase this PR into one commit. thanks.
http://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git/5189600#5189600

rafaelb:
@swalters Done

JLLeitschuh:
Does this still work for all of the aria roles? Or do those need to be inverted as well?

swalters:
@rafaelb To be consistent with our other api methods, can you please change expandRow to toggleRow?   This will allow the method to expand or collapse a row.

rafaelb:
@swalters I think that toggling is already handled by this method: https://github.com/angular-ui/ui-grid/blob/6709846/src/features/expandable/js/expandable.js#L145

For the project we are working on, we needed the ability to expand rows (or not do anything if they are already expanded)
I based this method on the "selectRow" method of the selectable feature: 
https://github.com/angular-ui/ui-grid/blob/6709846/src/features/selection/js/selection.js#L177

What I think the best solution would be is to have "expandRow", "collapseRow" and  "toggleRow" methods.
What do you think?
If you agree, I will add a "collapseRow" method.

swalters:
@rafaelb You suggestion is good.

rafaelb:
@swalters Added 'collapseRow' method

PR #5043, opened 2016-01-28
performance(core): add false flag to $timeout and $interval to prevent $apply

swalters:
Nice increase in scrolling performance when using mousewheel or trackpad

### DIFF
--- a/src/js/core/services/ui-grid-util.js
+++ b/src/js/core/services/ui-grid-util.js
@@ -21,7 +21,7 @@ if (typeof Function.prototype.bind !== "function") {
   };
 }
 
-function getStyles (elem) {
+function  getStyles (elem) {
   var e = elem;
   if (typeof(e.length) !== 'undefined' && e.length) {
     e = elem[0];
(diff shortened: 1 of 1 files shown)


src/js/core/services/ui-grid-util.js
```diff
@@ -21,7 +21,7 @@ if (typeof Function.prototype.bind !== "function") {
   };
 }
 
-function getStyles (elem) {
+function  getStyles (elem) {
```
JLLeitschuh: Whitespace change that isn't really needed.

src/js/core/services/ui-grid-util.js
```diff
@@ -1225,7 +1225,7 @@ module.service('gridUtil', ['$log', '$window', '$document', '$http', '$templateC
     function runFunc(endDate){
       lastCall = +new Date();
       func.apply(context, args);
-      $interval(function(){ queued = null; }, 0, 1);
+      $interval(function(){ queued = null; }, 0, 1, false);
```
JLLeitschuh: If the goal of this is to remove the reference to queued shouldn't this be calling `$interval.cancel(queued)`

Perhaps I don't fully understand how this function works.

swalters: I'm not sure either.  I got this code on the net and it seemed to work ok.  The false flag is needed to prevent unnecessary digest

swalters: ok, looked at it some more. 
```
 function runFunc(endDate){
      lastCall = +new Date();
      func.apply(context, args);
      $interval(function(){ queued = null; }, 0, 1, false);
    }
```
the 3rd parm with value of 1 specifies the interval to execute only once which cancels the interval.

Issue #5041, opened 2016-01-28
columnDefs watcher corrupts its value


debrice:
Hey guys,

I've been having a bug where the columnDefs value becomes an integer after changing the array. I tracked it down to this watcher:

https://github.com/angular-ui/ui-grid/blob/7de930404c3cecd13984c06d0612c788991bb7c0/src/js/core/directives/ui-grid.js#L57

```js
deregFunctions.push( 
    $scope.$parent.$watch(
        function() { 
            return $scope.uiGrid.columnDefs.length; // return the length of the array
        }, 
        columnDefsWatchFunction // will receive the length of the array
    ) 
);
```

Because the method used as `watchExpression` returns the length of the array, the watcher receives the length of the column array and not the array itself. The callback reads the new value which alternate between the array itself and the length of the array:

![image](https://cloud.githubusercontent.com/assets/56964/12656645/3f43ad6e-c5b4-11e5-8611-0e716e71a5bb.png)

I'd be glad to provide a PR if nobody's on it already.

debrice:
Also, this bug only occurs when `grid.options.fastWatch` is `True`

lpand:
Why is the watcher function only checking the length of the array? Doesn't fastWatch mean checking reference equality, that is the collection of column definitions is immutable and can only be replaced? Checking the collection length may not be enough, e.g. columnDefs is replaced with a collection of same length.

debrice:
I can't talk for the author of the `fastWatch` mode but I believe this is way to ensure mutation on arrays are also detected at a very low cost.

debrice:
@lpand The column definition is not immutable

lpand:
@debrice maybe i expressed myself wrong. I meant that fastWatch, as far as i know, means only reference check is performed, otherwise it would be collection watching, wouldn't be?

I believe the [404 Large Data Sets](https://github.com/PaulL1/ng-grid/blob/f7d42039513932b38e14d63f767eb3965caa9691/misc/tutorial/404_large_data_sets_and_performance.ngdoc#L37) tutorial proves my point.

debrice:
@lpand both length and **reference** are check in the current implementation. Also the columnDefs and the data aren't immutable.

```js
deregFunctions.push( $scope.$parent.$watch(function() { return $scope.uiGrid.columnDefs; }, columnDefsWatchFunction) );
[...]

PR #5044, opened 2016-01-28
fix(#5041): Ensure length watchers transmit data

debrice:
When in fastWatch mode, watcher methods returns the length of their array being watched. When this value (`int`). This value (`int`) is then being passed to the watcher's callback overwriting the array.


PaulL1:
Looks good to me.

PR #5052, opened 2016-02-01
fix(saveState): Fixes #4146 - Allow saving of pagination state

### LINKED ISSUE #4146: saveState not saving current page size
As i checked, $scope.gridApi.saveState.save() does not save current page size value and restore method doesn't restore previous page size value. So that i think there will be a wrong position scrollbar when:
_Call method to save current state of grid. (current page size in 100 and current position of scrollbar in row 47th)
_Change the page size (change to 20)
_Call method to restore the state 
_The scrollbar will be shown in wrong position.

dsjellz:
Saves the page save and current page position when saveState is called, as long as pagination is enabled. Fixes #4146

### DIFF
--- a/src/features/saveState/js/saveState.js
+++ b/src/features/saveState/js/saveState.js
@@ -287,6 +287,7 @@
           savedState.selection = service.saveSelection( grid );
           savedState.grouping = service.saveGrouping( grid );
           savedState.treeView = service.saveTreeView( grid );
+          savedState.pagination = service.savePagination( grid );
 
           return savedState;
         },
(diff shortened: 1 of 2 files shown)


JLLeitschuh:
Looks great! Thanks!

syedrehman:
Hi Everyone, I am using angular UI grid with Pagination. My issue is. if i am on page 3 and update a record using modal popup. Grid is getting reloaded with updated record.
But , The problem is Grid is displaying  with page 1 . I would like to see the  updated data grid  with page 3 .not with page 1. Could anyone help me out please.

Issue #3782, opened 2015-06-15
TypeError: styles is null


craftyInstigator:
This is an issue specific to Firefox that was detected in Firefox 38, and has been reproduced using ui-grid 3.0.0-rc.22:

If the ui-grid is located in an iframe, and that iframe is hidden via ng-show/hide (styled to display: none !important), there is an error in the console:

TypeError: styles is null

JLLeitschuh:
It would be helpful if you were to provide a minimal plunker that demonstrates this issue.

craftyInstigator:
http://plnkr.co/edit/pBmVUQbOu9ma5zCr89mF?p=preview
In Firefox just push the Hide button. I'm almost certain it's caused by the auto-resizing module, since I can't reproduce it without auto-resize.

PaulL1:
Have you looked at the hidden grids tutorial?  http://ui-grid.info/docs/#/tutorial/108_hidden_grids

craftyInstigator:
No, since it isn't related to my problem. The problem, as demonstrated in the plunker, has nothing to do with hiding the ui-grid. This is about a console error that occurs when the display of the whole iframe that happens to have ui-grid in it with auto-resizing is hidden. This is a defect on the auto-resizing module where it's missing a null check for the styles variable. This seems to be specific to Firefox, and the availability of the styles variable when an iframe display is none.

craftyInstigator:
Is there any movement on this issue? It seems to be completely ignored for 3 months.

JLLeitschuh:
If you think you have a solution please feel free to submit a pull request.

TheBoef:
https://github.com/angular-ui/ui-grid/pull/5061

mportuga:
Resolved by the following PR:
https://github.com/angular-ui/ui-grid/pull/5061

PR #5061, opened 2016-02-03
ISSUE: 3782 (https://github.com/angular-ui/ui-grid/issues/3782)

TheBoef:
e can be null in Firefox. This gives the styles is null error.

Add this extra !== null check and the error is gone.

### DIFF
--- a/src/js/core/services/ui-grid-util.js
+++ b/src/js/core/services/ui-grid-util.js
@@ -910,7 +910,7 @@ module.service('gridUtil', ['$log', '$window', '$document', '$http', '$templateC
         e = elem[0];
       }
 
-      if (e) {
+      if (e && e !== null) {
         var styles = getStyles(e);
         return e.offsetWidth === 0 && rdisplayswap.test(styles.display) ?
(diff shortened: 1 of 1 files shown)


swalters:
I am not following this logic.  

var e=null;
If(e)  should return false.

TheBoef:
When you put ui-grid in an iFrame and have the plugin 'auto-resizing' enabled. The var e can be null when you 'hide' or remove the grid. 

I have an ui-grid that I open in a iframe modal that I can close/remove again. When I close the modal it gives the error "TypeError: styles is null"

I dont know exactly why e passes the `if (e)` check,  when its value is null. But adding this extra condition fixes the error for me.

mportuga:
@dlgski @JLLeitschuh This doesn't appear to be a breaking change for us, just counter intuitive, but it might close an issue. Should we merge it?

JLLeitschuh:
Sure. Can we add a comment that links back to this PR explaining why this is this way?
I mean, I think this makes sense.

PR #5084, opened 2016-02-09
chore(commits): Adds Commit Hooks to build

JLLeitschuh:
Commits will be validated before they can be created.
We now also support commitizen for commit message creation.


swalters:
Tested out locally.  Looks good.

PR #5094, opened 2016-02-10
fix #5090 introduce gridDimensionChanged event

### LINKED ISSUE #5090: resize the grid it is not adjusting the menu heights
When auto-resize is on the gridHeight is constantly getting checked and adjusted when a change to the height occurs.

The grid menus are calculated off of the gridHeight, in order to include a scrollbar on the menu so all items are reachable.

However, when autoResize changes the gridHeight, the menu height is not updated.

dlgski:
The grid menu enables a scrollbar to keep it's height under the height of the grid by 30px.  However, it is only calculated when created.  If auto-resize is on then the grid menu height does not get updated and therefore may get cut off.  This PR adds a new event gridDimensionChanged to the gridApi which the menu can watch for to update it's style on change.

### DIFF
--- a/src/features/auto-resize-grid/js/auto-resize.js
+++ b/src/features/auto-resize-grid/js/auto-resize.js
@@ -41,6 +41,7 @@
             if (newGridHeight !== prevGridHeight || newGridWidth !== prevGridWidth) {
               uiGridCtrl.grid.gridHeight = newGridHeight;
               uiGridCtrl.grid.gridWidth = newGridWidth;
+              uiGridCtrl.grid.api.core.raise.gridDimensionChanged(prevGridHeight, prevGridWidth, newGridHeight, newGridWidth);
 
               $scope.$apply(function () {
                 uiGridCtrl.grid.refresh()
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,7 @@ branches:
   only:
     - master
     - /^3\.0$/
+    - /^3\.x$/
     - /^release-.*$/
 
 before_install:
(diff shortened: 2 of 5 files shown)


src/features/auto-resize-grid/js/auto-resize.js
```diff
@@ -41,6 +41,7 @@
             if (newGridHeight !== prevGridHeight || newGridWidth !== prevGridWidth) {
               uiGridCtrl.grid.gridHeight = newGridHeight;
               uiGridCtrl.grid.gridWidth = newGridWidth;
+              uiGridCtrl.grid.api.core.raise.gridHeightChanged(prevGridHeight, newGridHeight);
```
swalters: this could be raised even if the gridWidth was the only thing that changed. I'd like to see a check here to only raise for height changes.

OR, you could change to gridDimensionChanged(prevGridHeight, newGridHeight, prevGridWidth, newGridWidth).  I like that even better.

dlgski: gridDimensionChanged does not exist already, correct?  I should just update the event I made here with this?

swalters: yeah, I would just update the event name.  

Are there other ways a gridDimension can get changed?   We need to make sure we raise the event everywhere it can occur.

dlgski:
@swalters good idea on gridDimensionChanged.  This has been updated.

swalters:
https://github.com/angular-ui/ui-grid/pull/5107

PR #5106, opened 2016-02-11
feat(core): Allow optional suppression of sort column menu items

dsjellz:


swalters:
How is this different from the colDef.enableSorting option?

dsjellz:
My understanding was that enableSorting would both hide the column menu items, and disable sorting on the column altogether. I don't actually want to turn sorting off, I just want to disable the column menu item so I can make my own title/icon, and add a couple lines to the action.  Since none of the default menu items have id's that I can see, removeMenuItem wasn't an option.

swalters:
Let me talk this over with the team.  We need to keep the public api consistent or we'll have a huge mess on our hands in the future.

dsjellz:
Sure thing. Happy to make modifications if that what needs to happen. Maybe a better approach would be to simply add id's to the default menu items so that .removeMenuItem could be used on those as well?

mportuga:
@dlgski Could you instead make the other change that you suggested and add ids to the default menu items? That way we can control this with CSS.

mportuga:
Closing due to innactivity

PR #5130, opened 2016-02-17
fix(exporter): pass column seperator options as param to downloadFile

dennym:
Write the correct csv header seperator. When `gridOptions.exporterCsvColumnSeparator` was set to something else than coma the header and the seperators are not equal and result in wrong behavior with excel.

In addition changed the ieVersion handling since it was broken due to #4640 for the csv part of the exporter.


PaulL1:
Looks good, merging.

dennym:
Oh dang it, just saw that I forgot to delete the coma from the document.write section on L934. In case when a custom seperator like `;` is set, the header would be `sep=,;`. That was stupid :facepalm:

PR #5142, opened 2016-02-19
fix(core): Incorrect horizontal scroll percentage calculation

korolyovvalentin:
Firstly it is mentioned in #4046.
The way that scroll length is calculated was changed: it's encapsulated into a function and invoked in adjustColumns and scrollHorizontal. Before it was calculated differently.

In adjustColumns scrollPercentage was equal 1,01... instead of 1 when it was fully scrolled.

### DIFF
--- a/src/js/core/factories/GridRenderContainer.js
+++ b/src/js/core/factories/GridRenderContainer.js
@@ -303,6 +303,10 @@ angular.module('ui.grid')
     return this.getCanvasHeight() - this.getViewportHeight() + this.grid.scrollbarHeight;
   };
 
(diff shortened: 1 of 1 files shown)


swalters:
Please squash your commits using rebase interactive and then force push back to your branch. thanks.

korolyovvalentin:
@swalters , sorry. Done.

swalters:
there are still 6 commits.  We want to see 1 commit with only the code this PR changed.

korolyovvalentin:
Terribly sorry. Squashed last two commits, but didn't drop others.

PR #5158, opened 2016-02-23
Fix for resize handler being called 500 times

### LINKED ISSUE #5099: Mistake in docs example
On page http://ui-grid.info/docs/#/tutorial/110_grid_in_modal you have wrong placed parameters 'delay' and 'count' in $interval call https://docs.angularjs.org/api/ng/service/$interval, so, you have resize handler called 500 times instead of 10.

trueblue2704:
Call to handleResizeWindow now does what the comment says it should do.

This fixes #5099.


swalters:
please squash your commits using  rebase -i and  force push back to your branch

PR #5160, opened 2016-02-23
Fix for resizeHandler() being called 500 times

trueblue2704:
- resizeHandler() now called 10 times, every 500 ms.
- Fixed comment to match tutorial description.


swalters:
Thanks!!

trueblue2704:
No problem. Just getting started with Angular, and ui-grid is pretty awesome!

PR #5174, opened 2016-02-28
feat(pagination): Add custom pagination with variable page sizes

chatcher:
Use Case:
 Client requests page breaks by (e.g.) Date column. This can currently be
accomplished by external pagination, but the 'page size' dropdown
becomes irrelevant, and the page-count on the right can no longer
correctly determine the correct row range.

The pagination template has been modified to hide the page-size dropdown
when custom pagination is enabled, and to use new API methods for the
row range.

A new tutorial page has been added as lesson 406-Custom-Pagination
which demonstrates two usages: pre-sorted data, with pre-calculated pages
sizes; and unsorted data, with external pagination.

### DIFF
--- a/src/features/pagination/js/pagination.js
+++ b/src/features/pagination/js/pagination.js
@@ -77,7 +77,7 @@
                       return index < grid.options.paginationCurrentPage - 1 ? result + size : result;
                     }, 0);
                   }
-                  return ((grid.options.paginationCurrentPage - 1) * grid.options.paginationPageSize) + 1;
+                  return ((grid.options.paginationCurrentPage - 1) * grid.options.paginationPageSize);
                 },
                 /**
                  * @ngdoc method
--- a/src/features/pagination/js/pagination.js
+++ b/src/features/pagination/js/pagination.js
@@ -87,9 +87,7 @@
                  */
                 getLastRowIndex: function () {
                   if (grid.options.useCustomPagination) {
-                    return grid.options.paginationPageSizes.reduce(function(result, size, index) {
-                      return index <= grid.options.paginationCurrentPage - 1 ? result + size : result;
-                    }, 0);
+                    return publicApi.methods.pagination.getFirstRowIndex() + grid.options.paginationPageSizes[grid.options.paginationCurrentPage - 1];
                   }
                   return Math.min(grid.options.paginationCurrentPage * grid.options.paginationPageSize, grid.options.totalItems);
                 },
--- a/src/features/pagination/js/pagination.js
+++ b/src/features/pagination/js/pagination.js
@@ -67,6 +67,34 @@
                 },
                 /**
                  * @ngdoc method
+                 * @name getFirstRowIndex
+                 * @methodOf ui.grid.pagination.api:PublicAPI
+                 * @description Returns the index of the first row of the current page.
+                 */
+                getFirstRowIndex: function () {
+                  if (grid.options.useCustomPagination) {
+                    return grid.options.paginationPageSizes.reduce(function(result, size, index) {
+                      return index < grid.options.paginationCurrentPage - 1 ? result + size : result;
+                    }, 0);
+                  }
+                  return ((grid.options.paginationCurrentPage - 1) * grid.options.paginationPageSize) + 1;
+                },
+                /**
+                 * @ngdoc method
+                 * @name getLastRowIndex
+                 * @methodOf ui.grid.pagination.api:PublicAPI
+                 * @description Returns the index of the last row of the current page.
+                 */
+                getLastRowIndex: function () {
+                  if (grid.options.useCustomPagination) {
+                    return grid.options.paginationPageSizes.reduce(function(result, size, index) {
+                      return index <= grid.options.paginationCurrentPage - 1 ? result + size : result;
+                    }, 0);
+                  }
+                  return Math.min(grid.options.paginationCurrentPage * grid.options.paginationPageSize, grid.options.totalItems);
+                },
+                /**
+                 * @ngdoc method
                  * @name getTotalPages
                  * @methodOf ui.grid.pagination.api:PublicAPI
                  * @description Returns the total number of pages
@@ -74,6 +102,10 @@
                 getTotalPages: function () {
                   if (!grid.options.enablePagination) {
                     return null;
+                  }
+
+                  if (grid.options.useCustomPagination) {
+                    return grid.options.paginationPageSizes.length;
                   }
 
                   return (grid.options.totalItems === 0) ? 1 : Math.ceil(grid.options.totalItems / grid.options.paginationPageSize);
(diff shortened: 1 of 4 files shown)


chatcher:
hmm, this passes locally...

AgDude:
This looks like a very specific use case. I think I would prefer to see this in an independent plugin (or all of the pagination UI in a plugin for that matter).

mportuga:
@chatcher Can you fix unit tests?

src/features/pagination/js/pagination.js
```diff
@@ -146,12 +176,14 @@
             var visibleRows = renderableRows.filter(function (row) { return row.visible; });
             grid.options.totalItems = visibleRows.length;
 
-            var firstRow = (currentPage - 1) * pageSize;
+            var firstRow = publicApi.methods.pagination.getFirstRowIndex();
+            var lastRow  = publicApi.methods.pagination.getLastRowIndex();
```
mportuga: These definitions can be comma separated.

src/features/pagination/js/pagination.js
```diff
@@ -146,12 +176,14 @@
...
+
             if (firstRow > visibleRows.length) {
               currentPage = grid.options.paginationCurrentPage = 1;
               firstRow = (currentPage - 1) * pageSize;
             }
-            return visibleRows.slice(firstRow, firstRow + pageSize);
+            return visibleRows.slice(firstRow, lastRow);
```
mportuga: Your function for getLastRowIndex doesn't appear to match the firstRow + pageSize calculation that I see here. Am I missing something?

src/features/pagination/js/pagination.js
```diff
@@ -67,6 +67,32 @@
...
+                 * @ngdoc method
+                 * @name getLastRowIndex
+                 * @methodOf ui.grid.pagination.api:PublicAPI
+                 * @description Returns the index of the last row of the current page.
+                 */
+                getLastRowIndex: function () {
+                  if (grid.options.useCustomPagination) {
+                    return publicApi.methods.pagination.getFirstRowIndex() + grid.options.paginationPageSizes[grid.options.paginationCurrentPage - 1];
+                  }
+                  return Math.min(grid.options.paginationCurrentPage * grid.options.paginationPageSize, grid.options.totalItems);
```
mportuga: Why is Math.min being used here? That doesn't appear to be there in the original function for when useCustomPagination is false. Shouldn't this be the following?
`publicApi.methods.pagination.getFirstRowIndex() + grid.options.paginationPageSize `

src/features/pagination/js/pagination.js
```diff
@@ -368,8 +410,6 @@
           $scope.$on('$destroy', dataChangeDereg);
 
           var setShowing = function () {
```
mportuga: It looks like this is an empty function now. Was that on purpose? If so, what is the intention here?

src/features/pagination/test/pagination.spec.js
```diff
@@ -64,6 +64,8 @@ describe('ui.grid.pagination uiGridPaginationService', function () {
   describe('initialisation', function () {
     it('registers the API and methods', function () {
       expect(gridApi.pagination.getPage).toEqual(jasmine.any(Function));
+      expect(gridApi.pagination.getFirstRowIndex).toEqual(jasmine.any(Function));
```
mportuga: You should add more unit tests for these two functions than simply checking if they exist and are functions. Would you mind adding tests to ensure they return what is expected?

src/features/pagination/js/pagination.js
```diff
@@ -368,8 +410,6 @@
...
-            $scope.showingLow = ((options.paginationCurrentPage - 1) * options.paginationPageSize) + 1;
-            $scope.showingHigh = Math.min(options.paginationCurrentPage * options.paginationPageSize, options.totalItems);
```
mportuga: Are showingHight and last page index really the same thing? If so, how did you come to that conclusion?

src/features/pagination/js/pagination.js
chatcher: This was taken from `$scope.showingHigh` (line 372), which is replaced by `1+getLastRowIndex()`.

src/features/pagination/js/pagination.js
chatcher: `firstRow` is `(currentPage - 1) * pageSize`
Previously, `lastRow` was `firstRow + pageSize`, which is algebraically
```
lR = fR + pS          | fR = (cP - 1) * pS
   = (cP-1) * pS + pS | distribute pS over (cP-1)
   = cP*pS - pS + pS  | cancel addition
   = cP*pS
```
With the change, `lastRow` is `currentPage * pageSize` (or `totalItems`), which simply eliminates the math in the middle.

src/features/pagination/js/pagination.js
chatcher: That was in the original code; I could not find a usage or reassignment of it, but I didn't want to touch it in case I was wrong.

src/features/pagination/js/pagination.js
chatcher: The usage of `showingLow` and `showingHigh` was in the template to show the row range.
Instead of calculating the range here, I instead call the new methods from the template and add one there.

src/features/pagination/js/pagination.js
chatcher: like so?
```javascript
var firstRow = blahBlah(),
    lastRow  = blahBlah();
```
I don't see that pattern used elsewhere. Can you clarify what you mean?

mportuga: that is what I meant. But I suppose you are right and we are not using this anywhere else in this project right now. It is my personal preference. No need to change it, in that case.

src/features/pagination/js/pagination.js
mportuga: Thanks for the lesson in algebra right now. I should have caught that.

src/features/pagination/js/pagination.js
mportuga: set showing is called in lines [415](https://github.com/chatcher/ui-grid/blob/29fdb7cd485607e0c7e579df82880b4aceae0d35/src/features/pagination/js/pagination.js#L415) and [432](https://github.com/chatcher/ui-grid/blob/29fdb7cd485607e0c7e579df82880b4aceae0d35/src/features/pagination/js/pagination.js#L432). One of those usages also creates a watch, I would argue that if this function can definitely stay empty that you should remove this function, the watch on line 415 and the usage on line 432. Also, the call to destroy said watch on line [438](https://github.com/chatcher/ui-grid/blob/29fdb7cd485607e0c7e579df82880b4aceae0d35/src/features/pagination/js/pagination.js#L438)

src/features/pagination/test/pagination.spec.js
chatcher: I've added tests for custom pagination.

Rather than testing `getFirstRowIndex` and `getLastRowIndex` directly, we test the resulting pages.

The previous tests covered one branch inside the two methods, and the tests I've added cover the other branch.

Issue #5182, opened 2016-03-03
[infinite scroll] Data sometimes does not load when scrolling to top/bottom of grid too quickly


lbntly:
Sometimes if the user is scrolling quickly all the way to the top or bottom, the scroll handler can get confused about the direction and thus not load more data. This is because the scroll handler is not called until the scroll is *over*, and at that point they are no longer actively scrolling in that direction because they made it all the way to the bottom/top.

If you look at the example (here)[http://ui-grid.info/docs/#/tutorial/212_infinite_scroll] you can sometimes reproduce it.

nkoterba:
👍 Have encountered this issue frequently.  I noticed you submitted a MR to fix this...was it ever merged in?

Have you just forked your own copy of `ui-grid` to proceed?

lbntly:
@nkoterba Here's the PR: https://github.com/angular-ui/ui-grid/pull/5183 Nope, it hasn't been merged. Yes, I ended up forking & publishing to a private npm registry since I needed it for a time-sensitive feature. It would be great to get it merged, though -- I'll try to contact the ui-grid owners.

PR #5183, opened 2016-03-03
[infinite scroll] make sure more data is always loaded if scrolled to top/bottom quickly

lbntly:
Resolves issue https://github.com/angular-ui/ui-grid/issues/5182


lbntly:
@swalters I noticed someone else experiencing this issue -- would be great to get this merged in!

PR #5210, opened 2016-03-09
fix(batch): Prevent batch event from firing for single row

twh1808:
Prevent `rowSelectionChangedBatch` from firing when a single row is changing. As seen in this [plunkr](http://plnkr.co/edit/Dtdm88XJtXg9QA8dBkyq?p=preview), when changing from one row to another we get an unnecessary batch log in the Developer tools. Batch event(s) should only fire when `changedRows.length > 1`.

### DIFF
--- a/src/features/selection/js/selection.js
+++ b/src/features/selection/js/selection.js
@@ -600,7 +600,7 @@
          * if we're doing batch events
          */
         decideRaiseSelectionBatchEvent: function( grid, changedRows, evt ){
-          if ( changedRows.length > 0 ){
+          if ( changedRows.length > 1 ){
             grid.api.selection.raise.rowSelectionChangedBatch(changedRows, evt);
           }
         }


AgDude:
This has potential to be a breaking change for a lot of applications. While I can agree with the semantics you describe, I don't feel it is worth the breaking change. Your handler can check the number of rows.

twh1808:
That's exactly what semantic versioning is for. If you believe something will create a breaking change, you document it in a CHANGELOG and increment the appropriate version number as appropriate. Just leaving it broken and telling people to add length checks to their code isn't a solution.

viveleroi:
Wouldn't it be worth merging into a bleeding branch or something, so that it can be released bundled with any other breaking changes? I'd much rather see a library evolve toward being consistent and behaving as the average developer would expect, not leaving in problematic functionality simply to avoid a future, inevitable release with breaking changes.

AgDude:
Implementing any breaking change is a cost benefit analysis, regardless of versioning, and my opinion is that this benefit to this change is minimal. Clearly we wouldn't intentionally put a breaking change in without a major version bump, but even with a version bump, the fewer breaking changes the better.

If you feel strongly about this bring it up on gitter. I haven't carefully reviewed what events are already present on selection, but I will insist on an which fires for 1 or more rows selected, as the "batch" event does now. Perhaps this one was poorly named, but I think that was the original intent, and it is how I have seen it used in many cases.

twh1808:
I've never been a fan of the UI-Grid events, as I felt they were lacking, not written to common standards, or very friendly to use. For example, there's no `beforerowschanged` event and no ability to prevent row selection from changing (which is I feature I had to hack together).

With that rant said, here's what my code looks like:
```
var onRowSelectionChanged = function() {
    var rows = api.selection.getSelectedGridRows();

    ...
};

api.selection.on.rowSelectionChanged($scope, onRowSelectionChanged);
api.selection.on.rowSelectionChangedBatch($scope, function(rows) {
    // A batch event fires that we don't want, so let's catch and throw it out
    if (!_(rows).filter({ isSelected: true }).isEmpty()) {
        onRowSelectionChanged();
    }
});
```
It'd be nice if I didn't need the extra check for every grid I have, which is a lot and growing, but yes you're right it's a cost benefit evaluation (although I find it next to no cost). I suppose I will ignore this and just keep an eye out to see if your team changes their mind.

Thanks,
Troy

PaulL1:
I think you're looking for different semantics than I had in mind when I wrote it.

My intent was that you could just listen to the batch event, rather than listening to both the single event and the batch event.  So that makes sense to me.  

You appear to have different processing for single and for batch.  So you want different semantics.

The interface you've suggested can't support both uses, and unfortunately the current use woudl take priority.  You could add a switch, then it wouldn't be a breaking change.

twh1808:
From what I remember testing, listening on batch didn't work 100% either but I may be mistaken since I tested/developed this a long time ago. I can't remember the exact details, but I can try digging them up again in a plunkr if that'll help.

viveleroi:
There shouldn't be two events. I think you'd agree because you wanted people to be able to use only one yet this problem arises when people need to listen to both for unrelated reasons. I'd *only* use the batch event but it doesn't cover some cases that the singular event does.

Both events are fired at the grid-level (I don't believe ui-grid fires events on any other levels) so it's not intuitive why they're separate. I feel that a single event can easily tell when the selection changes while covering all cases.

I'd only feel a second event makes sense if every row instance fired its own `selectionStateChanged` event. It would clearly be singular (for that specific row), and the grid itself would have only one event.

twh1808:
There were two key scenarios that I can remember that required me to listen to both events. When these scenarios were the case, I was forced to listen to both events (and throw out the extra one).

- No rows selected and select a single row results in no batch event
- Select/deselect any row (no matter the current selection) using CTRL results in no batch event

PR #5218, opened 2016-03-11
Added simple clearExistingGridSortsOnColumnHeaderMenuSort option to c…

rmechamatadobe:
…olumn defs to allow for default column menu sorting to reset any active sort set before the new sort is applied.

Also added documentation and example.

### DIFF
--- a/src/js/core/factories/GridColumn.js
+++ b/src/js/core/factories/GridColumn.js
@@ -81,6 +81,23 @@ angular.module('ui.grid')
    */
 
   /**
+   * @ngdoc property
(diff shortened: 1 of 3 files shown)


src/js/core/factories/GridColumn.js
```diff
@@ -81,6 +81,23 @@ angular.module('ui.grid')
    */
 
   /**
+   * @ngdoc property
```
mportuga: It looks like you added the documentation twice here. Please remove the one without the example.

mportuga:
Closing due to inactivity.

PR #5226, opened 2016-03-14
Workaround to solve issue with enable___ (enableRowSelection, ...)

neolanders:
When trying to set enable(RowSelection) in gridOptions the row still remain selectable.

$scope.gridOptions = {
    enableRowSelection: **false**,
    enableSelectAll: true,
    selectionRowHeaderWidth: 35,
    rowHeight: 35,
    showGridFooter:true
  };

When setting ui-grid-selection here the parameter enableRowSelection have no effect to the grid whatever is value (true / false)

<div ui-grid="gridOptions" ui-grid-selection class="grid"></div>

I've made a workaround to solve this issue here:
https://github.com/neolanders/angular-ui-grid-dynamic-wrapper

### DIFF
--- a/misc/tutorial/101_intro.ngdoc
+++ b/misc/tutorial/101_intro.ngdoc
@@ -34,7 +34,10 @@ Steps:
    </pre>
  </li>
  <li>
-   Add a css style to your app css so the grid knows it's dimensions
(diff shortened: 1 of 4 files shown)


mportuga:
@neolanders Can you add unit tests for your change?

mportuga:
Closing due to inactivity.

Issue #4541, opened 2015-10-19
Detect keyDown on grid


Havock94:
Hi,
I would like to detect a keydown event on the grid.
So far I've found the cellNav.on.viewPortKeyDown event but unfortunately it doesn't work with arrow keys/enter/esc and tab (default navigation buttons).

Is there a way to capture those keys too??
Thank you!

dregan-gh:
We also had this requirement, the only way I could find to handle it was via global keyboard bindings since there isn't a 'html control' element to bind to.

fallXone:
I've created a pull request to resolve this issue that will allow you to override any of the default events that the grid is handling.

mportuga:
Closed by https://github.com/angular-ui/ui-grid/pull/5248

PR #5248, opened 2016-03-23
Option to override cell navigation defaults

fallXone:
Option to override default cell navigation

keyDownOverrides will force the viewPortKeyDown event
to be triggered if the specified key event was pressed
down, even if it is a default key command for the grid.
The handling of the event should then be done in the
viewPortKeyDown handler since the grid is no longer
handling the event.

This fixes issue: https://github.com/angular-ui/ui-grid/issues/4541

### DIFF
--- a/misc/tutorial/202_cellnav.ngdoc
+++ b/misc/tutorial/202_cellnav.ngdoc
@@ -28,6 +28,8 @@ Provides an example of scrolling to and setting the focus on a specific cell, us
 Provides an example of utilizing the modifierKeysToMultiSelectCells option and getCurrentSelection API function to
 extract values of selected cells.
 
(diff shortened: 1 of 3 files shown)


JLLeitschuh:
This looks good and it is tested. Awesome! I want to run this by someone else first.
@swalters you want to take a look?

AgDude:
This looks fine to me. I think it belongs in a minor, not a patch release.

aih:
This is a great addition. The override does not appear to work within a cell-edit input cell. @fallXone, @mportuga, Is there a way to override the keys from within cell being edited as well?

Backi03:
Hi, nice addition. When will this one going to release?

Issue #5254, opened 2016-03-24
Column Reorder:  hidden columns triggering unnecessary re-order


dlgski:
If your grid has hidden columns and you try to  move columns around.  The hidden columns are triggering a change even when none is visibly occurring

Check out this Plunker:  http://plnkr.co/edit/DoPGDWGV4sxuKfaTt85k?p=preview

There are 2  hidden columns between name and age.  Move the age column around, but then place it back where it currently is.  So, visibly, it did not move.  However, due to the hidden columns, ui-grid picks up a move from column 4 to column 2 and will trigger the columnMove events.

PR #5255, opened 2016-03-24
fix(col-movable): prevent hidden columns triggering unnecessary re-or…

dlgski:
This fix checks all the columns between the originalPosition and newPosition.  IF they are all hidden, then no re-order event is necessary.  If a visible column is found, column re-ordering will be triggered.

Fixes issue: #5254


dlgski:
I do not understand what check this PR failed.  Can someone advise?

JLLeitschuh:
Adding some comments to this would be really helpful.
Explaining why that bit of code is there, what its intentions are and how it works.

PR #5300, opened 2016-04-08
Update exporter.js

Havock94:
Added separator parameter documentation in the formatAsCsv function


dennym:
+1 since I forgot that completely :)

Issue #4514, opened 2015-10-14
headerCellTemplate not properly resolved when providing a URL


SoundLogic:
I've created a custom `headerCellTemplate` which is very much like the regular `headerCellTemplate`, but I wanted the column filters to display above the column titles as I need the column header text to wrap per business requirements. This all works great, but I found when I was providing a URL for the `headerCellTemplate`, my header wasn't rendering. I looked at the `gridUtil.getTemplate()` function, but this appears to be resolving the template as expected so I imagine something is somehow happening out of band (maybe a side effect of one of the features I'm using?). I was able to get around this problem by setting my `headerCellTemplate`s to a promise returned from the following helper function:

``` javascript
function getTemplatePromise(templateUrl) {
    return $http.get(angular.isFunction(templateUrl) ? (templateUrl)() : templateUrl, { cache: $templateCache }).then(function (result) { return result.data; });
}
```

here is my configuration and features I am using:

``` javascript
    //configuration
    $scope.vm.gridOptions = {
            enableGridMenu: true,
            enableFiltering: true,
            enablePaginationControls: true,
            paginationPageSizes: [25, 50, 75, 100],
            paginationPageSize: 50,
            enableSorting: true,
            data: 'vm.myData',
            onRegisterApi: function(gridApi) {
                $scope.vm.gridApi = gridApi;
            }
    }
```

``` html
        <!-- features -->
        <div ui-grid="vm.gridOptions"
             ui-grid-auto-resize
             ui-grid-resize-columns
             ui-grid-pagination>
        </div>
```

anyone have any insight into what might be causing this issue?

SoundLogic:
quick update on this as I added more custom templates I was still running into the issue so I had to initialize the `gridOptions` after a `$q.all()` to retrieve all templates and then used an `ng-if` to ensure the grid would not be rendered until the `gridOptions` were set

Edifear:
the problem was in promise template type condition in `getTemplate` funtion,
PR created:
https://github.com/angular-ui/ui-grid/pull/5311

david-hanger:
Since I recently had a similar problem trying to use a URL value for **filterHeaderTemplate** (and getting empty content), I posted a [comment with a recommended fix](https://github.com/angular-ui/ui-grid/pull/5311#issuecomment-324158897) to the pull request #5311.  I thought I'd reference that comment here, in case anyone else comes across this before a better fix is applied.

[...]

PR #5311, opened 2016-04-12
fix(getTemplate): Updated custom templates as promises condition

Edifear:
now custom promise templates will handled properly, e.g.:
`headerCellTemplate: $http.get('./template.html').then(function(res) {return res.data;})`

fixes issue https://github.com/angular-ui/ui-grid/issues/4514

### DIFF
--- a/src/js/core/services/ui-grid-util.js
+++ b/src/js/core/services/ui-grid-util.js
@@ -351,7 +351,7 @@ module.service('gridUtil', ['$log', '$window', '$document', '$http', '$templateC
       }
 
       // See if the template is itself a promise
-      if (template.hasOwnProperty('then')) {
+      if (angular.isFunction(template.then)) {
         return template.then(s.postProcessTemplate);
       }
 


david-hanger:
The referenced fix for issues #4514  and #5717 works for the above example of example, but it is not quite a complete fix.  You are correct that headerCellTemplate now handles a promise value, but it still does not correctly handle a URL as the value.  The problem is that the **uiGridHeaderCell.compile.pre** presumes that **$scope.col.headerCellTemplate** is already set, but **gridClassFactory.defaultColumnBuilder** leaves **col[templateType]** _undefined_ until **after** resolving the GET request on the URL.  The reason the above example works is because headerCellTemplate is immediately assigned the promise as a value.  If you change it to this:
`headerCellTemplate: './template.html'`
... then the **headerCellTemplate** is treated as empty - because, again, it is _undefined_ when read (unless it happens to get assigned before being read, which is unlikely).

Please consider updating this fix to include something akin to the following fix I made for the same issue with **filterHeaderTemplate** (which can trivially be applied to **headerCellTemplate** and any other templates whose value is read without dealing with promises).  I'm basing this off the [current master, commit 84ea8fa](https://github.com/angular-ui/ui-grid/commit/84ea8fa8db9c9979ee46452a773ae7da028d0dc2).
(sorry; my fixes are on a network without internet access, so I can't provide a proper diff; I'm also attempting to format similarly to the adjacent code).

**src/js/core/services/gridClassFactory.js**

replace line **101**:
```
             templateGetPromises.push(gridUtil.getTemplate(col[providedType])
```
... with:
```
             var promise;
             // reuse existing promise
             if ( col[templateType] && angular.isFunction(col[templateType].then) ){
                 promise = col[templateType];
             }
             // likewise if there is a related promise, e.g. cellTemplatePromise
             if ( col[templateType + 'Promise'] ){
                 promise = col[templateType + 'Promise'];
             }
             else {
                 promise = gridUtil.getTemplate(col[providedType])
```

Add this after line **118** (following the `}` after `col[templateType] = template;`):
```
                   // resolve the final value for this promise chain
                   return col[templateType];
```

replace lines **122-123**:
```
                }).catch(angular.noop)
            );
```
... with:
```
                }).catch(angular.noop);
             }
             // add the promise to the list (either new or reused prior promise)
             templateGetPromises.push(promise);
             // assign the promise as an intermediate template value, until resolved
             if ( !col[templateType] ){
               col[templateType] = promise;
             }
```

**src/js/core/directives/ui-grid-filter.js**

replace line **4**:
```
  angular.module('ui.grid').directive('uiGridFilter', ['$compile', '$templateCache', 'i18nService', 'gridUtil', function ($compile, $templateCache, i18nService, gridUtil) {
```
... with (adding $q):
```
  angular.module('ui.grid').directive('uiGridFilter', ['$compile', '$templateCache', '$q', 'i18nService', 'gridUtil', function ($compile, $templateCache, $q, i18nService, gridUtil) {
```

replace lines **13-15**:
```
                var template = $scope.col.filterHeaderTemplate;

                $elm.append($compile(template)($scope));
```
... with:
```
                // handle promise values
                $q.when( $scope.col.filterHeaderTemplate ).then( function( template ){
                  $elm.append($compile(template)($scope));
                });
```

PR #5331, opened 2016-04-19
Fix javascript file reference format error in README.md

AsCat:
Hi, I fixed a error of missing the end part of javascript reference tag in README.md, which is not allowed in html standard syntax : )


AsCat:
Does anyone know why the CI could not complete?

Issue #4653, opened 2015-11-06
Sort number on header displaying wrong value


bjdash:
When I sort a column by clicking on headers 1 is displayed to indicate sort order. Shift clicking on another column and 2 on that. Now if I continue to  shift click on first column that I sorted earlier (marked with label 1), after one click 1 is removed and after another click it is labeled with 3. Now I have two columns labeled 2 and 3 while 1 is missing. 

One would expect that it should retain the number or decrease the other numbers by 1 before adding a label.

imbalind:
Can you please make a plunkr that shows this behaviour?

bjdash:
Here it is http://plnkr.co/edit/i4HRAaLr9RhkwWM5ZTWc?p=preview.
I am using the same example from the ui-grid tutorial. 

To replicate the issue sort the table by 'Gender', now shift click on "Name" to sort it. At this stage 'Gender' is tagged 1 and 'Name' is tagged 2. Now try to sort "Gender" again but do it by holding Shift. After few clicks "Gender" will be tagged 3 and there is no 1.

Also if you continue Shift clicking 'Name' it will get 4.

imbalind:
The problem seems to be function getNextColumnSortPriority at [Grid.js](https://github.com/angular-ui/ui-grid/blob/e89598a588791a6fc18a040c00a805dc1b6c5d25/src/js/core/factories/Grid.js#L1874).

This function now looks for the greatest column priority and then returns that value + 1.

A better version would find the least unused priority.

A further improvement would be to edit [sortColumn](https://github.com/angular-ui/ui-grid/blob/e89598a588791a6fc18a040c00a805dc1b6c5d25/src/js/core/factories/Grid.js#L1944) function and lower the priority af any previously prioritized column whose value was higher then the one the user edited.

bjdash:
Is there any word when this is will be fixed, beacuse I am using it and I have a defect logged for this.

mattmutt:
problem still exists -- even on the UI Grid home page

on the basic example grid . 
Use the arrow menu on the "Name" column Set it to Sort Ascending

 
Use the arrow menu on the "Gender" column Set it to Sort Descending


Use the arrow menu on the "Name" column Set it to Sort Descending.
Notice the count is incremented incorrectly. It is easy to really mess up the internal accounting by playing with the sort directions on multiple columns


hint : if you have only one number to work with , then refactor the code to use a sign ( + or - ) and the priority to designate sort and priority.

[...]

Issue #4196, opened 2015-08-14
Displaying sort priority: doesn't renumber after column menu -> remove sort


meyerds:
I am using a custom header cell template that displays the sort priority for each column to the user. It works well, except when there are 2 or more columns sorted and a higher-priority column's sort is removed. The sort priorities do not update to reflect the removed sort.

Plunker demo: http://plnkr.co/edit/49GWnKBVRZuWqM0CJ6Fb?p=preview

PR #5343, opened 2016-04-21
Fix: bug wrong sort priorities #4653 and #4196

jgrasl:
#4653: sort Priority 1 is presented as 0 in the sort object. It shows true for !column.sort.priority so it should be checked with column.sort.priority === undefined.
#4196: add function to decrease higher priorities if a column sort with lower priority is removed.

Added test cases to verify behavior.

### DIFF
--- a/src/js/core/factories/Grid.js
+++ b/src/js/core/factories/Grid.js
@@ -1979,7 +1979,7 @@ angular.module('ui.grid')
       // Get the actual priority since there may be columns which have suppressRemoveSort set
       column.sort.priority = self.getNextColumnSortPriority();
     }
-    else if (!column.sort.priority){
+    else if (column.sort.priority === undefined){
       column.sort.priority = self.getNextColumnSortPriority();
     }
 
@@ -1997,7 +1997,7 @@ angular.module('ui.grid')
       if (column.sortDirectionCycle[i]) {
         column.sort.direction = column.sortDirectionCycle[i];
       } else {
-        column.sort = {};
+        removeSortOfColumn(column, self);
       }
     }
     else {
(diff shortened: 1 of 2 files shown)


src/js/core/factories/Grid.js
```diff
@@ -1899,6 +1899,20 @@ angular.module('ui.grid')
     return p;
   };
 
+  Grid.prototype.removeSortOfColumn = function removeSortOfColumn(column) {
```
JLLeitschuh: Is there a reason that this is being added as a public method??

mthiruvengadam:
Please let me know if this bug is fixed in the latest code. I have a defect logged for this item.

src/js/core/factories/Grid.js
jgrasl: No I'm gonna change it not to be part of the prototype.

jgrasl: I updated the pull request with the private function.

AgDude:
Thanks for the fix.

nguyenpc:
This bug still persists in v4.0.2. 
Assume that I have three columns or more. first_name, last_name and age,...
1. Firstly, I will choose sorting by two collumns E.g. first_name (priority 1) and age (priority 2).  
2. Then I choose to remove sorting by first_name. the grid now is sorting by column Age. 
3. I choose to add first_name to sort list (shift+click on first_name column)

Then I get: column Age now has priority 2, column first_name now has priority 3.

jgrasl:
@nguyenpc Unfortunately sort priority change doesn't run on the action Remove Sort only if you change the priority with shift click again. I'll take a look and fix it to rearrange priority on the Remove Sort function too.

PR #5361, opened 2016-04-26
fix(core): Do not clear condition when clearing all filters

### LINKED ISSUE #4657: "Clear all filters" on gridMenu wipes out filter conditions
"Clear all filters" on gridMenu wipes out filter conditions as shown in the test:

http://plnkr.co/edit/7O9OkqFHoVmzHx3OiAH2

The conditions are printed to console when typing in a filter field.  Can I remove "Clear all filters" on gridMenu (as a workaround)?

lebolo:
**Current Behavior**: Grid menu option _Clear all filters_ removes all filter `condition`s as well as search `term`s.

**Problem**: Users expect _Clear all filters_ to only remove search `term`s, but leave `condition`s as is so that filters can still be used to search the table. Essentially, we expect _Clear all fitlers_ to be the same as going through each filter and clicking the small _x_ next to it.

**Fix**: _Clear all filters_ only removes search `term`s and leaves `condition`s as is.

Fixes #4657

### DIFF
--- a/src/js/core/directives/ui-grid-menu-button.js
+++ b/src/js/core/directives/ui-grid-menu-button.js
@@ -193,7 +193,7 @@ angular.module('ui.grid')
       var clearFilters = [{
         title: i18nService.getSafeText('gridMenu.clearAllFilters'),
         action: function ($event) {
-          $scope.grid.clearAllFilters(undefined, true, undefined);
+          $scope.grid.clearAllFilters();
         },
         shown: function() {
           return $scope.grid.options.enableFiltering;


JLLeitschuh:
I'm not sure I understand what/why this change is necessary. What is the behaviour before without this fix? What is it now??

lebolo:
Sorry about that, I thought it was explicit enough to say that "clear all filters" should not remove the filter's `condition`, i.e. it should only remove the `term`.

More details are in the linked issue, particularly https://github.com/angular-ui/ui-grid/issues/4657#issuecomment-154906647

drGarbinsky:
Currently hitting this bug. Anything can to do help move this forward? I've tested the change proposed and all looks good.

lebolo:
I believe the majority of ui-grid developers have moved on or are really busy. They're currently looking to fill the the void, but I don't think we'll be seeing a release anytime soon :(

AgDude:
I am in favor of merging this, but it will be a breaking change. @dlgski

el-davo:
Hi All, Any updates on when this is likely to be merged, or is it going to be merged?

mportuga:
@drGarbinsky We will likely include it in the next big release.

PR #5370, opened 2016-04-29
Update to ngdocs in selection.js

alimavrakis:
Simple typo in the docs if i'm not mistaken.


AgDude:
This docs appear correct as is to me. Please open an issue with a plunker to demonstrate if you still think they are incorrect.

PR #5378, opened 2016-05-05
perf(core): debounce scrollHandler in uiGridHeader and uiGridViewport

lootek:
Use scrollDebounce param from global ui-grid options.


JLLeitschuh:
What is the behaviour before and what is the behaviour now? How has it changed?
Can we get more details on this?

lootek:
Before, `scrollHandler` was getting called on every scroll event (every pixel/few pixels).
After wrapping it with `debounce` util, it gets called once, just after end of scrolling. Just as it's stated in the docs for debounce: 
> [debounce] Takes a function, decorates it to execute only 1 time after multiple calls

This has major performance impact, noticeable especially for the case of `ui-grid-viewport.js` where the `scrollHandler` function reads `scrollTop` DOM property (which forces immediate page reflow).

You may be interested in Google results for [javascript debounce scroll event](https://www.google.com/search?q=javascript%20debounce%20scroll%20event)

mportuga:
@lootek Can you resolve merge conflicts?

JLLeitschuh:
Can you please rebase your changes instead of merging up?

mportuga:
Closing due to inactivity.

PR #5389, opened 2016-05-06
removed log debug statement

Treysin:
There was a debug log statement in ui-grid-menu.js that was causing a log in the browser console any time the method itemAction was called. This has been removed.


prettycode:
+1

PR #5404, opened 2016-05-16
fix(cellNav): notifyText incorrect if cellFilter had string literal parameter

MikeMatusz:
GridRowColumn had an alternate implementation for getting the filtered value of a cell, possibly brought forward from ng-grid, but in certain situations, particular when handling cellFilters with parameters that were string literals, it would not always produce the correct output.  Grid already has a getCellDisplayValue with works perfectly (by storing the actual value, not attempting to duplicate all of angular's filter processing), and the single call in cellNav was the only place getIntersectionValueFiltered was being used.  Maintaining two functions that do the same thing in different ways, particularly when one is an attempted re-implementation of internal angular logic seems dangerous, so I'm proposing it be removed altogether in favor of using getCellDisplayValue.

### DIFF
--- a/src/features/cellnav/js/cellnav.js
+++ b/src/features/cellnav/js/cellnav.js
@@ -815,7 +815,7 @@
                   var values = [];
                   var currentSelection = grid.api.cellNav.getCurrentSelection();
                   for (var i = 0; i < currentSelection.length; i++) {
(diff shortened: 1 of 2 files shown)


JLLeitschuh:
Yep. This makes sense. I didn't realize that other method existed.
Thank you!

PR #5421, opened 2016-05-20
Permit to disable watchers on filters

jibap:
Add an option (disableFilteringWatchers) to permit to disable watchers on filters which can slow behaviour and not necessarly needed (external filtering and/or custom filter templates)

### DIFF
--- a/src/js/core/directives/ui-grid-header-cell.js
+++ b/src/js/core/directives/ui-grid-header-cell.js
@@ -257,7 +257,7 @@
                 }
 
                 // if column is filterable add a filter watcher
-                if ($scope.filterable) {
+                if ($scope.filterable && !uiGridCtrl.grid.options.disableFilteringWatchers) {
                   $scope.col.filters.forEach( function(filter, i) {
                     filterDeregisters.push($scope.$watch('col.filters[' + i + '].term', function(n, o) {
                       if (n !== o) {


src/js/core/directives/ui-grid-header-cell.js
```diff
@@ -257,7 +257,7 @@
                 }
 
                 // if column is filterable add a filter watcher
-                if ($scope.filterable) {
+                if ($scope.filterable && !uiGridCtrl.grid.options.disableFilteringWatchers) {
```
JLLeitschuh: @swalters or @AgDude You guys have a thought on this?

AgDude:
I would like to see a benchmark on the improvement before considering this. A single watcher per column seems like any performance improvement would be tiny.

jibap:
Performance is slowed not by watcher but by the callback of this one : 
```
uiGridCtrl.grid.api.core.raise.filterChanged();
uiGridCtrl.grid.api.core.notifyDataChange( uiGridConstants.dataChange.COLUMN );
uiGridCtrl.grid.queueGridRefresh();
```

The visible effect is that the cursor into input is frozen for some milliseconds (seconds if the grid is big or complex)

AgDude:
That sounds reasonable then. Can you please provide a plunker, I have not observed this.

At the very least this will need documentation.

I assume then it is left up to the developer to refresh the grid after the filter input is complete. Might it be better to build in the refresh on blur instead of removing it completely?

mportuga:
Closing due to inactivity.

Issue #5326, opened 2016-04-15
Sorting is wrong when flatEntityAccess and sortCellFiltered are enabled


lebolo:
If `flatEntityAccess: true` and `sortCellFiltered: true` are set for a column definition, the grid doesn't sort.

In this case, [`rowSorter`](https://github.com/angular-ui/ui-grid/blob/v3.1.1/src/js/core/services/rowSorter.js#L450) uses `grid.getCellDisplayValue` instead of `grid.getCellValue` and the first cell value encountered gets cached*. Meaning all subsequent cell values get returned as the first cell value (so the grid ultimately fails to sort since it thinks all cell values are the same).

Here is a [plunkr](http://plnkr.co/edit/8LWQTs?p=preview) - try to sort the _gender_ column to see the problem.

I can disable `flatEntityAccess` as a workaround, but am not sure how to properly fix this. If anyone has some guidance, I can try creating a pull request.

*This happens because [`getCellDisplayValue`](https://github.com/angular-ui/ui-grid/blob/v3.1.1/src/js/core/factories/Grid.js#L1857) uses `$parse(row.entity[col.field] + custom_filter)`, which becomes `$parse("first_cell_value_encountered | some_filter")` when `flatEntityAccess` is enabled, instead of `$parse(row.getEntityQualifiedColField(col) + custom_filter)`, which becomes `$parse("entity['some_field'] | some_filter")`

lebolo:
I originally thought this only happened when `cellFilter` was not set, but that's not the case. I edited the title and original issue.

lebolo:
Obviously, this also affects situations where `flatEntityAccess` and `filterCellFiltered`are both enabled.

jasonlikescats:
Here's a somewhat more direct [reproduction plunkr](http://plnkr.co/edit/TPoyO0ktcwFYfDoNlH7Z?p=preview) that just calls getCellDisplayValue directly. Click the button to trigger the error.

PR #5426, opened 2016-05-22
fix(core): getCellDisplayValue fix for when flatEntityAccess enabled

jasonlikescats:
Calls to `getCellDisplayValue` would throw when `flatEntityAccess` was enabled. Fixes [#5326](https://github.com/angular-ui/ui-grid/issues/5326).

I took the liberty of moving the check for the `flatEntityAccess` option into `GridRow`'s `getEntityQualifiedColField`, which looks to be a general entry point for getting the row entities "accessor". It makes more sense to me to check in here rather than to branch functionality elsewhere (which also places the burden on the caller and could lead to accidental bugs).

Also, just a note that `getCellDisplayValue` may still be broken for the case of using `$$col.uid` accessors...I'm not too familiar with how that all works (and it's outside the scope of the bug I was trying to fix), so I'm not able to verify that at this time. This is just a heads up given that it looks like it suffers from the same root issue (calling `$parse` with a value rather than a value "accessor"). Maybe someone can confirm and create an issue?

### DIFF
--- a/src/js/core/factories/Grid.js
+++ b/src/js/core/factories/Grid.js
@@ -1875,8 +1875,6 @@ angular.module('ui.grid')
 
       if (typeof(row.entity['$$' + col.uid]) !== 'undefined') {
         col.cellDisplayGetterCache = $parse(row.entity['$$' + col.uid].rendered + custom_filter);
-      } else if (this.options.flatEntityAccess && typeof(col.field) !== 'undefined') {
-        col.cellDisplayGetterCache = $parse(row.entity[col.field] + custom_filter);
       } else {
         col.cellDisplayGetterCache = $parse(row.getEntityQualifiedColField(col) + custom_filter);
       }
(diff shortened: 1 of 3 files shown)


JLLeitschuh:
Do you have a plunker to show the before and after?
I'm not 100% sure I understand what the issue is but I haven't touched this code before.

jasonlikescats:
There's a plunker for the before in the issue I referenced above ([#5326](https://github.com/angular-ui/ui-grid/issues/5326)). I'm not too sure how to make a plunker for the after...can I hotlink from plunker to the source file from the pull request directly on Github?

JLLeitschuh:
I'm not sure. I have seen it done but I don't know.

jasonlikescats:
Hmm, fair enough. Another angle that might help is running the unit test I added against the original code, and with my pull request changes. Otherwise I can probably find some time this weekend to put together an "after" plunker.

dlgski:
@jasonlikescats We are making an effort to update ui-grid with all the contributing PRs.  This PR is now out-of-date as gridRow.js has already been updated in https://github.com/angular-ui/ui-grid/commit/65e49fd59a165672c71738e4ba7df553e7f6e673  Can you update this PR if it still resolves an issue?  Thank you!

mportuga:
Closing due to lack of response.

PR #5443, opened 2016-05-30
Update gridValidate.js

basarawu66:
Change parameter memo

### DIFF
--- a/src/features/validate/js/gridValidate.js
+++ b/src/features/validate/js/gridValidate.js
@@ -47,7 +47,7 @@
        *  ```
        *  {
        *    validatorFactory: function(argument) {
-       *                        return function(newValue, oldValue, rowEntity, colDef) {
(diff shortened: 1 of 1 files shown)


mportuga:
Why are you changing the parameter memo here? Also, please provide an issue

phillipspc:
I opened an issue about this... May just be a documentation error.
https://github.com/angular-ui/ui-grid/issues/5515

mportuga:
@JLLeitschuh @dlgski Do you guys think that we should update the code to match the documentation or the documentation to match the code? I feel like updating the documentation might be better as updating the code would likely break other people who have realized the same thing that @phillipspc realized and wrote their code the way that it is actually behaving.

JLLeitschuh:
Best practice is to always avoid API breaking changes where possible.

mportuga:
This is not the correct fix for this issue. Instead of changing the API, we should fix the documentation.

PR #5468, opened 2016-06-13
Fix #5467 - Column moving doesn't work well when setting width in columnDefs

### LINKED ISSUE #5467: Column moving doe's work well when setting width in columnDefs
Hi,
When enabling column move and setting width (even for one column) to column in columnDefs setting (or using column resize) the movement does't work well.

Scenario:
* 3 column
* 1 and 3 don't have width setting
* 2 has been set with any value in width section.
* Try to switch between 1 and 2.

Result: if you don't drug 1 over 50%-70% of 2 width the columns wont be changed.

alongros:
Change column move position calculation.


alongros:
How can I have write access in order to help and contribute the community?

dlgski:
@alongros This looks like an interesting fix.  Can you provide the issue that it's fixing (ideally issue has a plunker included) and a description of your fix?  Also, did you verify that this would not break tests?  It looks like it may.  thank you

mportuga:
Closing due to innactivity.

PR #5473, opened 2016-06-15
fix for error when element doesn't have the select function

oostmeijer:
minor change to solve errors when an element doesn't have a select function.
I ran into this problem when creating an edit cell template using a custom directive.

### DIFF
--- a/src/features/edit/js/gridEdit.js
+++ b/src/features/edit/js/gridEdit.js
@@ -938,7 +938,7 @@
                   $timeout(function () {
                     $elm[0].focus();
                     //only select text if it is not being replaced below in the cellNav viewPortKeyPress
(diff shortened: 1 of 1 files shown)


oostmeijer:
who is managing the pull requests? Seems there are a lot of them still open....

AgDude:
This PR should reference an issue with a plunker reproducing the bug. A unit test would be nice also.

oostmeijer:
https://embed.plnkr.co/Le3FOBIgkk3vN0RHGfGj/

double click the checkboxes, gives an error in the console.

mportuga:
@oostmeijer Would it also be possible for you to add a unit test so that this issue doesn't ever get reintroduced again?

PR #5491, opened 2016-06-22
fix: clear white space if not exist scrollBar

cybermerlin:


dlgski:
This PR should reference an issue with a plunker reproducing the bug. A unit test would be nice also.

cybermerlin:
sorry, but I not have many time for your pleasures.
just trust me

AgDude:
This looks to me like it will break with pinned containers. But I haven't tried it.

Also note that the tests related to multiple containers are failing.

PR #5493, opened 2016-06-24
chore: Release v3.2.1

dlgski:


dlgski:
not sure why this is pulling in all the latest changes as separate commits.  This is simply the chore to set up 3.2.1 so I can put it in NPM.  I do want all of these changes to be included, but I don't understand why they are being listed out here.

PR #5514, opened 2016-07-01
feat(i18n): add Bulgarian translation

DGarvanski:
Add Bulgarian translation to i18n feature.


AgDude:
please squash commits.

PR #5535, opened 2016-07-09
adding showTopFooter option

mikeytag:
Ok, first off I apologize if this breaks standards on this plugin or anything else, but here was my problem. I needed to add a top totals row along with a bottom totals row for some of the tables in my project.

I dug into bower_components/angular-ui-grid/ui-grid.js and was able to copy the ui-grid/ui-grid-footer template and append it in the ui-grid/ui-grid-header template and add an ng-if statement to only do it if grid.options.showTopFooter is defined and true.

When I forked the repo I saw that the project is laid out differently than my bower install. (No duh). I browsed through the src/templates folder and found ui-grid-header and copied my code there as I imagine this is the file that gets pulled into the template cache in the final ui-grid.js used in bower.

This solved a problem for me. I'm not sure if there is another request for this type of feature on the Issues board but I'll look for one. Again, I am new to angular so I apologize if I needed to make this change somewhere else or if there is a more elegant way to handle it. I figured I can't be the only one that wanted a duplicate of the footer underneath the header and I thought it would be a good idea to contribute to a project that has helped me so much.

Thanks!

Mike

### DIFF
--- a/src/templates/ui-grid/ui-grid-header.html
+++ b/src/templates/ui-grid/ui-grid-header.html
@@ -25,4 +25,18 @@
       </div>
     </div>
   </div>
+  <div ng-if="typeof(grid.options.showTopFooter) != 'undefined' && grid.options.showTopFooter"
+     class="ui-grid-footer-panel ui-grid-footer-aggregates-row"><!-- tfooter -->
+    <div class="ui-grid-footer ui-grid-footer-viewport">
+        <div class="ui-grid-footer-canvas">
+            <div class="ui-grid-footer-cell-wrapper" ng-style="colContainer.headerCellWrapperStyle()">
+                <div role="row" class="ui-grid-footer-cell-row">
+                    <div ui-grid-footer-cell role="gridcell"
+                         ng-repeat="col in colContainer.renderedColumns track by col.uid" col="col"
+                         render-index="$index" class="ui-grid-footer-cell ui-grid-clearfix"></div>
+                </div>
+            </div>
+        </div>
+    </div>
+</div>
 </div>
--- a/src/templates/ui-grid/ui-grid-header.html
+++ b/src/templates/ui-grid/ui-grid-header.html
@@ -25,7 +25,7 @@
       </div>
     </div>
   </div>
-  <div ng-if="typeof(grid.options.showTopFooter) != 'undefined' && grid.options.showTopFooter"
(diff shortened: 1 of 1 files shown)


mikeytag:
Ooops. Forgot to add a screenshot of what this looks like.
screen shot 2016-07-08 at 10 32 52 pm

src/templates/ui-grid/ui-grid-header.html
```diff
@@ -25,4 +25,18 @@
       </div>
     </div>
   </div>
+  <div ng-if="typeof(grid.options.showTopFooter) != 'undefined' && grid.options.showTopFooter"
```
Southpaw17: The first half of this expression should be unnecessary.  ngIf checks if the expression is truthy.  Therefore, if grid.options.showTopFooter is undefined, null, "false", etc. it won't display.

mikeytag: Thanks. I've been around JS for decades, but new to angular. I tend to be a bit anal with typeof to make sure I'm not asking for something that hasn't been declared yet, but it's nice to know that angular automatically traps this and I don't have to worry about it in an ng-if scenario.

mportuga: Would you mind removing the unnecessary check?

mportuga:
@dlgski It seems that this developer is no longer active. However, his changes seem okay other than that, and could be a good feature, do you think we should drop this or merge this?

mikeytag:
Hey guys, I'm still here if you want me to adjust my PR at all. We've been using the heck out of ui-grid and are really grateful for this project.

Oh, I saw you wanted me to remove the typeof() != 'undefined stuff. I can change that if you would like.

dlgski:
@mikeytag I like the concept of this, but given this is a global project we cannot have a "footer" in the "header".  I would recommend you update this PR to toggle a gridOption of "headerAggregateRow" and then update the "footer" classes to no longer specify "footer" 

as this is a new feature.  I would also request that you update documentation and add unit tests.

I think this is a great feature, and hope you can provide these updates.  Thank you.

mikeytag:
Thanks @dlgski will do. I hacked together really quickly for our own app, but I'll put in the changes you are talking about along with CSS classes and unit tests and then do another PR.

mportuga:
Closing this PR since the developer said he will do another later on.

Issue #2475, opened 2015-01-02
Need raw filter term


eirenik:
When implementing a custom filter, the search term is modified before being provided to the filter function. These modifications make it difficult or impossible to implement desired filtering logic.

There needs to be some way to get the "raw" term that the user typed in.  I think this should be the default behavior for all custom filters, but an option like "filterRaw:true" might be fine if needed for compatibility.

axelstudios:
This caused problems for me as well until I [started searching through the code](https://github.com/angular-ui/ng-grid/blob/8f7ee4c7d861557b0ff188831c8cd2aec371c787/src/js/core/services/rowSearcher.js#L215) and saw that I just needed to use the escaped searchTerm in a RegExp constructor rather than doing any kind of substring matching.

Previously, my workaround was to strip backslashes from `searchTerm`, which worked great for everything except actual backslashes in the raw filter, but RegExp is the way to go.

eirenik:
That would be fine if one's goal was to do exactly what the grid code does with filters, but the whole point of a custom filter is to do something different. Possibly extremely different.

The whole idea of extensibility is recognizing that you can't predict everything that will make sense for everyone and every possible application's data.  Maybe I want '@' to be a wildcard.  Maybe I want to append '5' to every 'k' in the filter.  Whatever it is, providing the raw input from the user is essential for permitting customization.

PaulL1:
I can also mark this as a future enhancement, and happy to accept a pull request that adds this functionality.

At the moment the (quite small) development team are focusing their time on stability and reaching an initial release of 3.0, not new functionality.

PaulL1:
We have provided the ability to define custom filter widgets and custom filter behaviour.  I think this should meet the needs here.

PR #5593, opened 2016-07-29
feat(filter): Add rawTerm option to columnDef filter options

StrangelyTyped:
Adds the rawTerm option to the columnDef filter options, when enabled
it causes the rowSearcher to skip trimming and escaping the search term,
providing the raw value to the specified filter condition. This is
intended for use with custom search condition functions where pre-escaping
all regular expression characters or trimming whitespace is not desired.

Relates to #2475


Southpaw17:
If we can get the merge conflicts resolved we can go ahead and get this merged in

PR #5606, opened 2016-08-04
Missing test for whether value is a number

prettycode:
This operation is performed in GridColumn.js, line 133, but not here in tree-base.js.

The result of not performing this check in tree-base is that "avg" aggregate counts `undefined` values as `0`, throwing off (miscalculating) averages.


prettycode:
While this fixes "average" aggregate, grouped columns now have a count of 0. Maybe someone with more time can take this problem of undefineds causing miscalculated averages and fix it in a way that doesn't have breaking side effects.

prettycode:
Here is a demonstration of the bug described in the original pull request.

demo

prettycode:
Figured out the correct fix. Closing this in favor of a new, clean, tested pull request.

Issue #5359, opened 2016-04-26
No mechanism to bind column to the row object itself


StrangelyTyped:
Occasionally I find I need a grid with just one column, and often I find that the data I need to bind to it is just an array of strings. I've tried a few different things and haven't been able to find a mechanism to populate a grid from an array of strings without reformatting the data into an array of objects (each with a single property - the original string for that row).

Attempting to do so results in a grid with the correct number of rows but no content in the specified column, as you can see in my example plunk [here](http://plnkr.co/edit/YK45yvWqvCDzzT7q11yD)

Omitting the field column property altogether or providing an empty value, or passing other possible values like 'row' result in an empty cell.

The ui-grid documentation says that the 'field' property "Should be an angular expression that evaluates against grid.options.data array element." and to "See the angular docs on binding expressions.". The angular docs specify this in a section on the [expressions page](https://docs.angularjs.org/guide/expression) - "It is possible to access the context object using the identifier this", however using 'this' as the field value does not appear valid either, resulting in an empty cell as well.

As a result I have yet to find a mechanism for using an array of strings as source data for a UI Grid. I'd like to request that such functionality be added, possibly through a special value in the column's field property ('this' or empty string would appear to be good candidates, the latter being my preference). If such a mechanism already exists and I simply haven't found it then I'd like to request that the documentation be updated to provide examples of how this is done.

PR #5611, opened 2016-08-06
feat(core): Allow binding a column to the row entity itself

StrangelyTyped:
This allows grids to be created for an array of primitives, such as
an array of Strings, and bind a column to each value in the data array
instead of having to bind to a property within each value.

Satisfies #5359


angular-git-user:
This sounds helpful, or else extra efforts of binding cols with data[i] was required

Issue #5620, opened 2016-08-11
runValidators function does not return a promise


imranamans:
I have a page with ui-grid and some form elements on it. When user clicks submit button, I am invoking the validation on the grid using below code. But since the runValidators function does not return anything. There is no way of knowing, if the validations are complete.

```javascript
// Caller invoked on submit
function caller() {

    validateGrid($scope.myGridOptions);

    // if validation is successful submit the form otherwise display message.
}

function validateGrid(gridOptions) {
    angular.forEach(gridOptions.data, function(rowEntity) {
        angular.forEach(gridOptions.columnDefs, function(colDef) {
            gridOptions.gridApi.grid.validate.runValidators(rowEntity, colDef, rowEntity[colDef.name], oldValue, gridOptions.gridApi.grid);
        })
    });
}
```
Unfortunately, runValidators does not return anything (let alone a promise). And so the caller is out of sync as the validation completes at a later time.
If runValidators were to return a promise, the below code would solve the problem.

```javascript
// Caller invoked on submit
function caller() {

    var promise = validateGrid($scope.myGridOptions);

    promise
        .then(function() {
            // success
        })
        .catch(function() {
            // failure
        })
}

function validateGrid(gridOptions) {

    var promises = [];

    angular.forEach(gridOptionss.data, function(rowEntity) {

        angular.forEach(gridOptions.columnDefs, function(colDef) {

            var promise = gridOptions.gridApi.grid.validate.runValidators(rowEntity, colDef, rowEntity[colDef.name], oldValue, gridOptions.gridApi.grid);
            promises.push(promise);
        })
    });
    return $q.all(promises);
}
```
I was able to modify ui.grid.validate.service:uiGridValidateService:runValidators to return a promise

```javascript
runValidators: function(rowEntity, colDef, newValue, oldValue, grid) {
        
        if (newValue === oldValue) {
          // If the value has not changed we perform no validation
          return;
        }
        
        if (typeof(colDef.name) === 'undefined' || !colDef.name) {
          throw new Error('colDef.name is required to perform validation');
        }
        
        service.setValid(rowEntity, colDef);
        
        var validateClosureFactory = function(rowEntity, colDef, validatorName) {
          return function(value) {
            if (!value) {
              service.setInvalid(rowEntity, colDef);
              service.setError(rowEntity, colDef, validatorName);
              if (grid) {
                grid.api.validate.raise.validationFailed(rowEntity, colDef, newValue, oldValue);
              }
            }
[...]

PR #5626, opened 2016-08-15
fix for #5620: runValidators function does not return a promise 

imranamans:
Please share comments if any. Thanks.


mportuga:
In the future please follow our commit message standards outlined here:
https://github.com/angular-ui/ui-grid/blob/master/CONTRIBUTING.md#commit-message-format

I suggest using commitizen since it makes it really easy to have the standard be followed.

Also, if you could squash your commits ahead of time as outlined there, it would help too.
https://github.com/angular-ui/ui-grid/blob/master/CONTRIBUTING.md#pull-requests

PR #5633, opened 2016-08-19
Change doc comments - 'RowSorter' to 'rowSorter'

Cfeusier:
When glancing at docs, the module is displayed as 'RowSorter'. Users (like myself) might try to inject `RowSorter` instead of `rowSorter`, with resulting DI errors and frustration.


mportuga:
In the future please follow our commit message standards outlined here:
https://github.com/angular-ui/ui-grid/blob/master/CONTRIBUTING.md#commit-message-format

I suggest using commitizen since it makes it really easy to have the standard be followed.

PR #5639, opened 2016-08-23
Unable to export all filtered data to CSV/PDF

ramprasadm:
Facing issue while exporting filtered data to PDF/CSV. It is exporting only Visible Data/All Data, but not all filtered rows. Found an workaround to export filtered data, tweak the logic in getData() function in ui-grid.js solves my problem. Added a AND condition "row.visible" while preparing data for PDF/CSV. After this change I am able to export all the filtered rows to PDF/CSV.

Code Snippet ui-grid.js:

```
getData: function (grid, rowTypes, colTypes) {

rows.forEach( function( row, index ) {
            if (row.exporterEnableExporting !== false && row.visible) {
              var extractedRow = [];
```
Custom Filter:

```
$scope.filter = function() {
          $scope.gridApi.grid.refresh();
      };

      $scope.singleFilter = function( renderableRows ){
              var matcher = new RegExp($scope.filterValue);
              renderableRows.forEach( function( row ) {
                var match = false;
                [ 'name'].forEach(function( field ){
                      if ( row.entity[field] != null && row.entity[field].match(matcher) ){
                       match = true;
                      }
                });
                if ( !match ){
                  row.visible = false;
                }
              });
              return renderableRows;
      };
```


mportuga:
There are too many commits in this PR. Please reopen it with a singular commit and fix unit tests.

PR #5659, opened 2016-09-01
Fix csv export on ionic application

### LINKED ISSUE #4930: Doesn't Export to CSV when used with IONIC
I am trying to use UI-GRID with IONIC framework, and export to CSV doesn't work. Export to pdf is working is working fine though. I am using Ionic 1.1.1 and uiGrid 3.0.7 with Angular 1.4.3

gine:
Fix #4930 

Here the plunker to reproduce the problem: http://plnkr.co/edit/rb0JFHYr1LGPzscFYAmm?p=preview
Builded from the example on this page: http://ui-grid.info/docs/#/tutorial/206_exporting_data

### DIFF
--- a/src/features/exporter/js/exporter.js
+++ b/src/features/exporter/js/exporter.js
@@ -958,7 +958,10 @@
           a.setAttribute('style', 'display:none;');
           D.body.appendChild(a);
           setTimeout(function() {
-            if (a.click) {
+						if(window.ionic && window.saveAs){
+							//Fix on ionic app
+							saveAs(blob, fileName);
+            } else if (a.click) {
               a.click();
               // Workaround for Safari 5
             } else if (document.createEvent) {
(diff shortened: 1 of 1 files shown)


dlgski:
Can you update your commits so they pass the the test suite?  thanks!

gine:
Doesn't it has pass already the test suite? Confused.

gine:
Still waiting clarification..

AgDude:
I am not in favor of UI grid making special cases for other frameworks (even ionic/cordova). From a quick glance at the code, I think the right solution is to break the `uiGridExporterService.downloadFile` into two smaller methods, one which creates the content, and the other which triggers the download. The first could then be exposed in the public api and in the case of cordova the download could be handled at the application level.

This PR is also probably wrong in that this is not specific to ionic, but more generally to cordova.

PR #5660, opened 2016-09-02
added italian pagination translation

Marchrius:


mportuga:
Is there an issue number or chore associated with this?

Issue #4400, opened 2015-09-18
Moving columns on iPad does not work (e.pageX, e.PageY are not defined)


yurigenin:
When you try to move a column on iPad (try Column Moving demo in Tutorial) it just does not work. The problem is that e.pageX, e.PageY are not defined in 'touch...' events. You have to instead use this code:

```javascript
var pointerEventToXY = function(e){
		var out = {
		  x: 0, 
		  y: 0
		};
		
		if (e.type == 'touchstart' || e.type == 'touchmove' || e.type == 'touchend' || e.type == 'touchcancel'){
			var touch = e.originalEvent.touches[0] || e.originalEvent.changedTouches[0];
			out.x = touch.pageX;
			out.y = touch.pageY;
		} 
		else if (e.type == 'mousedown' || e.type == 'mouseup' || e.type == 'mousemove' || e.type == 'mouseover'|| e.type == 'mouseout' || e.type == 'mouseenter' || e.type == 'mouseleave') {
			out.x = e.pageX;
			out.y = e.pageY;
		}
		
		return out;
    };
```

And you call it like this: 
```javascript
	var pageXY = pointerEventToXY(e);
	var pageX = pageXY.x;
	var pageY = pageXY.y;
```

dlgski:
we have come up with a solution to use originalEvent.pageX that will be pushed in 3.2.7

PR #5666, opened 2016-09-09
fix(columnMoving): handle touch events properly when jQuery is used

evmac:
When using jQuery, touch events are wrapped in new object (`event`) that removes the `event.pageX` property, causing column moving to break when used on mobile devices. However, jQuery stores the original native object in `event.originalEvent` ([docs](https://api.jquery.com/category/events/event-object/)).

If `event.pageX` is not set, this logic will instead use `event.originalEvent.pageX`, allowing column moving to work in conjunction with jQuery on mobile devices.

### DIFF
--- a/src/features/move-columns/js/column-movable.js
+++ b/src/features/move-columns/js/column-movable.js
@@ -324,7 +324,7 @@
                     gridLeft += $scope.grid.renderContainers.left.header[0].getBoundingClientRect().width;
                   }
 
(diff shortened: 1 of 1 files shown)


evmac:
[Issue #4400](https://github.com/angular-ui/ui-grid/issues/4400) is associated with this.

mportuga:
Please squash your commits

Southpaw17:
+1

PR #5670, opened 2016-09-09
fix(Grid): Add missing promise wait in refresh().

### LINKED ISSUE #2901: ui-grid auto-resize and flex-box layout together
I want a flexbox layout with header, main content and footer .
I choose a flexbox because it is very simple and straighforward solution.
Key feature is main content block can be any height, but not less then 100% of remaining screen size. So if main content will be only one word then footer will still be at the bottom of the page. But if content will be big then footer will slide down off the screen.

For this to work I set css property `min-height: 100%` to `.site` class (container for header, main content and footer). And all works fine.

But when I put ui-grid with auto-resize module to main content's section it initially has a right size but every tick (250ms) increases it's size by 0.2 pixel. So the main content's section is growing all the time. Every 1 sec for 1 pixel.

This bug only appears if browser zoom is set to not 100% (110, 125, 150 etc).
Try to open http://plnkr.co/edit/0jPhNxJlLF3VJeOvYh3u?p=preview and then press `CTRL +` to set zoom to 110-150%. Then rerun the plunkr. You will see the vertical scrollbar continiously growing.

I tested it in latest Chrome and Firefox. Both have this bug if zoom is not 100%.

You can try to uncomment `height: 100%;` css rule for the `.site` class and see this fix the problem. But now if we put any other content into `main` section (big enought for vertical scrolling) then our footer still be on screen. That is not what we want.

elijahlofgren:
Fix issue with ui-grid-auto-resize where every tick (250ms) the grid increases by 0.2 pixel. Also fixes Firefox issue where grid rows would sometimes be endlessly redrawn if ui-grid-auto-resize was used.

Fixes #2901

### DIFF
--- a/src/js/core/factories/Grid.js
+++ b/src/js/core/factories/Grid.js
@@ -2060,10 +2060,10 @@ angular.module('ui.grid')
       self.setVisibleColumns(renderableColumns);
     });
 
-    return $q.all([p1, p2]).then(function () {
-      self.redrawInPlace(rowsAltered);
+    var p3 = self.refreshCanvas(true);
 
-      self.refreshCanvas(true);
+    return $q.all([p1, p2, p3]).then(function () {
+      self.redrawInPlace(rowsAltered);
     });
   };
 


dlgski:
@elijahlofgren please update the unit tests so this PR passes

mportuga:

Issue #4138, opened 2015-08-07
Custom cellFilter and 'getRowVisibility' used in 'onFilterChanged' not returning correct rows


Nightmaster:
I have find a problem. If you apply a custom cell-filter to a column, and then you use the ```gridApi.core.on.filterChanged``` function to call a function wich calls ```gridApi.core.getVisibleRows()```, the returned rows by the getVisibleRows are the one visible before the filter was applied and the function do not have any argument.

I have done a Plunker (based on ui-grid.info/docs/#/tutorial/103_filtering page) to show what I mean
http://embed.plnkr.co/52wNFcQtA6ACuXkHP8Yy/preview

xak2000:
I have the same problem.
```
gridApi.core.on.filterChanged($scope, function () {
  // here rows returned are not for current filter but for previous filter
  var rows = $scope.gridApi.core.getVisibleRows($scope.gridApi.grid);
});
```

My filter condition is:
```
            filter: {
              condition: function (searchTerm, cellValue) {
                return cellValue.fio().toLowerCase().indexOf(searchTerm.toLowerCase()) !== -1;
              }//uiGridConstants.filter.CONTAINS
            }
```

Each time `filterChanged` fired, `getVisibleRows` returns not new visible (filtered) rows but rows before new filter is applied.

Currently I find a workaround - run `getVisibleRows` from `filterChanged` in `$timeout` after 100ms.
```
gridApi.core.on.filterChanged($scope, function () {
  $timeout(function () {
    // here rows returned are for current filter
    var rows = $scope.gridApi.core.getVisibleRows($scope.gridApi.grid);
  }, 0);
});
```
But I worry that it is not correct as based on assumption that 100ms is enought. But in some cases it can be not.

xak2000:
I tried to use `gridApi.core.on.rowsVisibleChanged` event with better results.
It doesn't work without timeout but at least it consistently work with timeout 0.

```
        gridApi.core.on.rowsVisibleChanged($scope, function () {
          // still doesn't work
          var rows = $scope.gridApi.core.getVisibleRows($scope.gridApi.grid);
        });
```

```
        gridApi.core.on.rowsVisibleChanged($scope, function () {
          // does work!
          $timeout(function () {
            var rows = $scope.gridApi.core.getVisibleRows($scope.gridApi.grid);
          }, 0);
        });
```

But there are another problem: `rowsVisibleChanged` is not fired after filter is cleared! I think it is also a bug.

[...]

Issue #3901, opened 2015-07-01
When clearing filter with "x" key, event filterChanged fires but not rowsVisibleChanged


valepu:
see this plunkr: http://plnkr.co/edit/NPazbil7fCU39SgHviDf?p=preview

To test it:
1) Open the javascript console while on the page
2) Clear the "Company Name" filter by clicking on the x next to it
Result:
on the console you will see "Event fired: filterChanged" but not "Event fired: rowsVisibleChanged"

jjcad:
+1

jjcad:
I did notice that clearing filters in either the column header or using "Clear All Filters" in the menu does kick off the 'rowsRendered' event.

It seems like the issue might be with `Grid.prototype.setVisibleRows`, which only raises a 'rowsRendered' event.  Since it's responsible for setting the visible rows, it seems sensible that it should also let everybody know that the visible rows changed instead of only letting everybody know that the rows are rendered.

dlgski:
applied PR to fix this v3.2.10

PR #5674, opened 2016-09-10
fix(3901): Raise rowsVisibleChanged on setVisibleRows.

jeremy-j-ackso:
Several issues and Stackoverflow threads indicate that one would expect the rowsVisibleChanged event to be emitted when a filter is applied. It seems pretty rational that rowsVisibleChanged ought to be raised by setVisibleRows(), which is what this small PR implements.

A related issue not addressed by this PR is that many of the SO threads and issues also seem to believe that a filtersChanged event should be raised after the visible rows are changed via filters. This is a common misunderstanding of how ui-grid operates in that changing filters are a different operation from changing visible rows. Since there seems to be a timing/ordering issue between when filtersChanged is raised and setVisibleRows is called this probably merits discussion before/if a change is made. Perhaps a tutorial page on events could alleviate this confusion?

https://github.com/angular-ui/ui-grid/issues/3901
https://github.com/angular-ui/ui-grid/issues/4138
http://stackoverflow.com/questions/27301690/angular-ui-grid-ng-grid-filterchanged-firing-too-frequently
http://stackoverflow.com/questions/33764946/how-to-get-filtered-data-from-paged-ui-grid

### DIFF
--- a/src/js/core/factories/Grid.js
+++ b/src/js/core/factories/Grid.js
@@ -1391,6 +1391,7 @@ angular.module('ui.grid')
         self.renderContainers[targetContainer].visibleRowCache.push(row);
       }
     }
+    self.api.core.raise.rowsVisibleChanged(this.api);
     self.api.core.raise.rowsRendered(this.api);
   };
 


mportuga:
@jeremy-j-ackso It might make sense to add a unit test for this.

dlgski:
Accepting this PR.  Can you update tutorial as you suggest as well?

PR #5687, opened 2016-09-21
Feature - Empty Base Layer

louwie17:
This pull request adds a new feature 'emptyBaseLayer', that provides a background for the ui-grid, of continues rows, incase the height of the ui-grid is greater then the number of rows displayed.
This simulates a similar scenario as Excel, where the rows continue forever. The only exception is that the rows can't be edited, and just act as a background.
The rows/columns work with column resizing, and pinned rows.

A new tutorial page has been added. This is what it looks like in the tutorial:
emptybaselayer

### DIFF
--- a/src/features/empty-base-layer/js/emptyBaseLayer.js
+++ b/src/features/empty-base-layer/js/emptyBaseLayer.js
@@ -1,0 +1,164 @@
+(function () {
+  'use strict';
+
+  /**
+   * @ngdoc overview
+   * @name ui.grid.emptyBaseLayer
+   * @description
+   *
+   * # ui.grid.emptyBaseLayer
+   *
+   * <div class="alert alert-warning" role="alert"><strong>Alpha</strong> This feature is in development. There will almost certainly be breaking api changes, or there are major outstanding bugs.</div>
+   *
+   * This module provides the ability to have the background of the ui-grid be empty rows, this would be displayed in the case were
+   * the grid height is greater then the amount of rows displayed.
+   *
+   * <div doc-module-components="ui.grid.emptyBaseLayer"></div>
+   */
+  var module = angular.module('ui.grid.emptyBaseLayer', ['ui.grid']);
+
+
+  /**
+   *  @ngdoc service
+   *  @name ui.grid.emptyBaseLayer.service:uiGridBaseLayerService
+   *
+   *  @description Services for the empty base layer grid
+   */
+  module.service('uiGridBaseLayerService', ['gridUtil', '$compile', function (gridUtil, $compile) {
+    var service = {
+      initializeGrid: function (grid, disableEmptyBaseLayer) {
+
+        /**
+         *  @ngdoc object
+         *  @name ui.grid.emptyBaseLayer.api:GridOptions
+         *
+         *  @description GridOptions for emptyBaseLayer feature, these are available to be
+         *  set using the ui-grid {@link ui.grid.class:GridOptions gridOptions}
+         */
+        grid.baseLayer = {
+          emptyRows: []
+        };
+
+        /**
+         *  @ngdoc object
+         *  @name enableEmptyGridBaseLayer
+         *  @propertyOf  ui.grid.emptyBaseLayer.api:GridOptions
+         *  @description Enable empty base layer, which shows empty rows as background on the entire grid
+         *  <br/>Defaults to true, if the directive is used.
+         *  <br/>Set to false either by setting this attribute or passing false to the directive.
+         */
+        //default option to true unless it was explicitly set to false
+        if (grid.options.enableEmptyGridBaseLayer !== false) {
+          grid.options.enableEmptyGridBaseLayer = !disableEmptyBaseLayer;
+        }
+      },
+
+      setNumberOfEmptyRows: function(viewportHeight, grid) {
+        var rowHeight = grid.options.rowHeight,
+          rows = Math.ceil(viewportHeight / rowHeight);
+        if (rows > 0) {
+          grid.baseLayer.emptyRows = [];
+          for (var i = 0; i < rows; i++) {
+            grid.baseLayer.emptyRows.push({});
+          }
+        }
+      }
+    };
+    return service;
+  }]);
+
+  /**
+   *  @ngdoc object
+   *  @name ui.grid.emptyBaseLayer.directive:uiGridEmptyBaseLayer
+   *  @description Shows empty rows in the background of the ui-grid, these span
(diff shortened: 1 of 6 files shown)


src/features/empty-base-layer/test/emptyBaseLayer.spec.js
```diff
@@ -0,0 +1,53 @@
...
+
+  var scope, element, timeout, viewportHeight, emptyBaseLayerContainer;
+
+  beforeEach(module('ui.grid.emptyBaseLayer'));
+
+  beforeEach(inject(function (_$compile_, $rootScope, $timeout, $httpBackend) {
+
+    var $compile = _$compile_;
+    scope = $rootScope;
+    timeout = $timeout;
```
mportuga: Is this timeout variable really needed? It does not get used anywhere. Even $timeout itself is only used inside this beforeEach function.

src/features/empty-base-layer/test/emptyBaseLayer.spec.js
```diff
@@ -0,0 +1,53 @@
...
+      });
+    };
+
+    //$httpBackend.when('GET', 'expandableRowTemplate.html').respond("<div class='test'></div>");
+    element = angular.element('<div class="col-md-5" ui-grid="gridOptions" ui-grid-empty-base-layer></div>');
+
+    $timeout(function () {
+      $compile(element)(scope);
+    });
+    $timeout.flush();
```
mportuga: Why are you wrapping the compilation in a timeout? You could just as easily add a scope.$apply() function call after compiling the element instead of adding a timeout.

src/features/empty-base-layer/js/emptyBaseLayer.js
```diff
@@ -0,0 +1,152 @@
...
+         */
+
+        /**
+         *  @ngdoc object
+         *  @name enableEmptyGridBaseLayer
+         *  @propertyOf  ui.grid.emptyBaseLayer.api:GridOptions
+         *  @description Enable empty base layer, which shows empty rows as background on the entire grid
+         *  <br/>Defaults to true, if the directive is used.
+         */
+        grid.options.enableEmptyGridBaseLayer = grid.options.enableEmptyGridBaseLayer !== false;
```
louwie17: No, the two lines are different, I wanted to give the option of passing in true/false into the uiGridEmptyBaseLayer directive, like this: ui-grid-empty-base-layer="true/false". I wanted that to trump the grid.options value. That's why I check that before hand.
Unless you have a different suggestion?

src/features/empty-base-layer/js/emptyBaseLayer.js
```diff
@@ -0,0 +1,152 @@
...
+      },
+
+      setNumberOfEmptyRows: function(viewportHeight, grid) {
+        var rowHeight = grid.options.rowHeight;
+        var rows = Math.ceil(viewportHeight / rowHeight);
```
louwie17: Will make that change!

src/features/empty-base-layer/js/emptyBaseLayer.js
```diff
@@ -0,0 +1,152 @@
...
+        return {
+          priority: -200,
+          scope: false,
+          compile: function ($elm, $attrs) {
+            var emptyBaseLayerContainer = $templateCache.get('ui-grid/emptyBaseLayerContainer');
+            $elm.prepend(emptyBaseLayerContainer);
+            return {
+              pre: function ($scope, $elm, $attrs, controllers) {
+              },
+              post: function ($scope, $elm, $attrs, controllers) {
```
louwie17: I did this as it seemed to be the standard for the other features, mostly looked at https://github.com/angular-ui/ui-grid/blob/master/src/features/expandable/js/expandable.js#L531

src/features/empty-base-layer/less/emptyBaseLayer.less
```diff
@@ -0,0 +1,9 @@
+@import '../../../less/variables';
+
+.ui-grid-viewport .ui-grid-empty-base-layer-container {
+  position: absolute;
+  height: 100% !important;
```
louwie17: The reason for using is because the same div inherits from 'ui-grid-canvas' as well, which gets a dynamic height/width, but I always wanted this div to take the full height of the viewport. So I should be able to find a different way.

src/features/empty-base-layer/test/emptyBaseLayer.spec.js
louwie17: Probably the only reason for doing this was going of these tests: https://github.com/angular-ui/ui-grid/blob/master/src/features/expandable/test/expandable.spec.js#L29

src/features/empty-base-layer/js/emptyBaseLayer.js
```diff
@@ -0,0 +1,152 @@
...
+   */
+  module.service('uiGridBaseLayerService', ['gridUtil', '$compile', function (gridUtil, $compile) {
+    var service = {
+      initializeGrid: function (grid, baseLayerAttr) {
+
+        grid.baseLayer = {
+          emptyRows: []
+        };
+
+        grid.options.enableEmptyGridBaseLayer = baseLayerAttr !== "false";
```
mportuga: undefined and null are both different than "false" couldn't that cause a problem? Is there a way for you to ensure that baseLayerAttr is a boolean when it comes in instead of a string?

src/features/empty-base-layer/js/emptyBaseLayer.js
mportuga: I see. I can't think of another suggestion right now, but perhaps one of the other devs will.

src/features/empty-base-layer/js/emptyBaseLayer.js
mportuga: Makes sense. I guess this should stay this way then.

src/features/empty-base-layer/less/emptyBaseLayer.less
mportuga: That would be great, thanks!

src/features/empty-base-layer/test/emptyBaseLayer.spec.js
mportuga: I will try to fix those later in that case. But in the meantime there is no reason that new code should follow the same bad practice.

louwie17:
@mportuga I updated your suggested changes, see what you think? Also I squashed all the commits, hopefully that looks a little better. It was my first time doing that actually, so I had some fun screwing up the rebasing here and there :)

PR #5740, opened 2016-10-13
vertical scroll length should never be 0. This will result in division by 0 errors.

Ladekarl:
If a grids viewport height, canvas height and scrollbar height has values so that the calculation:
canvas height - viewport height + scrollbar height returns 0, all the rows will dissapear when the grid is scrolled.

### DIFF
--- a/src/js/core/factories/GridRenderContainer.js
+++ b/src/js/core/factories/GridRenderContainer.js
@@ -300,7 +300,7 @@ angular.module('ui.grid')
   };
 
   GridRenderContainer.prototype.getVerticalScrollLength = function getVerticalScrollLength() {
-    return this.getCanvasHeight() - this.getViewportHeight() + this.grid.scrollbarHeight;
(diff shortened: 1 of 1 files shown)


dlgski:
can you provide an issue/plunker?

Ladekarl:
Scroll the grid in this plunker.
The screen must have **1200 pixels in height** (tested on 1600x1200 and 1920x1200 screens).
The chrome browser must be **maximized**.
Of course this issue can occur on other screen resolutions but the grid here is modified so that the issue occurs on screens with 1200 pixels in height.

http://run.plnkr.co/plunks/vXxKGY/

**Before scroll:**
image

**After scroll:**
image

Ladekarl:
Noticed the same calculation is done when scrolling horizontally. Updated the pull request to also fix horizontal scrolling

mportuga:
@Ladekarl I tried setting my browser window to the sizes that you mention and I don't have a scroll bar at those points. I also tried resizing my window as to get a scroll bar and I can't reproduce it.

Ladekarl:
@mportuga it should be clear from the code that a division by 0 error is plausible. Do you want me to record a video for you?

Ladekarl:
I recorded a video: http://webmshare.com/jMJ1d

PR #5744, opened 2016-10-16
fix(cellnav): when grid has only one focusable column, should navigat…

sglogowski:
When only one columns is focusable function 'getRowColRight' should return 'GridRowColumn' from next row and function 'getRowColLeft' should return 'GridRowColumn' from previous row. Similar behave occurred now(only when grid has more than one focusable columns) from first and last focusable columns in row.

### DIFF
--- a/src/features/cellnav/js/cellnav.js
+++ b/src/features/cellnav/js/cellnav.js
@@ -128,7 +128,7 @@
         var nextColIndex = curColIndex === 0 ? focusableCols.length - 1 : curColIndex - 1;
 
         //get column to left
-        if (nextColIndex > curColIndex) {
+        if (nextColIndex >= curColIndex) {
           // On the first row
           // if (curRowIndex === 0 && curColIndex === 0) {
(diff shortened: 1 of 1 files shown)


dlgski:
can you provide a plunker?

sglogowski:
Yes, sure : )

Here are two examples, first has grid with two focusable columns:

https://embed.plnkr.co/gFOVEDQxlcARf1n7Wc5V/

Second has grid with only one focusable column:

https://embed.plnkr.co/l1LPBdAz4vMnbmdwdhJv/

Try on both grids move between cells, like you can see on first grid after last cell in row focus move to next cell in next row, in second example focus is locked. The behave with 'Tab' key(and 'Shift' + 'Tab') is even worst, the focus in second example move out of grid(where in first example moves over cells and rows)

I'm trying create example with my fix, but i can't attach that large file to plunker 

I hope i described this in clear way, if you have more question about this, please feel free to ask.

PR #5761, opened 2016-10-20
Fixes #3832

### LINKED ISSUE #3832: paginationPageSizes breaks if rowHeight is a string
screen shot 2558-06-22 at 09 57 09

I have problem about "paginationPageSizes: [20, 50, 75]" when I have change number of view item in a footer like below picture 
screen shot 2558-06-22 at 09 58 32

I get error about data cannot show follow by number of change eg. I have set default size = 20 it seem to woke fine and when I change other number a table show my data just 7 items  

Thank you.
Best regard

DeusEx-Machina:


src/js/core/factories/GridOptions.js
```diff
@@ -224,10 +224,17 @@ angular.module('ui.grid')
...
        */
-      baseOptions.rowHeight = baseOptions.rowHeight || 30;
+
+      if (typeof baseOptions.rowHeight === "string") {
+        baseOptions.rowHeight = parseInt(baseOptions.rowHeight) || 30;
+      }
+
+      else {
+        baseOptions.rowHeight = baseOptions.rowHeight || 30;
+      }
```
dlgski: you can get rid of the condition and simply have:
```
baseOptions.rowHeight = parseInt(baseOptions.rowHeight) || 30;
```

parseInt on an integer will just return it's integer value.

dlgski: I'm not 100% sure we should be bothering with this.  I think we document that it should be an integer value, that should be enough

mportuga:

Issue #4778, opened 2015-11-27
Sorting grouped columns fails in chrome


robertreynisson:
When sorting grouped columns where row-count exceeds 9 the sorting fails. The data in the column is an integer. Strings work.

http://plnkr.co/edit/yqLPx1bu40ainD1VT9Bw?p=preview

Thanks
Róbert

robertreynisson:
This is bound to chrome

robertreynisson:
This is bound to chrome

a002194:
We facing the same issues in chrome, Custom Sorting Algorithm with Grouping is creating duplicate records in Grid.

grouping: {
                         groupPriority: 0
                     },
                     sort: {
                         direction: uiGridConstants.DESC,
                         priority: 0,
                     },
                     sortingAlgorithm: function (a, b, rowA, rowB, direction) {
                         var nulls = $scope.gridApi.core.sortHandleNulls(a, b);
                         if (nulls !== null) {
                             return nulls;
                         }
                         else {
                             if (a === b) {
                                 return 0;
                             }
                             if (a.indexOf('New') != -1) {
                                 return 1;
                             }
                             if (b.indexOf('New') != -1) {
                                 return -1;
                             }
                             if (a.indexOf('On-Hold') != -1) {
                                 return 1;
                             }
                             if (b.indexOf('On-Hold') != -1) {
                                 return -1;
                             }
                             if (a.indexOf('In-Progress') != -1) {
                                 return 1;
                             }
                             if (b.indexOf('In-Progress') != -1) {
                                 return -1;
                             }
                             if (a.indexOf('In-Queue') != -1) {
                                 return 1;
                             }
                             if (b.indexOf('In-Queue') != -1) {
                                 return -1;
                             }
                             if (a.indexOf('In-Review') != -1) {
                                 return 1;
                             }
                             if (b.indexOf('In-Review') != -1) {
                                 return -1;
                             }
                             return 0;
                         }
                     }

brafkind:
This is due to the fact that the group header rows are strings by default. Note the comment on _groupingShowCounts_ under [GridOptions ui.grid.grouping](http://ui-grid.info/docs/#/api/ui.grid.grouping.api:GridOptions)

> shows counts on the groupHeader rows. Not that if you are using a cellFilter or a sortingAlgorithm which relies on a specific format or data type, showing counts may cause that to break, since the group header rows will always be a string with groupingShowCounts enabled. 
[...]

Issue #4478, opened 2015-10-05
Booleans (checkboxes) with Safari v9 (El Capitan)


EricWVGG:
I am finding that booleans — both with my own code, the tutorials, and various plunkers — do not work with the latest version of Safari 

Double-clicking a cell will bring up a checkbox, but checking it simply unfocuses the cell. The same pages work fine in Chrome, so I believe this is definitely a browser bug.

I do not have an older version of Safari available to test against “Yosemite” or “Mavericks”.

EricWVGG:
The checkboxes do function correctly in iOS Safari 9, incidentally.

farooqu:
This seems to be the same issue as #1785.

Issue #4782, opened 2015-11-28
Not able to change/click boolean value checkboxes on iPhone


qdouble:
If I click on a boolean cell to enable editing, if I try to check/uncheck it, it simply removes focus from the row. It seems impossible to change the value. Anyone know a fix or workaround for this issue?

Issue #1785, opened 2014-10-10
Boolean type Edit by clicking issue on Firefox/Safari


yuwong:
From the Edit Feature tutorial http://ui-grid.info/docs/#/tutorial/201_editable on Firefox/Safari, the Active column's boolean value is not changeable by clicking. It can only be edited when in focus using the spacebar. The checkbox appears to work fine on Chrome using clicks.

PaulL1:
I think the problem is html5 widgets - and the fact that Safari and IE don't implement some of the html5 widget types.  Does that sound correct?

PaulL1:
Closing, no further comments for 2 weeks, and no plausible path to a resolution using the vanilla widgets.

yuwong:
Thanks Paul, decided against using the Boolean type altogether now

bradlys:
Anyone know how to resolve this? Custom templates or not.

I'm willing to contribute code if anyone knows where the issue lies.

PR #5771, opened 2016-10-26
fix(edit): fix boolean edit issue on Firefox and Safari on macOS

farooqu:
Add mousedown listener to ui.grid.edit.directive:uiGridEditor to disable
blur handler if the element is a checkbox, and then re-focus the
checkbox and enable the blur handler after a $timeout. This change is to
deal with Safari and Firefox behavior in macOS where clicking the
checkbox causes a blur event, and the value is not updated.

GitHub issues: #1785, #4778, #4782

### DIFF
--- a/src/features/edit/js/gridEdit.js
+++ b/src/features/edit/js/gridEdit.js
@@ -966,9 +966,20 @@
                     });
                   }
 
-                  $elm.on('blur', function (evt) {
-                    $scope.stopEdit(evt);
+                  // macOS will blur the checkbox when clicked in Safari and Firefox,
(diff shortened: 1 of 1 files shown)


farooqu:
This failure doesn't seem to be related to any of these changes.

mportuga:
@farooqu You mention, Issue #4778 on your description, but that does not appear to be related, can you explain what is the connection? Other than that, this PR seems fine.

farooqu:
I mistyped that issue. This does not fix issue #4778, but does fix #4478.

PR #5797, opened 2016-11-08
Fixes issue #5799, eventOrigin not getting passed to begin edit when edit mode is triggered by cellNav

jedancona:
Fixes an issue where when ALL FEATURES are applied to grid and
enableCellEditOnFocus is set to true the edit event would not pass the
event to the beginEdit method causing issues with custom editors that
rely on the type of event that triggered the cell into edit mode.


dlgski:
Hi @jedancona can you reference the issue as part of your PR description?  plunkers are helpful as well

jedancona:
The plunker below demonstrates the issue that this PR fixes.  Take a look at the comments in the ui-custom-editor.js file.

http://plnkr.co/edit/FCqb3G5I70hyE2JrPFhN?p=preview

Issue #5558, opened 2016-07-17
es6 module import for ui.grid


chyzwar:
Hi, 

I am using es6 modules in webpack. When importing uiGrid I get empty object insted name that can be used in module definition. uiRouter and others in this example properly give string name for module.

``` js
import 'angular-material/angular-material.css';
import 'angular-ui-grid/ui-grid.css';

import angular from 'angular';
import ngAnimate from 'angular-animate';
import ngAria from 'angular-aria';
import ngMaterial from 'angular-material';

import uiRouter from 'angular-ui-router';
import uiGrid from 'angular-ui-grid';

import AppComponent from './app.component';
import Common from './common/common';
import Components from './components/components';

angular.module('ng-app', [
  ngAnimate,
  ngAria,
  ngMaterial,

  uiRouter,
  //uiGrid -> This gives empty object

  Common.name,
  Components.name
])
.directive('app', AppComponent)
```

Chyzwar:
after checking what others do, It might be needed to add index.js with content simmilar to:

```js
require('./angular-aria');
module.exports = 'ngAria';
```

This would solve problem for most modules lie clj, es, browserity...

d-amit:
+1

GregePorter:
And for a quick solution, you can just use 'ui.grid' instead of uiGrid in your module statement

Chyzwar:
Yeah I know that this is solution, but it would be nice to have consistent usage with ES6.

GregePorter:
@Chyzwar , sure sure. I wanted to make a note here just in case someone else came across this page looking for a solution.

csvan:
+1

PR #5801, opened 2016-11-09
build(package.json): add index.js

poloagustin:
Added index.js with commonjs for webpack

#5558

I don't know how to make the index.js file be a part of the published package. Please advice on this.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,0 +1,2 @@
+require('./ui-grid.js');
+module.exports = 'ui.grid';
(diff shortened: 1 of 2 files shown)


el-besto:
Any idea when this would be merged and a new release sent to NPM?

mportuga:
Probably for release 4.1 or so, we were already in progress with 4.0, so it will just take a little longer.

mportuga:
This alone will not work. You need to also update the copy task to ensure this file gets published.

dlgski:
I'm ok including this in our next patch.  I would take it on with hesitation as I believe there is a reason why we did not have a 'main' in our package.json currently.

Issue #5764, opened 2016-10-21
ui-grid-expandable - comumns mixing values after expand


dark0q:
When I expand row the columns in below row are mixing (display values in wrong position).

I had example form tutorial: http://ui-grid.info/docs/#/tutorial/216_expandable_grid - but I added more columns and set width for each column.

This is an example of bad behaviour:
http://plnkr.co/edit/nYXQ2A0HszXTF7WiiLTr?p=preview

PR #5811, opened 2016-11-14
fix ui-grid-expandable - columns mixing values after expand

dark0q:
Fix for issue https://github.com/angular-ui/ui-grid/issues/5764


mportuga:
In the future please follow our [commit standard](https://github.com/angular-ui/ui-grid/blob/master/CONTRIBUTING.md#-git-commit-guidelines) so that your contribution can show up in our change log. I suggest using commitzen.

PR #5818, opened 2016-11-16
Update getCellDisplayValue

sagarkhatri200:
Currently any filter does not have any access to the coldef. Because of this filters that make use of editabledropdownarray can't use getCellDisplayValue method. this fix should allow ones to  build filters that has access to both row and column.

### DIFF
--- a/src/js/core/factories/Grid.js
+++ b/src/js/core/factories/Grid.js
@@ -1886,11 +1886,11 @@ angular.module('ui.grid')
       } else if (this.options.flatEntityAccess && typeof(col.field) !== 'undefined') {
         col.cellDisplayGetterCache = $parse(row.entity[col.field] + custom_filter);
       } else {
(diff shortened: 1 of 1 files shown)


mportuga:
@sagarkhatri200 Can you add unit tests for your change?

@dlgski this looks like it might be a breaking change, since it changes the contract for that function.

src/js/core/factories/Grid.js
```diff
@@ -1886,11 +1886,11 @@ angular.module('ui.grid')
       } else if (this.options.flatEntityAccess && typeof(col.field) !== 'undefined') {
         col.cellDisplayGetterCache = $parse(row.entity[col.field] + custom_filter);
```
mportuga: Wouldn't the same change that you made on line 1889 be needed here?

mportuga:
Also, in the future, please follow our guidelines for commit messages so that we can insure your change shows up in our changelog. You can find them in Contributing.md

mportuga:

Issue #5007, opened 2016-01-21
Digest cycles initiated by the grid


yccteam:
There are way too many digest cycles for the grid load/rendering. I've managed to reduce the load by setting some of the `$timeout`s to `$timeout(method, 0, false)`.  I've managed to reduce the number by 2/3 this way, with no seeming side effects (and my app now loads much faster - I have around 9-10 grids on one page, that can open side menus with more, so this is critical...). 

I've created a plnkr that shows that every load, you get roughly 35 digest cycles - with no extra feature involved:
http://plnkr.co/edit/uOYKeJvANJu1b9zk8Ac1?p=catalogue (it also shows that the grid increases the number of digests each time it reloads which might indicate a memory leak, but then again, it might be just the demo app).

For instnace:
https://github.com/angular-ui/ui-grid/blob/1dc9931f88d7b2183ae25d16df0ab6192cba997d/src/js/core/directives/ui-grid.js#L244
init is being called iniside a `$timeout`.  It then calls `grid.refreshCanvas` which in turn does another `$timeout` (https://github.com/angular-ui/ui-grid/blob/a1e6a04ced73dfc4ef83000f5e44fc6aeec6d7b3/src/js/core/factories/Grid.js#L2121), hence calleing twich to a digest cycle...

As I said, I've managed to reduce this number of cycles with the method above.  Is there a special reason why there are so many `$timeout`s inside the core code?

swalters:
If all the unit tests work and all of the tutorials work, then I see nothing but positives for this change.  You want to submit a PR?

csvan:
Made a PR for this: https://github.com/angular-ui/ui-grid/pull/5829

csvan:
@YonatanKra https://github.com/angular-ui/ui-grid/pull/5829 just got merged, if you are happy with the changes maybe this can be closed.

orneryd:
Thanks for contributing! This repository has been transferred, and we're preparing the new 5.x release. We're closing older issues and pull requests as part of the reset. Please join us on Discord to follow the work or continue the conversation: https://discord.gg/Baz4w8ZWN

PR #5829, opened 2016-11-23
feat(core): Reduce digest triggers when using $timeout

csvan:
Angular's $timeout takes a boolean 3rd parameter which, if `false`, causes the wrapped function to not run inside $apply. Since $timeout is used extensively in core, this parameter has been passed where possible in order to improve performance. Previously, the grid would trigger a large number of digests even when working normally, see issue: https://github.com/angular-ui/ui-grid/issues/5007.

This is very low-hanging fruit. It is possible the same investigation can be applied across the plugins as well.

### DIFF
--- a/src/js/core/services/ui-grid-util.js
+++ b/src/js/core/services/ui-grid-util.js
@@ -831,7 +831,7 @@ module.service('gridUtil', ['$log', '$window', '$document', '$http', '$templateC
         } else {
           s.logWarn('[focus.byId] Element id ' + elementID + ' was not found.');
         }
-      });
+      }, 0, false);
       this.queue.push(promise);
       return promise;
     },
@@ -856,7 +856,7 @@ module.service('gridUtil', ['$log', '$window', '$document', '$http', '$templateC
         if (element){
           element[0].focus();
         }
-      });
+      }, 0, false);
       this.queue.push(promise);
       return promise;
     },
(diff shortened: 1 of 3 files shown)


csvan:
Anyone around to take a look at this?

mportuga:
@MichalJakubeczy @csvan i will try to see if I can include at least part of these changes in the next release of UI-grid. The original version was causing some ui-problems. The most noticeable one that I remember, was a case where I had a toggle-able filter row with a pinned column. In that scenario, for whatever reason, there was either a delay between the pinned and the unpinned filter row display, or it was just partially displaying things. I seem to recall some other UI issues, but none quite as memorable.

MichalJakubeczy:
@mportuga thanks for your answer. Any performance improvements would be awesome as we are struggling with that a bit. Anyway, when is a new release planned would it be angular 1.6.2 compatible?

mportuga:
We are trying to do a minor release soon with the PRs that we merged recently. I think there was [one of them](https://github.com/angular-ui/ui-grid/pull/5949) that adds some catch statements that make it work a bit better with angular 1.6, but it is not necessarily making it angular 1.6 compatible. Right now we are concentrating on performance and memory leaks since those are the big issues that we need fixed. But we would welcome any PR which makes 1.6 support possible.

src/js/core/services/ui-grid-util.js
```diff
@@ -886,8 +886,8 @@ module.service('gridUtil', ['$log', '$window', '$document', '$http', '$templateC
       };
       this._purgeQueue();
       if (aSync){ //Do this asynchronysly
-        var promise = $timeout(focusBySelector);
-        this.queue.push($timeout(focusBySelector));
+        var promise = $timeout(focusBySelector, 0, false);
+        this.queue.push($timeout(focusBySelector), 0, false);
```
samybob1: @csvan there is a mistake here, should be:

```js
this.queue.push($timeout(focusBySelector, 0, false));
```

right?

csvan: Yes that's a mistake

PR #5886, opened 2016-12-12
feat(core): expose GridMenuTemplate

vwongg:
Exposes gridMenuTemplate in gridOptions so that users can alphasort or custom sort their gridMenu items as needed. (may need to splice out the "Columns:" entry by user's custom sort)

**1 test is failing and I'm having issues isolating the root cause. Can you please advise?

### DIFF
--- a/src/js/core/directives/ui-grid-menu.js
+++ b/src/js/core/directives/ui-grid-menu.js
@@ -43,8 +43,15 @@ function ($compile, $timeout, $window, $document, gridUtil, uiGridConstants, i18
     templateUrl: 'ui-grid/uiGridMenu',
     replace: false,
     link: function ($scope, $elm, $attrs, uiGridCtrl) {
-
       $scope.dynamicStyles = '';
+      if (uiGridCtrl && uiGridCtrl.grid && uiGridCtrl.grid.options && uiGridCtrl.grid.options.gridMenuTemplate) {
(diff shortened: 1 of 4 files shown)


vwongg:
Hi @mportuga, opened up a new PR (as the old one closed due to inactivity on my part) to allow custom gridMenu templates but am still running into an issue with 1 test that I'm having difficulty resolving.

src/js/core/directives/ui-grid-menu.js
```diff
@@ -46,6 +46,15 @@ function ($compile, $timeout, $window, $document, gridUtil, uiGridConstants, i18
 
       $scope.dynamicStyles = '';
 
+      if (uiGridCtrl && uiGridCtrl.grid.options.gridMenuTemplate) {
```
mportuga: I think your issue with the unit tests is that your if statement should be more detailed on the checks. Try changing it to the following:

```
if (uiGridCtrl && uiGridCtrl.grid && uiGridCtrl.grid.options && uiGridCtrl.grid.options.gridMenuTemplate) {
```

It may seem a bit overkill, but it will avoid any issues of the style, "Can't find blank of undefined"

mportuga:
@vwongg I gave you a suggestion. I would also like to see you add at least one unit test to check the scenario where a new menu template is used once you manage to get the existing tests running.

Issue #5101, opened 2016-02-11
enableColumnResizing accumulates watchers with each table $digest cycle


pedrobjorn:
Using the enableColumnResizing=true and triggering a new $digest cycle adds 6 watchers every time.

I found the culprit in the displayResizers function.

```
                // Don't append the left resizer if this is the first column or the column to the left of this one has resizing disabled
                if (otherCol && renderContainer.visibleColumnCache.indexOf($scope.col) !== 0 && otherCol.colDef.enableColumnResizing !== false) {
                  var resizerLeft = angular.element(columnResizerElm).clone();
                  resizerLeft.attr('position', 'left');

                  $elm.prepend(resizerLeft);
                  $compile(resizerLeft)($scope);
                }

                // Don't append the right resizer if this column has resizing disabled
                if ($scope.col.colDef.enableColumnResizing !== false) {
                  var resizerRight = angular.element(columnResizerElm).clone();
                  resizerRight.attr('position', 'right');

                  $elm.append(resizerRight);
                  $compile(resizerRight)($scope);
                }
```

It seems the $compile is the problem. ui-grid needs to clean up the $watchers added with every successive $compile that runs every time the $digest takes place.

This happens PER column. I'll be cleaning it on my side, not sure if I should handle this issue, not super experienced with working in such big  open-source projects.

Let me know what you think!

----------------------------------- Potential fix ------------------------------

So the $compile is constantly adding $scopes, and thus $watchers without ever removing them. This seems to happen in other cases as well.

The solution I found is related to this:

http://roubenmeschian.com/rubo/?p=51

so basically:

```
                // Don't append the left resizer if this is the first column or the column to the left of this one has resizing disabled
                if (otherCol && renderContainer.visibleColumnCache.indexOf($scope.col) !== 0 && otherCol.colDef.enableColumnResizing !== false) {
                  var resizerLeft = angular.element(columnResizerElm).clone();
                  resizerLeft.attr('position', 'left');

                 *******NEW CODE*******
                  var newScope = $scope.$new(); // create a new scope from the current scope
                  var content = $compile(resizerRight)(newScope); // compile and link to the new scope
                  $elm.append(content); // add to the DOM

                  // .. then later, we want to update the content of the element ...

                  newScope.$destroy();
                  var newScope2 = $scope.$new(); // create a new scope
                  var content2 = $compile(resizerLeft)(newScope2); // compile and link
                  content.replaceWith(content2); // replace the html content with the new content
[...]

PR #5933, opened 2017-01-10
Fix for Issue: #5101 - enableColumnResizing accumulates watchers with each table $digest cycle

jedancona:
Fixes Issue: enableColumnResizing accumulates watchers with each table $digest cycle #5101.

https://github.com/angular-ui/ui-grid/issues/5101
 
Creates a new scope for each resizer, stores them into an array so that the scopes can be destroyed before resetting the resizers.

### DIFF
--- a/src/features/resize-columns/js/ui-grid-column-resizer.js
+++ b/src/features/resize-columns/js/ui-grid-column-resizer.js
@@ -211,13 +211,9 @@
                 rtlMultiplier = -1;
               }
 
-              var displayResizers = function(){
+              var displayResizers = function () {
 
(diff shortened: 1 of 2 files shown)


mportuga:
@jedancona Can you add unit tests for your removeResizers function?

jedancona:
Added unit testing by calling the destroying the scope and checking the length of watchers.  The previous code leaves a watcher behind.  The new code does not leave a watcher behind.

oricalvo:
What about this PR? It fixes an important scope leak inside the grid which causes significant performance degredation after a time

PR #5951, opened 2017-01-17
Styled scrollbars with jquery-scrollbar

mobichel:
Here is small non obtrusive changes allowing to use "styled" scrollbars with help of  [jquery-scrollbar](https://github.com/gromo/jquery.scrollbar) library.

ui-grid continues to use native scrollbars when jquery-scrollbar is not loaded.

+ Demo

### DIFF
--- a/misc/demo/styled-scrollbars.html
+++ b/misc/demo/styled-scrollbars.html
@@ -1,0 +1,79 @@
+<!DOCTYPE html>
+<html class="no-js" ng-app="test"><!--<![endif]-->
+<head>
+    <meta charset="utf-8">
+    <meta content="IE=edge,chrome=1" http-equiv="X-UA-Compatible">
+    <title></title>
+    <meta content="width=device-width" name="viewport">
+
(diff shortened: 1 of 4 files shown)


mportuga:
I am sorry, but we want to keep this project free of thirdparty dependencies like this. I would accept a PR for supporting adding a different scrolling library to override the regular scroll bar, but not one that overrides the standard scrollbar by default.

mobichel:
@mportuga thanks for the response. 

Please note that standard scrollbar will work unless you have loaded jquery-scrollbar directive in your app.

Will you accept PR with added ability to specify custom uiGridViewport template and small change from the current PR in the ui-grid.html then I can prepare separate plugin for enabling query-scrollbar in the grid. So project will not be tied with extra third party dependencies?

piller187:
Has this feature been added then?

mobichel:
Here is example how to stlye scrollbar without modification in ui-grid: [example](https://jsfiddle.net/mobichel/qpnfL06n/4/)

PR #5964, opened 2017-01-21
Fix #5773 - Wrap $parse expression to in quotes so it can handle spaces

### LINKED ISSUE #5773: Exception thrown editing a string containing a space using flatEntityAccess
We get the issue if the value is set through the underlying model (rather than through the grid directly) to a string containing a space and then click in the cell to edit.

When `getCellDisplayValue` is called `col.cellDisplayGetterCache = $parse(row.entity[col.field] + custom_filter);` throws an exception if the string has a space.

This can be fixed by simply wrapping the expression in quotes:

```
 col.cellDisplayGetterCache = $parse("'" + row.entity[col.field] + "'" + custom_filter);
```

There was a [pr](https://github.com/angular-ui/ui-grid/pull/5426) raised in May for the same issue (I think) - but wasn't merged

MattOakley:
#5773

### DIFF
--- a/src/js/core/factories/Grid.js
+++ b/src/js/core/factories/Grid.js
@@ -1891,7 +1891,7 @@ angular.module('ui.grid')
       if (typeof(row.entity['$$' + col.uid]) !== 'undefined') {
         col.cellDisplayGetterCache = $parse(row.entity['$$' + col.uid].rendered + custom_filter);
       } else if (this.options.flatEntityAccess && typeof(col.field) !== 'undefined') {
-        col.cellDisplayGetterCache = $parse(row.entity[col.field] + custom_filter);
+        col.cellDisplayGetterCache = $parse("'" + row.entity[col.field] + "'" + custom_filter);
       } else {
         col.cellDisplayGetterCache = $parse(row.getEntityQualifiedColField(col) + custom_filter);
       }


src/js/core/factories/Grid.js
```diff
@@ -1891,7 +1891,7 @@ angular.module('ui.grid')
       if (typeof(row.entity['$$' + col.uid]) !== 'undefined') {
         col.cellDisplayGetterCache = $parse(row.entity['$$' + col.uid].rendered + custom_filter);
       } else if (this.options.flatEntityAccess && typeof(col.field) !== 'undefined') {
-        col.cellDisplayGetterCache = $parse(row.entity[col.field] + custom_filter);
+        col.cellDisplayGetterCache = $parse("'" + row.entity[col.field] + "'" + custom_filter);
```
mportuga: Could you not have done this with a ".toString()"?

MattOakley: For my example where it doesn't work, `row.entity[col.field]` is already a string. 

It's just that `$parse` throws a Syntax Error: Token is unexpected, and wrapping the expression in quotes fixes this.

Although thinking about it now, this still doesn't work if `row.entity[col.field]` contains a single quote. So I guess this needs to be escaped.

MattOakley: For my example where it doesn't work, `row.entity[col.field]` is already a string. 

It's just that `$parse` throws a Syntax Error: Token is unexpected, and wrapping the expression in quotes fixes this.

Although thinking about it now, this still doesn't work if `row.entity[col.field]` contains a single quote. So I guess this needs to be escaped.

PR #6011, opened 2017-02-14
Second value of pagination menu fix

kbroncel:
That +1 in second value of pagination menu was unnecesary.


ginz:
You're trying to work around +1 bug in `getLastRowIndex`.

What actually should be fixed is the `getLastRowIndex` method itself, which should return last **index** according to its name:
https://github.com/angular-ui/ui-grid/pull/5998

mportuga:
We chose to merge @ginz's PR instead

PR #6024, opened 2017-02-18
Fix getCellDisplayValue with flatEntityAccess

### LINKED ISSUE #5773: Exception thrown editing a string containing a space using flatEntityAccess

MattOakley:
For flatEntityDataAcess the `col.cellDisplayGetterCache` was always returning the first cached item for the column, and effectively ignoring the row parameter after the `col.cellDisplayGetterCache` had been set.

Closes #5773

### DIFF
--- a/src/js/core/factories/Grid.js
+++ b/src/js/core/factories/Grid.js
@@ -1891,7 +1891,7 @@ angular.module('ui.grid')
       if (typeof(row.entity['$$' + col.uid]) !== 'undefined') {
         col.cellDisplayGetterCache = $parse(row.entity['$$' + col.uid].rendered + custom_filter);
       } else if (this.options.flatEntityAccess && typeof(col.field) !== 'undefined') {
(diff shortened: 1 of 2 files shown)


dlgski:
Can you update unit tests to go with this?

MattOakley:
Added tests

PR #6029, opened 2017-02-20
Just add the missing "gridMenuShowHideColumns" 

TLKG:
/**
* @ngdoc property
* @name gridMenuShowHideColumns
* @propertyOf ui.grid.class:GripOptions
* @descripton boolean. Display all columns in the grid menu at right upper corner, with check mark or cross for users to show or hide the column, default: true
*/


dlgski:
Not sure why this is needed?  The check for this option is looking for !== false, so defaulting to true seems moot

am I missing something?

Please update so it passes CI

mportuga:
Closing due to inactivity.

Issue #6048, opened 2017-03-02
$elm.focus is not a function when we click on check box 


bsivakiran:
Hi part of 4.0.0(also in latest release 4.0.2) release, in order to fix a issue with respect to macOs some changes have been made with respect to focus on check box. 

So the line which is causing the issue is $elm.focus(); because of which check box is not working properly in the ui-grid.
We can see the issue in the tutorials given as well(for cross reference)
http://ui-grid.info/docs/#/tutorial/201_editable(click on any check box and see the console)

The fix would be $elm[0].focus(); which is also mentioned as part of one of the previous issues as below.
https://github.com/angular-ui/ui-grid/issues/5938

I tested it my local and when  i made changes the checkbox is working fine.

Here is the code in ui-grid.js which is causing the issue.

// macOS will blur the checkbox when clicked in Safari and Firefox,
                  // to get around this, we disable the blur handler on mousedown,
                  // and then focus the checkbox and re-enable the blur handler after $timeout
                  $elm.on('mousedown', function(evt) {
                    if ($elm[0].type === 'checkbox') {
                      $elm.off('blur', $scope.stopEdit);
                      $timeout(function() {
                        **$elm.focus();**
                        $elm.on('blur', $scope.stopEdit);
                      });
                    }
                  });

                  $elm.on('blur', $scope.stopEdit);
                });

 Can you please have a look at this and make changes so that Issue can be resolved

Thanks & Regards
sivakiran B

PR #6050, opened 2017-03-02
$elm.focus is not a function when we click on checkbox

bsivakiran:
I am making changes based on the issue created by me as follows.
https://github.com/angular-ui/ui-grid/issues/6048

Please have a look at this and commit the changes if you find it feasible.


dlgski:
same as https://github.com/angular-ui/ui-grid/pull/5939

Issue #6017, opened 2017-02-16
chrome 56 scroll issue on ui-grid 3


HaroldStruwig:
issue is not on ui-grid 4 but only on ui-grid 3

## issue description
grid auto scrolls to the bottom once one starts to scroll down

## reproduce issue:
- use chrome 56
- open pen: [https://codepen.io/anon/pen/JEVYwN](https://codepen.io/anon/pen/JEVYwN)
- start scrolling

## reason for issue:
[https://www.chromestatus.com/feature/5700102471548928](https://www.chromestatus.com/feature/5700102471548928)

## fix for issue:
add CSS
```css
.ui-grid-viewport {
  overflow-anchor: none;
}
```

mloutraris-ge:
We were seeing the same problem. Thanks for finding the fix!

anx-ckreuzberger:
I have the same issue for a couple of days now.

Your CSS Snippet fixed it for me, thank you very much!

prksean:
Awesome! Thanks for the solution. If I came here 6 days earlier, I would have struggled with this so much. Am I supposed to simply add overflow-anchor: none; to already existing ui-grid.css?

sebastianchacko89:
@HaroldStruwig Awesome! Thanks for finding the fix. Im facing the same issue. 
Could you please raise a pull request with your fix, so its available in the next release?

ravivantaku:
Awesome! Thanks for solution.

mportuga:
I release a ui-grid version 3.2.10 with @Joel-Kornbluh's fix

PR #6051, opened 2017-03-02
fix(6017) Fixes ui-grid-viewport scroll issue in chrome v56

Joel-Kornbluh:
Chrome v56 introduced a new feature called "scroll anchoring" which causes the scroll position to update if element content above the current scroll position has changed.
Since ui-grid-viewport updates the margin of the first line whenever a user is scrolling, scroll anchoring will cause it to get into a scroll loop and scroll indefinitely (i.e. scroll to the end of the grid)

This issue is outlined in detail on the ["Scroll Anchoring" feature explainer](https://github.com/WICG/ScrollAnchoring/blob/master/explainer.md#flow-change-suppression):

> If the DOM changes made by a scroll event handler alter the position of the anchor node, we may encounter feedback loops in which scroll anchoring and the scroll event handler trigger each other ad infinitum.

See issue #6017 for more details.

### DIFF
--- a/src/less/body.less
+++ b/src/less/body.less
@@ -17,7 +17,10 @@
   position: relative;
   overflow-y: scroll;
   -webkit-overflow-scrolling: touch;
-
+  // overflow-anchor is a new feature/bug in chrome 56. See https://www.chromestatus.com/feature/5700102471548928
+  // It causes ui-grid-viewport to scroll indefinitely. See https://github.com/angular-ui/ui-grid/issues/6017
+  overflow-anchor: none;
+  
   &:focus {
     outline: none !important;
   }


dlgski:
According to #6017 this is not an issue in ui-grid 4.  This PR would be merged into Ui-Grid 4.0.3 if merged.  Is this necessary??

Joel-Kornbluh:
Shouldn't we at least merge it for <=3.2.9?
This issue will soon start affecting all chrome users and it literally makes scrolling in the grid impossible.

dlgski:
Anything we merge will go into 4.0.3, so the only way anyone can get this change would be by upgrading to 4.0.3 where this is not an issue anyway

mportuga:
@Joel-Kornbluh We are no longer supporting version 3.2.9. Our goal is to keep everyone on the latest version of UI-Grid as much as possible. Our team is too small to maintain older versions of it and guarantee stability.

mportuga:
I reopened this PR into a new branch for release 3.x which will only ever take minor updates like this.

PR #6079, opened 2017-03-15
use $watch instead of setTimeout call every 250 ms in auto-resize directive

asvishnyakov:
Now ui-grid resize occurs a 250 ms after container resize. Because of that, for exampple, I have scroll bar flick when I resize ui-grid container
screenshot 74
If you will use $watch, there will be no delay between container resize and ui-grid resize.

### DIFF
--- a/src/features/auto-resize-grid/js/auto-resize.js
+++ b/src/features/auto-resize-grid/js/auto-resize.js
@@ -15,54 +15,24 @@
   var module = angular.module('ui.grid.autoResize', ['ui.grid']);
 
 
-  module.directive('uiGridAutoResize', ['$timeout', 'gridUtil', function ($timeout, gridUtil) {
+  module.directive('uiGridAutoResize', ['gridUtil', function (gridUtil) {
     return {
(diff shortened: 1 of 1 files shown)


dlgski:
I've been wanting to change this for ages, thanks for the PR.  Did you not need to update any unit tests???

derekm:
Great fix, thanks! This is another one of the performance issues in ui-grid that we've had our eye on!

asvishnyakov:
Revoked. In some sitations it will not work at all, because angular watch will not call function until you will scroll grid, for example. I think, this issue can't be fixed until browers will not provide API to watch width/height changes on element

derekm:
@asvishnyakov -- True... This change would require that height/width only be modified through the two-way bound ui-grid configuration.

derekm:
You could monkey patch container element's `setAttribute` methods to watch for heigh/width changes then pass the call on to the original `setAttribute` that you moved out of the way. This is how lots of features are implemented in GNOME Shell, for example.

PR #6137, opened 2017-03-29
ScrollToIfNeeded

Andey-d:
ScrollToIfNeeded incorrectry calculate area where scroll is not nedded 
in case when header stretched by content or decreased by css
Header and footer height removed from calculations of non-scroll area calculations


mportuga:
Please fix unit tests.

mportuga:
Closing due to lack of answer

PR #6211, opened 2017-05-12
Fixed grid height growing indefinitely with auto-resize.

### LINKED ISSUE #6197: Grid height growing indefinitely with auto-resize
A grid with height:100% and auto-resize (slowly) grows indefinitely in height if the header has a non-integer height. This can happen e.g. by setting the line-height to a fractional number. See this [example](http://plnkr.co/edit/la5e5mAdNJBzERlqnKoj?p=preview). The problem doesn't occur with Chrome though, but it can be seen in both Firefox and Edge.

As far as I can see the problem is cause by rounding the headerHeight down to the next integer [here](https://github.com/angular-ui/ui-grid/blob/ba77c8750431236ff4f729fe8a77eb44cbbfc9b7/src/js/core/factories/Grid.js#L2193). Since this value is later used to calculate the total height again, the auto-resize underestimates the height of the grid and increases it slightly. The easiest fix would probably be to remove the parseInt, but I'm not sure if this could cause some new problems.

Iteem:
Fixes #6197 by removing parseInt from the calculation of the header height.

### DIFF
--- a/src/js/core/factories/Grid.js
+++ b/src/js/core/factories/Grid.js
@@ -2190,7 +2190,7 @@ angular.module('ui.grid')
           }
 
           if (container.header) {
-            var headerHeight = container.headerHeight = getHeight(container.headerHeight, parseInt(gridUtil.outerElementHeight(container.header), 10));
+            var headerHeight = container.headerHeight = getHeight(container.headerHeight, gridUtil.outerElementHeight(container.header));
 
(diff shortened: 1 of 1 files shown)


sjoerd222888:
What should the unit-test test in this case?

mportuga:
My suggestion would be to call that function with a number that normally would be parsed and return a bad value and make sure that headerHeight is the height that you would expect rather than something else.

PR #6241, opened 2017-06-01
Prevent scrollEvent when scroll percentage is greater then 1.

rdkleine:
In the scenario where cell edit and cell navigation is enabled the scrollPercentage will be greater then 1. But this only happens in the bottom row. The check prevents scrolling when this happens.

### DIFF
--- a/src/js/core/factories/Grid.js
+++ b/src/js/core/factories/Grid.js
@@ -2394,7 +2394,9 @@ angular.module('ui.grid')
 
           // Turn the scroll position into a percentage and make it an argument for a scroll event
           percentage = scrollPixels / scrollLength;
-          scrollEvent.y = { percentage: percentage  };
+          if (percentage <= 1) {
+            scrollEvent.y = { percentage: percentage  };
+          }
         }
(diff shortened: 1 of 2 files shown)


rdkleine:
Will add some unittests.

mportuga:
Weird, the karma tests are passing for me. I will take a look when I have a chance.

rdkleine:
Updated my fork. Tests are running again.

rdkleine:
Added coverage for the changes in grid.js.

rdkleine:
Cool thanks!

RobPierce:
This seems to have broken scrollTo - debugging through, when I add a new row to the bottom of my grid, which is off the screen, the scrollToIfNecessary calculates the percentage to scroll to be 1.09 in this particular case and due to this change it means it won't scroll as:

if (percentage <= 1)

is not true. I'm not sure why this change was made but it doesn't seem that the logic is correct - if a row is off the screen then we want to scroll to it; either the logic to determine the percentage value is incorrect or the logic that decides not to scroll if percentage > 1 is wrong.

I'll open a new issue for this and link it.

PR #6244, opened 2017-06-02
fix #5760: only show ui-grid-scrollbar-placeholder if there's a horizontal scrollbar in another container

### LINKED ISSUE #5760: ui-grid-scrollbar-placeholder blocks half the height of last ui-grid-cell
Hi all, 

I have picked up this issue on multiple browsers (Chrome, IE, Firefox). I am on v3.2.9. I have scoured the web for a solution, so I truly hope this is not a duplicate issue.

In the bottom left corner of `ui-grid` there is a `ui-grid-scrollbar-placeholder` that obscures half the last tick box/cell (image below). 

ui-grid-scrollbar-placeholder_1

Here is the `<div>` markup from the developer tools:

ui-grid-scrollbar-placeholder_markup

I have also noticed that one can scroll an extra couple pixels on this tick box/cell column, that throws out the alignment (image below).

ui-grid-scrollbar-placeholder_2

I would appreciate it if you could show me how to fix this. Thank you very much.

jeffantosh:
Basically adding a check to see if the sum of all of the widths containers in the grid exceeds the viewport size. If true and if horizontal scrolling is not disabled, then show the ui-grid-scrollbar-placeholder div.

### DIFF
--- a/src/js/core/factories/GridRenderContainer.js
+++ b/src/js/core/factories/GridRenderContainer.js
@@ -743,7 +743,11 @@ angular.module('ui.grid')
   };
 
   GridRenderContainer.prototype.needsHScrollbarPlaceholder = function () {
-    return this.grid.options.enableHorizontalScrollbar && !this.hasHScrollbar && !this.grid.disableScrolling;
+    var bodyRenderContainerWidth = this.grid.renderContainers.body ? this.grid.renderContainers.body.canvasWidth : 0;
+    var leftRenderContainerWidth = this.grid.renderContainers.left ? this.grid.renderContainers.left.canvasWidth : 0;
+    var rightRenderContainerWidth = this.grid.renderContainers.right ? this.grid.renderContainers.right.canvasWidth : 0;
+    var fullGridWidth = bodyRenderContainerWidth + leftRenderContainerWidth + rightRenderContainerWidth;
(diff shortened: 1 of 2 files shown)


jeffantosh:
I don't know enough about the inner workings of ui grid to make these unit tests work. I also was unable to get the unit tests to even execute on my machine.


PhantomJS 2.1.1 (Linux 0.0.0) GridRenderContainer factory needsHScrollbarPlaceholder  should return false if the container does not have the horizontal scrollbar enabled FAILED
	Expected 0 to be false.
	/home/travis/build/angular-ui/ui-grid/test/unit/core/factories/GridRenderContainer.spec.js:242:53
PhantomJS 2.1.1 (Linux 0.0.0) GridRenderContainer factory needsHScrollbarPlaceholder  should return true if another container does have a horizontal scrollbar and the current container does not have the horizontal scrollbar enabled AND the total canvas width is greater than the grid width FAILED
	Expected false to be true.
	/home/travis/build/angular-ui/ui-grid/test/unit/core/factories/GridRenderContainer.spec.js:246:53

mportuga:
Closing since we are unable to fix the unit tests at this time.

piller187:
Just wondering what's the deal with this issue as it's still an issue it seems.

PR #6254, opened 2017-06-07
Corrected spelling for German description

classix-as:
So far it read "... von ... Einträge". However, von implies the noun to be of dative case. This again is "Einträgen" instead of "Einträge".

See:
https://en.wiktionary.org/wiki/von#Preposition
https://en.wiktionary.org/wiki/Einträgen#German


mportuga:
This was already fixed by PR #6262

PR #6260, opened 2017-06-12
feat(cellnav): output aria text for cells that come from selection feature

adamascg:
Added feature to output aria-text for cells that come from selection feature ('selection ticks'). 
I added e2e test, but for unkown reason it causes next test suite to fail. It looks as if the next test fired before the page was actually loaded. Therefore I ignored the test. Does anybody have a clue, how can I make it work?
Besides I manually tested the feature with screen readers: NVDA and JAWS.

### DIFF
--- a/misc/tutorial/210_selection.ngdoc
+++ b/misc/tutorial/210_selection.ngdoc
@@ -57,7 +57,7 @@ auto-selects the first row once the data is loaded.
 
 <example module="app">
   <file name="app.js">
-    var app = angular.module('app', ['ngTouch', 'ui.grid', 'ui.grid.selection']);
+    var app = angular.module('app', ['ngTouch', 'ui.grid', 'ui.grid.selection', 'ui.grid.cellNav', 'ngAria']);
 
     app.controller('MainCtrl', ['$scope', '$http', '$log', '$timeout', 'uiGridConstants', function ($scope, $http, $log, $timeout, uiGridConstants) {
       $scope.gridOptions = {
@@ -192,7 +192,7 @@ auto-selects the first row once the data is loaded.
       <button type="button" class="btn btn-success" ng-click="toggleFullRowSelection()">Toggle full row selection</button>
       <br/>
 
-      <div ui-grid="gridOptions" ui-grid-selection class="grid"></div>
+      <div id="grid1Selection" ui-grid="gridOptions" ui-grid-selection ui-grid-cellnav class="grid"></div>
     </div>
     <div ng-controller="SecondCtrl">
       Single selection grid without rowHeader, and allowing rowSelection to be toggled on and off dynamically:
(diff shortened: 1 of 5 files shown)


mportuga:
I have faced the same issue with e2e tests before. I am still trying to figure out what is going on myself. So, I wouldn't concern myself too much with that.

misc/tutorial/210_selection.ngdoc
```diff
@@ -207,4 +207,26 @@ auto-selects the first row once the data is loaded.
       height: 400px;
     }
   </file>
+  <file name="scenario.js">
+    var gridTestUtils = require('../../test/e2e/gridTestUtils.spec.js');
+    var GridObjectTest = require('../../test/e2e/gridObjectTestUtils.spec.js');
+    var grid1Selection = new GridObjectTest('grid1Selection');
+
+    describe('210 tutorial', function() {
+      xit('should output aria text for cells that come from selection feature', function () {
```
mportuga: You probably need a done here.
`it('should output aria text for cells that come from selection feature', function (done) {
 +          var focuser = element(by.css('.ui-grid-focuser'));
 +          grid1Selection.selectRow(1).then(function () {
 +            return gridTestUtils.click(focuser);
 +          })
 +          .then(function () {
 +             expect(element(by.css('.ui-grid-a11y-ariascreenreader-speakable.ui-grid-offscreen')).getInnerHtml()).toEqual('Row selected ');
 +             return focuser.sendKeys(protractor.Key.ARROW_DOWN);
 +          })
 +          .then(function () {
 +             expect(element(by.css('.ui-grid-a11y-ariascreenreader-speakable.ui-grid-offscreen')).getInnerHtml()).toEqual('Row not selected ');
 +             done();
 +          });
 +      });`

adamascg:
I tried your solution, but it didn't help. However, I manually changed files generated by grunt (in .tmp\doc-scenarios\doc-tutorial.spec.js). What solved my problem was change in beforeEach in test suites that failed
from:
`    beforeEach(function() {
      browser.driver.get("http://127.0.0.1:9999/docs/#/tutorial/210_selection");
    });`
to:
`    beforeEach(function() {
      browser.get("http://127.0.0.1:9999/docs/#/tutorial/210_selection");
    });`
[browser.get docs](http://www.protractortest.org/#/api?view=ProtractorBrowser.prototype.get). The difference is that browser.get waits for angular to load.
I am not sure, if that is a correct solution and which build task must be changed to modify the output.

mportuga:
If that works, then I am all for it. I say make the change.

adamascg:
The problem is, that \*.spec.js code is generated by external library- grunt-ngdoc. Even worse is that ui-grid uses custom fork, that is pretty much dead (last change over year ago): https://github.com/c0bra/grunt-ngdocs/tree/master. I have tried switching to official version, but ui-grid relies heavily on changes from forked repo (\*.spec.js files aren't generated). I don't think, I can solve that issue in simple way.

mportuga:
Thank you for your analysis @adamascg. I will try to see if switching over to the official version is possible. I myself am not a fan of the fact that we have a custom version, but that was done prior to me joining the team, but I will make it my next action item now.

mportuga:
But for  now, I will approve and merge your PR

PR #6262, opened 2017-06-12
Fixed i18n typo in German language

bitflower:
The plural form of "items" (= "Einträge") in german is "Einträgen" of used like this in the pagination:
"1 of 30 items" => "1 von 20 Einträge**n**"


bitflower:
Oh great, two acurate Germans at work ;-)

PR #6269, opened 2017-06-16
Accessibility and role=contentinfo

classix-as:
Our accessibility expert commented that it would be confusing for paging to have the role 'contentinfo' as this is the place for the footer with copyright and such:
https://www.w3.org/TR/wai-aria/roles#contentinfo

I found that this role was introduced in ee04132 for the sake of accessibility. However, I think it to be overeager and counterproductive.

### DIFF
--- a/misc/tutorial/210_selection.ngdoc
+++ b/misc/tutorial/210_selection.ngdoc
@@ -57,7 +57,7 @@ auto-selects the first row once the data is loaded.
 
 <example module="app">
   <file name="app.js">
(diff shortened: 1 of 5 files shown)


mportuga:
@classix-as What role do you suggest we add there instead? Or do you believe no role should be there?

classix-as:
@mportuga As I have done in the commit, I would suggest removing the role because it is misleading. However, I am open to other suggestions.

classix-as:
This change ist part of #6341.

PR #6271, opened 2017-06-16
fix #6270: Update getCellDisplayValue to manage special case with flatEntityAccess

### LINKED ISSUE #6270: getCellDisplayValue broken if column name special char and flatEntityAccess true
I've come across issues to display column with special chars.
I fixed it by adding `flatEntityAccess: true` option, now it display fine.

But I'm also using the getCellDisplayValue for external search and this one is broken because of this :
```
        col.cellDisplayGetterCache = $parse('entity.' + col.field + custom_filter);
```

I guess it should be managed that way instead, as we know that its a flat object.
  ```
        col.cellDisplayGetterCache = $parse('entity[\'' + col.field + '\']' + custom_filter);
```

Here is a plunker with the issue : http://plnkr.co/edit/M1Kb3kZZu2aeZewTDbTt?p=preview

rloisel-forcity:
Context
- When using columnName with special chars and grid has option flatEntityAccess: true, the function getCellDisplayValue is broken.

What changed
- Replace the string handed to $parse to use square brackets notation instead of dot

Related to https://github.com/angular-ui/ui-grid/issues/6270

### DIFF
--- a/src/js/core/factories/Grid.js
+++ b/src/js/core/factories/Grid.js
@@ -1899,7 +1899,7 @@ angular.module('ui.grid')
       if (typeof(row.entity['$$' + col.uid]) !== 'undefined') {
         col.cellDisplayGetterCache = $parse(row.entity['$$' + col.uid].rendered + custom_filter);
       } else if (this.options.flatEntityAccess && typeof(col.field) !== 'undefined') {
-        col.cellDisplayGetterCache = $parse('entity.' + col.field + custom_filter);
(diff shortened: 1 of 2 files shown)


mportuga:
@rloisel-forcity Would you mind adding a unit tests that tries to call that function with a column name with special chars. That way, we can make sure that this never breaks in that same way again.

rloisel-forcity:
@mportuga added a test for this

PR #6284, opened 2017-06-21
Fix #4949 - pinnedRight adding an extra blank column

### LINKED ISSUE #4949: pinnedRight adding an extra blank column
Hi,

Setting `pinnedRight` to true is adding an extra blank column to its left.
Here's a plunker for reference: http://plnkr.co/edit/h3TzNPIhFp6BsV4C3PHE?p=preview

Thanks,
Anirudha

ronansi:
Using @mani301 approach


mportuga:
Could you add a unit test that checks that the drawn width of the column is question fills the available space?

PR #6338, opened 2017-07-26
feat(cellnav): output aria text for cells that come from selection feature - works on IE with JAWS screen reader

adamascg:
This is an extension PR to https://github.com/angular-ui/ui-grid/pull/6260 There is a need of another PR concerning this, because what I have already done worked in Chrome, but not on Internet Explorer with JAWS screen reader. Therefore I ask to merge this PR into master branch as well. The fix is very simple, but it changes the semantics of aria implementation a little bit. You have used the aria-live attribute which should already work as a live region regardless of the role of the element, but unfortunately the JAWS screen reader doesn't see those live messages on Internet Explorer. It does see it when the role is set to 'alert'. It's not very pretty, but it's the only way it works on IE.

### DIFF
--- a/src/features/cellnav/js/cellnav.js
+++ b/src/features/cellnav/js/cellnav.js
@@ -792,7 +792,7 @@
                                            'id="' + grid.id +'-aria-speakable" ' +
                                            'class="ui-grid-a11y-ariascreenreader-speakable ui-grid-offscreen" ' +
                                            'aria-live="assertive" ' +
-                                           'role="region" ' +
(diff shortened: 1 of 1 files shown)


hohwille:
Interesting is also that the code already contains a comment right above the changed code pointing to this link:
http://www.paciellogroup.com/blog/2012/06/html5-accessibility-chops-aria-rolealert-browser-support/

This linked article proposes to use the `alert` role so IMHO this PR fixes something that should have already been intended initially.

PR #6341, opened 2017-07-26
ARIA-Roles

classix-as:
While there was no progress on #6269 I went on to talk to our accessiblity expert to fix the role-problems we have encountered so far.

To our understanding contentinfo and menubar should be removed without replacement as they are not applicable.

We went on to make the selection checkboxes accessible and present the grid as one grid similar to an example we have referenced in commit 066ba4c.

I understand that basic support has been added in 377485a by @JLLeitschuh. I see this as an expansion of this.

### DIFF
--- a/src/features/pagination/templates/pagination.html
+++ b/src/features/pagination/templates/pagination.html
@@ -6,11 +6,9 @@
     role="navigation"
     class="ui-grid-pager-container">
     <div
-      role="menubar"
       class="ui-grid-pager-control">
       <button
         type="button"
-        role="menuitem"
         class="ui-grid-pager-first"
         ui-grid-one-bind-title="aria.pageToFirst"
         ui-grid-one-bind-aria-label="aria.pageToFirst"
@@ -25,7 +23,6 @@
       </button>
       <button
         type="button"
-        role="menuitem"
         class="ui-grid-pager-previous"
         ui-grid-one-bind-title="aria.pageBack"
         ui-grid-one-bind-aria-label="aria.pageBack"
@@ -53,7 +50,6 @@
       </span>
       <button
         type="button"
-        role="menuitem"
         class="ui-grid-pager-next"
         ui-grid-one-bind-title="aria.pageForward"
         ui-grid-one-bind-aria-label="aria.pageForward"
@@ -63,7 +59,6 @@
       </button>
       <button
         type="button"
-        role="menuitem"
         class="ui-grid-pager-last"
         ui-grid-one-bind-title="aria.pageToLast"
         ui-grid-one-bind-aria-label="aria.pageToLast"
--- a/src/features/pagination/templates/pagination.html
+++ b/src/features/pagination/templates/pagination.html
@@ -1,5 +1,4 @@
 <div
-  role="contentinfo"
   class="ui-grid-pager-panel"
   ui-grid-pager
   ng-show="grid.options.enablePaginationControls">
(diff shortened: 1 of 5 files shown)


src/features/pagination/templates/pagination.html
```diff
@@ -1,17 +1,14 @@
 <div
-  role="contentinfo"
   class="ui-grid-pager-panel"
   ui-grid-pager
   ng-show="grid.options.enablePaginationControls">
   <div
     role="navigation"
     class="ui-grid-pager-container">
     <div
-      role="menubar"
```
JLLeitschuh: Isn't the bottom bar a `menubar`? This is more for my own learning more than anything else.

src/features/pagination/templates/pagination.html
```diff
@@ -64,7 +59,6 @@
       </button>
       <button
         type="button"
-        role="menuitem"
```
JLLeitschuh: Can you explain why these are wrong?

JLLeitschuh:
Hi, I'm the original author of the accessibility PR. Thank you for expanding upon my work!
I'm no longer an active maintainer of ui-grid as I'm now at a different job working on a server backend (no more a11y).

I'm not really in a place where I can approve this PR as I can't test it with a screen reader, perhaps @dlgski or @mportuga can have a look.

Can you explain your removal and why they aren't applicable?

src/features/pagination/templates/pagination.html
classix-as: @JLLeitschuh the description of menubar referenced in the commit message (https://www.w3.org/TR/wai-aria-practices-1.1/#keyboard-interaction-11) states that a menubar has to be navigable by the arrow keys right and left.
> * Right Arrow:
>   * When focus is in a menubar, moves focus to the next item, optionally wrapping from the last to the first.

This is something, one could fix in the future. However, I think that the role of menubar is not meant to be used here.

1. Items in a menubar are assumed to have submenus, although I am not sure whether this is really compulsory:
> 2. All items in the menubar have child submenus that contain multiple vertically arranged items.

2. A menubar can hold Buttons, Checkboxes or Radiobuttons but no input field:
> * The items contained in a menu are child elements of the containing menu or menubar and have any of the following roles:
>   * menuitem
>   * menuitemcheckbox
>   * menuitemradio

A menuitem again is defined to be a command, which is a button or the like and _no_ input field:
https://www.w3.org/TR/wai-aria-1.1/#command

An example of a menubar can be found here:
https://www.w3.org/TR/wai-aria-practices/examples/menubar/menubar-1/menubar-1.html

classix-as:
@JLLeitschuh I have explain, why I think menubar should be removed in https://github.com/angular-ui/ui-grid/pull/6341#discussion_r130228477.

The role contentinfo should be removed as this landmark role is to be assigned to the footer of a site. As ui-grid is meant to be used as a component of a site, it is inappropriate to declare any part of it to be the sites footer.

From https://www.w3.org/TR/wai-aria-1.1/#contentinfo:
> Examples of information included in this region of the page are copyrights and links to privacy statements.

From https://www.w3.org/TR/wai-aria-practices-1.1/examples/landmarks/contentinfo.html
> A contentinfo landmark is a way to identify common information at the bottom of each page within a website, typically called the "footer" of the page, including information such as copyrights and links to privacy and accessibility statements.

> * Each page may have one contentinfo landmark.

> * If a page includes more than one contentinfo landmark, each should have a unique label.

PR #6414, opened 2017-10-05
feat(headerMenu): add an option to show/hide column menu buttons

classix-od:
Added an option to show/hide column menu buttons. The option can be set for specific columns or in general.


src/js/core/directives/ui-grid-column-menu.js
```diff
@@ -401,8 +401,11 @@ function ($timeout, gridUtil, uiGridConstants, uiGridColumnMenuService, $documen
           $scope.hideMenu( true );
 
           if ($scope.col) {
-            //Focus on the menu button
-            gridUtil.focus.bySelector($document, '.ui-grid-header-cell.' + $scope.col.getColClass()+ ' .ui-grid-column-menu-button', $scope.col.grid, false);
+            //Focus on the menu button if this is shown, otherwise focus the column header
+            var elementToFocus = '.ui-grid-header-cell.' + $scope.col.getColClass();
+            elementToFocus += ($scope.grid.options.showColumnMenuButtons && $scope.col.colDef.showColumnMenuButton !== false)?
```
mportuga: This is redundant. if $scope.grid.options.showColumnMenuButton returns truthy, then it is obviously different than false. You only need one of these checks.

src/templates/ui-grid/uiGridHeaderCell.html
```diff
@@ -37,7 +37,7 @@
     tabindex="0"
     ui-grid-one-bind-id-grid="col.uid + '-menu-button'"
     class="ui-grid-column-menu-button"
-    ng-if="grid.options.enableColumnMenus && !col.isRowHeader  && col.colDef.enableColumnMenu !== false"
+    ng-if="grid.options.enableColumnMenus && grid.options.showColumnMenuButtons && !col.isRowHeader  && col.colDef.enableColumnMenu !== false && col.colDef.showColumnMenuButton !== false"
```
mportuga: Same comment as above

src/js/core/directives/ui-grid-column-menu.js
classix-od: They are not the same. Please look carefully, the first is general and the second is column specific.

src/templates/ui-grid/uiGridHeaderCell.html
classix-od: Same comment as above

mportuga:
Closing due to age and conflicts.

PR #6415, opened 2017-10-06
feat(headerKeyDown): add options to specify SPACE/RETURN key down fun…

classix-od:
…ction on a column header.

columnHeaderKeyDownAction: This option specifies the action to be performed when the space bar or the return key is pressed on a column header. This option takes a value of the enum uiGridConstants.keyboardActions. This option can be overwritten by the options columnHeaderSpaceAction and columnHeaderReturnAction.

Available options are:
     * - `uiGridConstants.keyboardActions.NO_ACTION` - does nothing
     * - `uiGridConstants.keyboardActions.SHOW_MENU` - shows the column menu
     * - `uiGridConstants.keyboardActions.DO_SORT` - fires the click event on the column header

### DIFF
--- a/test/unit/core/factories/GridOptions.spec.js
+++ b/test/unit/core/factories/GridOptions.spec.js
@@ -52,7 +52,8 @@ describe('GridOptions factory', function () {
         gridFooterTemplate: 'ui-grid/ui-grid-grid-footer',
         rowTemplate: 'ui-grid/ui-grid-row',
         gridMenuTemplate: 'ui-grid/uiGridMenu',
-        appScopeProvider: null
+        appScopeProvider: null,
+        columnHeaderKeyDownAction: 2
       });
     });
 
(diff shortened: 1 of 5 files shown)


test/unit/core/factories/GridOptions.spec.js
```diff
@@ -143,7 +147,10 @@ describe('GridOptions factory', function () {
         rowTemplate: 'testRow',
         gridMenuTemplate: 'testGridMenu',
         extraOption: 'testExtraOption',
-        appScopeProvider : 'anotherRef'
+        appScopeProvider : 'anotherRef',
+        columnHeaderKeyDownAction: 0,
```
mportuga: Rather than using 0, 1 and 2, wouldn't it be easier to read this if you used the constants?

mportuga:
Closing due to age.

PR #6429, opened 2017-10-19
feat(sorting): add option to specify the keys to add/remove sort columns

classix-od:
With the option keyToAddSortColumns, we can enable adding/removing columns to/from the collection of sorted columns not only using the shift key, because the shift key has selection side effects in IE, but also the CTRL key, if this option enables that.


mportuga:
Closing due to age and conflicts.

PR #6435, opened 2017-10-26
added new event - before row selection - to the register api events

NigmaX:
I have added a new event to the **ui-grid-selection** feature, 
for the following reason:
I needed a way to cancel the toggle of a row selection. and there was no event i could catch that will stop the toggle after i choose a row. 
so what i did is adding a new event called, "beforeSelectionChanged" to the "onRegisterApi" events.
and it has the same parameters as the "rowSelectionChanged" with a callback function.
the callback function will be invoked from the place this event was registered, and if the parameter from this callback is true, it will stop the toggle.

**Example:**
```
onRegisterApi: function (gridApi) {
                gridApi.selection.on.beforeSelectionChanged($scope, function (row, event, callback) {
                    //return true to disable row selection and false tow enable row selection
                    callback(true);
                });
```

so basically if callback function is invoked with true parameter the 
**"rowSelectionChanged"** and **"rowSelectionChangedBatch"**
will be **skipped** and the new row will not be selected.

Tested with angularJS 1.4.8

Thank you.
Gabriel Sternberg.


mportuga:
@NigmaX Please resolve merge conflict

NigmaX:
sure, conflict resolved. 
thnx

mportuga:
Closing due to age.

PR #6453, opened 2017-11-08
feat(columnHeader): the sort arrow and priority do not disappear

classix-od:
When resizing a column, the sort arrow and priority do not disappear and are attached intelligently to the right side of the column

### DIFF
--- a/src/js/core/directives/ui-grid-header-cell.js
+++ b/src/js/core/directives/ui-grid-header-cell.js
@@ -37,6 +37,7 @@
             var uiGridCtrl = controllers[0];
             var renderContainerCtrl = controllers[1];
 
+
             $scope.i18n = {
               headerCell: i18nService.getSafeText('headerCell'),
               sort: i18nService.getSafeText('sort')
@@ -61,6 +62,8 @@
             };
 
             $scope.grid = uiGridCtrl.grid;
+
+            $scope.showColMenuButton = $scope.grid.options.enableColumnMenus && !$scope.col.isRowHeader  && $scope.col.colDef.enableColumnMenu !== false;
 
             $scope.renderContainer = uiGridCtrl.grid.renderContainers[renderContainerCtrl.containerId];
 
--- a/src/less/header.less
+++ b/src/less/header.less
@@ -56,6 +56,40 @@
   position: relative
 }
 
+.ui-grid-header-cell-label {
+  max-width: 100%;
+  position: relative;
+  display: inline-block;
+
+  &.with-button-with-sort {
+    max-width: calc(~"100% - 30px") !important; 
+  }
+  
+  &.with-button-without-sort {
+    max-width: calc(~"100% - 22px") !important; 
+  }
+  
+  &.without-button-with-sort {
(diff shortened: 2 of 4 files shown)


src/less/header.less
```diff
@@ -56,6 +56,40 @@
...
+}
+
+.ui-grid-header-cell-sort-info {
+  position: absolute;
+  top: 0px;
+  right: -19px;
+}
+
+.ui-grid-header-cell-primary-focus {
+  text-overflow: unset !important;
```
mportuga: Why do you need these important tags?

src/js/core/directives/ui-grid-header-cell.js
```diff
@@ -62,6 +63,8 @@
 
             $scope.grid = uiGridCtrl.grid;
 
+            $scope.showColMenuButton = $scope.grid.options.enableColumnMenus && !$scope.col.isRowHeader  && $scope.col.colDef.enableColumnMenu !== false;
```
mportuga: It is redundant to check if $scope.grid.options.enableColumnMenus is truthy and then immediately check if it is different than false. I suggesting doing just the false check

src/templates/ui-grid/uiGridHeaderCell.html
```diff
@@ -10,35 +10,41 @@
     class="ui-grid-cell-contents ui-grid-header-cell-primary-focus"
     col-index="renderIndex"
     title="TOOLTIP">
-    <span
+    <div
```
mportuga: What is the extra div for?

src/js/core/directives/ui-grid-header-cell.js
classix-od: $scope.grid.options.enableColumnMenus and $scope.col.colDef.enableColumnMenu are not the same. The first is general and the second is column specific.

src/templates/ui-grid/uiGridHeaderCell.html
classix-od: To enable the ellipses text overflow, the inner div should have overflow set to hidden. At the same time, the sort information should not be hidden, although it is part of the overflowing content. To solve this conflict I introduced the outer div.

src/less/header.less
classix-od: From a theoretical point of view, these tags are not needed. I can't remember why it didn't work without them. When I have enough time, I will test it again and remove these tags if they are not necessary.

src/js/core/directives/ui-grid-header-cell.js
mportuga: whoops, I didn't see the s. good point

src/templates/ui-grid/uiGridHeaderCell.html
mportuga: thanks for the explanation.

src/js/core/directives/ui-grid-header-cell.js
mportuga: Would it be possible to add some unit tests for this line?

mportuga:
Closing due to age.

PR #6456, opened 2017-11-09
[BUG FIX] registerRowsProcessor priority was missplaced

NigmaX:
Find:

```
// Reset all rows to visible initially
          grid.registerRowsProcessor(function allRowsVisible(rows) {
            rows.forEach(function (row) {
              row.evaluateRowVisibility( true );
            }, 50);

            return rows;
          });
```

Replace:

```
// Reset all rows to visible initially
          grid.registerRowsProcessor(function allRowsVisible(rows) {
            rows.forEach(function (row) {
              row.evaluateRowVisibility( true );
            });

            return rows;
          }, 50);

```


mportuga:
Please add some unit tests and resolve merge conflict before I can review this.

NigmaX:
**registerRowsProcessor** gets **2 parameters**, 
first param is a **function**, second param is the **priority**,
the bug was that no second parameter was introduced.
and by mistake added as a second parameter for the **foreach** function. 
it makes no sense.


```
rows.forEach(function (row) {
              row.evaluateRowVisibility( true );
            }, --->>> 50 <<<---);
```
iv'e just moved it to the correct possition. 
the way it should work.
thnx

mportuga:
Closing due to age.

PR #6461, opened 2017-11-10
fix(uiGridHeader) stop preserving explicit header heights 

### LINKED ISSUE #3705: Filters disappear if ui-grid-selection and dynamic filters in columns are active
When use dynamic filters in columns and `ui-grid-selection` then filters in column disappear. 
Noticed that if remove `padding=5px` from `.ui-grid-cell-contents` then filter appear, but are not aligned (see below).
grid

To fix it only helps to set `enableRowHeaderSelection: false`, then first column with checkboxes is hidden and the problem doesn't occur. 
grid2

AgDude:
Stop preserving explicit header heights when styles are rebuilt. The buildStyles function fails to expand header height when it was previously set explicitly, and if explicit definition is needed it will be reset to the same value below.

fixes #3705

### DIFF
--- a/src/js/core/factories/Grid.js
+++ b/src/js/core/factories/Grid.js
@@ -2148,8 +2148,8 @@ angular.module('ui.grid')
         }
 
         if (container.header || container.headerCanvas) {
-          container.explicitHeaderHeight = container.explicitHeaderHeight || null;
-          container.explicitHeaderCanvasHeight = container.explicitHeaderCanvasHeight || null;
+          container.explicitHeaderHeight = null;
+          container.explicitHeaderCanvasHeight = null;
 
           containerHeadersToRecalc.push(container);
         }


mportuga:
Won't making this change bring back a few issues? I mean, I looked at the git blame history of those lines and when you did that change originally it fixed 3 issues.
https://github.com/angular-ui/ui-grid/commit/65ad61f4439983179be4d44bac3a6cc9151ac094

AgDude:
I went back and looked at the issues and linked plunkers I addressed with this original change in 2015 and I could not reproduce them. I think the underlying issue was that styles were not being recalculated as often as they should have. I suspect the changes in https://github.com/angular-ui/ui-grid/commit/553d72bc4153dc5945978c9f7ece8bcb0040cccc fixed that.

If anyone can reproduce those earlier issues I will certainly take another look.

PR #6471, opened 2017-11-19
Develop

cybermerlin:
what do y'all think about combining the branches?


mportuga:
@cybermerlin I like the idea, but could we do it in smaller steps. It is hard to review 60 files all at once, if you could break it down into groups of 10 files for example, that probably would help. You still need to fix unit tests on your branch as well prior to us doing this merge.

Issue #6481, opened 2017-11-27
v.4.0.11 AutoResize prevents page scrolling


classix-od:
The new version (4.0.11) raised a general problem in all browsers, which had occurred only in IE before. If the grid has the property autoResize set, with some border or padding, and has neither horizontal nor vertical scrollbars, the page (the browser) is prevented from scrolling down when the mouse is on the grid. 

To fix this problem we need to use gridUtil.elementHeight($elm) instead of $elm[0].clientHeight (for width calculations gridUtil.elementWidth($elm)) which respects borders and padding. This would fix the problem for Chrome, Firefox and Edge. For IE we need either to use the JQuery functions outerHeight and outerWidth (which are not allowed in ui-grid according to the JQuery-freeness) or we need to augment the function elementWidth and elementHeight to respect the special case of IE.

I use the JQuery functions as a solution. 

One another question:
Isn't it better for the performance to use a timeout of 250 ms for autoResize than using a watcher? A watcher would be triggered more often than a timeout function and every time the width and height need to be calculated and compared, which is not quite lite for browsers like IE.

mportuga:
https://github.com/angular-ui/ui-grid/pull/6490

PR #6490, opened 2017-11-30
fix(uiGridAutoResize): Changed [0].clientHeight to gridUtil.elementHe…

tfilo:
I made change based on this issue: https://github.com/angular-ui/ui-grid/issues/6481
I changed `$elm[0].clientHeight` to `gridUtil.elementHeight($elm)` and `$elm[0].clientHeight` to `gridUtil.elementWidth($elm)`.
I made another change too. I merged two watchers into one watchGroup. And I changed how `uiGridCtrl.grid.refresh();` is called. Looks like it needs to be called inside 
```
$scope.$apply(function(){
  uiGridCtrl.grid.refresh();
});
```
because without that it cause some problems when resizing grid with many columns. All tests still pass without problem.

### DIFF
--- a/src/features/auto-resize-grid/js/auto-resize.js
+++ b/src/features/auto-resize-grid/js/auto-resize.js
@@ -26,7 +26,7 @@
           function() {
             return gridUtil.elementHeight($elm);
           }], function(newValues, oldValues, scope) {
-          if (newValues.toString() !== oldValues.toString()) {
+          if (!angular.equals(newValues, oldValues)) {
             uiGridCtrl.grid.gridWidth = newValues[0];
             uiGridCtrl.grid.gridHeight = newValues[1];
             setTimeout(function(){
--- a/src/features/auto-resize-grid/js/auto-resize.js
+++ b/src/features/auto-resize-grid/js/auto-resize.js
@@ -19,21 +19,30 @@
       require: 'uiGrid',
       scope: false,
       link: function($scope, $elm, $attrs, uiGridCtrl) {
+        var timeout = null;
+
+        var debounce = function(width, height) {
+          if (timeout !== null) {
+            clearTimeout(timeout);
+          }
+          timeout = setTimeout(function() {
+            uiGridCtrl.grid.gridWidth = width;
+            uiGridCtrl.grid.gridHeight = height;
+            uiGridCtrl.grid.refresh();
+            timeout = null;
+          }, 400);
+        };
+
         $scope.$watchGroup([
           function() {
             return gridUtil.elementWidth($elm);
           },
           function() {
             return gridUtil.elementHeight($elm);
-          }], function(newValues, oldValues, scope) {
+          }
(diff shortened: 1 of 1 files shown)


src/features/auto-resize-grid/js/auto-resize.js
```diff
@@ -19,25 +19,23 @@
...
-            uiGridCtrl.grid.gridWidth = newVal;
-            uiGridCtrl.grid.refresh();
+        $scope.$watchGroup([
+          function() {
+            return gridUtil.elementWidth($elm);
+          },
+          function() {
+            return gridUtil.elementHeight($elm);
+          }], function(newValues, oldValues, scope) {
+          if (newValues.toString() !== oldValues.toString()) {
```
mportuga: Why do you need to use the toString prior to comparing?

src/features/auto-resize-grid/js/auto-resize.js
```diff
@@ -19,25 +19,23 @@
...
+            uiGridCtrl.grid.gridWidth = newValues[0];
+            uiGridCtrl.grid.gridHeight = newValues[1];
+            setTimeout(function(){
```
mportuga: Whys is a timeout needed here? Specially if the time is set to 0.

tfilo: Here they talk about it more. https://stackoverflow.com/questions/22346990/why-is-using-ifscope-phase-scope-apply-an-anti-pattern

If I called scope apply directly, it offten happens that I saw error in javascript console because apply was called in middle of another digest cycle. Which lead to error in javascript. So when I was looking to other solutions. This looked like best one.

src/features/auto-resize-grid/js/auto-resize.js
tfilo: I used toString because it's simples method how to compare two arrays in javascript. Method toString is supported in all browsers. And I can't directly compare two arrays because newValues === oldValues will always return false because it will compare if instance of array is same. It will not check if content is the same. But when I convert arrays to string, if the content is same, I will get two identical strings and is easy to compare two strings.

tfilo: Now when I think about it I changed it to angular.equals. It's better way in angular to compare arrays. Thanks for questioning.

tfilo:
After I consider all negatives of implementation with watcher I made other changes to prevent negative impact on performance. This latest implementation check width and height with watcher but instead of changing grid size directly it creates timeout. And if some another change occur before this timeout, it will set new timeout and cancel old one. So changes to grid will only happen after you stop resizing window.

So it should be even better with performance like original solution with interval 250ms. I know that there was even suggestions to replace watcher with event but I don't know how to do that. If you listen only event on windows resize, you will miss resize of grid parent container when some content of page change. For example hiding side menu. And I didn't find any solution how to make onResize based on parent container. So it is on you if you accept my solution. If someone can implement better solution I will glad to learn something new.

PR #6498, opened 2017-12-04
test(ui-grid-column-menu): Fix test issue with angular 1.6.7

Teamop:
Hi @mportuga 
this issue is related to this commit in 1.6.5, [$timeout/$interval: do not trigger a digest on cancel](https://github.com/angular/angular.js/commit/a222d0b452622624dc498ef0b9d3c43647fd4fbc), so there is no need to use $timeout.flush() for another digest. So as per test under 1.6, the test can also pass, so I just deleted.


mportuga:
Thank you!

Issue #6371, opened 2017-08-22
space to right of rows when no width is given


Blindsyde:
I have a site where we show many uiGrid's. Previous to the latest update we always had scrollbars showing whether or not there was anything to scroll which worked well, we also never set widths on any columns, we just let uiGrid keep each column at what seemed to be an automatic width.

Today I have upgraded to 4.0.6 and now the scroll bars are missing, and adding `enableHortizontalScrollbar: uiGridConstants.scrollbars.ALWAYS` or `enableVerticalScrollbar: uiGridConstants.scrollbars.ALWAYS`  doesn't seem to make a difference nor does it add the scroll bars back in, also the more important part is every one of the grids now has white space to the right of all the rows (where the scrollbar would normally be)

![image](https://user-images.githubusercontent.com/8527885/29589520-c66209a6-8752-11e7-849e-b34bb2de8373.png)

My fiddle below will both illustrate the white space to right of rows and that the scrollbars setting doesn't actually give scrollbars:
https://jsfiddle.net/Blindsyde/rtj8vxwk/

RobPierce:
I've found the reason for the space to the right of columns, it's to do with the fix for #5981 because column widths are now rounded to integers and any remaining space is only distributed between asterisk column widths, not percentage column widths so if you have some or all percentage column widths and no asterisk column widths, you will have some width left over. I'm about to submit a PR to remedy this.

Blindsyde:
Great thank you!

mportuga:
https://github.com/angular-ui/ui-grid/pull/6513

PR #6512, opened 2017-12-21
Fix bug where space is created to the right of the last column

RobPierce:
The change to using integer column widths means that we sometimes have some grid width left over where column widths are rounded down for asterisk or percentage widths, see #6371. The logic that then re-allocates the remaining widths currently only does this over columns with asterisk widths so I've updated the logic to check if there are any asterisk columns and, if not, check for any percentage columns as a fall back. Then use one or other to allocate the remaining width over.

### DIFF
--- a/src/js/core/factories/GridRenderContainer.js
+++ b/src/js/core/factories/GridRenderContainer.js
@@ -609,7 +609,8 @@ angular.module('ui.grid')
         asteriskNum = 0,
         usedWidthSum = 0,
         ret = '',
-        pinRightColumn = false;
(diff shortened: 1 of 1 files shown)


RobPierce:
Failing unit tests so will update and submit a new PR - will update code so that it only uses percentage column widths as a fallback for asterisk widths if ALL column widths are percentage and the percentages add up to 100%. This will leave scenarios where the percentages are not meant to fill the width, or where some columns are fixed width, so that we don't try to add remaining grid width to the percentage columns.

PR #6553, opened 2018-01-30
Excel export

monster910:
Exporter was incorrectly exporting width of selection column. This is fixed.

Added example with date column and using cellFilters on export


src/features/exporter/js/exporter.js
```diff
@@ -1535,7 +1535,7 @@
             // The standard column width in Microsoft Excel 2000 is 8.43 characters based on fixed-width Courier font
             // Width of 10 in excel is 75 pixels
             var colWidths = [];
-            var startDataIndex = grid.treeBase ? grid.treeBase.numberLevels : 0;
+            var startDataIndex = grid.treeBase ? grid.treeBase.numberLevels : (grid.enableRowSelection !== false ? 1 : 0);
```
mportuga: Why are you checking for different than false rather than simply check for a truthy value?

```
grid.enableRowSelection? 1 : 0
```

misc/tutorial/410_excel_exporting_data_and_header.ngdoc
```diff
@@ -51,13 +51,35 @@ In this example we use the native grid menu buttons, and we show the pdf, csv an
...
+
       $scope.gridOptions = {
         columnDefs: [
           { field: 'name' },
           { field: 'gender', visible: false},
           { field: 'company' },
           { field: 'member' },
-          { field: 'total' }
+          { field: 'total' },
+          { field: 'updatedDate', displayName: 'Date', cellFilter: 'date:\'yyyy-MM-dd\'', width: 150,
```
mportuga: What do these updates have to do with the bug that you are trying to fix?

mportuga:
Is there a bug logged for the issue that you are attempting to fix? Also, would it be at all possible for you to add an unit test for you change? If not, I understand, but I would prefer it if we had one.

src/features/exporter/js/exporter.js
monster910: because it may be null. I was following the pattern already in the code.

monster910: line 361 selection.js

          gridOptions.enableRowSelection = gridOptions.enableRowSelection !== false;

mportuga: Well, that line in 361, ensures that enableRowSelection is a boolean from that point on. I don't believe there is ever a way to get into this function without running through 361 first. But if you feel strongly about keeping it this way, I won't hold your pull request because of it.

mportuga: Never mind, I decided to separate your doc changes from your bug fix by doing a cherry pick and creating a new PR for the fix alone. I will add unit tests and merge it on its own, Then we can address the doc changes together.

PR #6567, opened 2018-02-04
Ensure that getting selected rows returns row entities, excluding the grouping entities

yonatankarimish:
The _getSelectedRows(grid)_ function in uiGridSelectionService returns all selected rows. However, the _getSelectedRows()_ function being registered into the grid api returns the selected row's entities.

Because grouping header entities are now also selected, this could lead to confusion. 

**getSelectedGridRows()** will return all selected rows -for  **grouping headers** + standard rows
**getSelectedRows()** will return selected row entities - for **standard rows only**

### DIFF
--- a/src/features/selection/js/selection.js
+++ b/src/features/selection/js/selection.js
@@ -301,6 +301,8 @@
                 getSelectedRows: function () {
                   return service.getSelectedRows(grid).map(function (gridRow) {
                     return gridRow.entity;
+                  }).filter(function (entity) {
+                    return entity.hasOwnProperty('$$hashKey');
                   });
                 },
                 /**


src/features/selection/js/selection.js
```diff
@@ -254,7 +254,7 @@
                  * @description Selects all visible rows.  Does nothing if multiSelect = false
                  * @param {Event} event object if raised from an event
                  */
-                selectAllVisibleRows: function (event) {
+                selectAllVisibleRows: function (evt) {
```
falsyvalues: This change is not necessary at all.

yonatankarimish: Perhaps, but it would better comply with your coding style.

yonatankarimish: It is purely cosmetic

falsyvalues: The change should be as small as possible, thats all :smile:

yonatankarimish:
will open a new pull request, without the cosmetic changes

PR #6568, opened 2018-02-06
Ensure that getting selected rows returns row entities, excluding the grouping entities

yonatankarimish:
same as #6567, but only with the relevant changes

### DIFF
--- a/src/features/selection/js/selection.js
+++ b/src/features/selection/js/selection.js
@@ -301,6 +301,8 @@
                 getSelectedRows: function () {
                   return service.getSelectedRows(grid).map(function (gridRow) {
                     return gridRow.entity;
+                  }).filter(function (entity) {
+                    return entity.hasOwnProperty('$$hashKey');
                   });
                 },
                 /**


src/features/selection/js/selection.js
```diff
@@ -301,6 +301,8 @@
                 getSelectedRows: function () {
                   return service.getSelectedRows(grid).map(function (gridRow) {
                     return gridRow.entity;
+                  }).filter(function (entity) {
+                    return entity.hasOwnProperty('$$hashKey');
```
mportuga: Is there a better way to filter these other than relying on the hashKey. I few like this might now be trustworthy in the long run. If angular ever removes this property, this feature would be broken.

yonatankarimish: 1) The Angular team recently wrote (https://blog.angular.io/stable-angularjs-and-long-term-support-7e077635ee9c) that they will not be making any breaking changes after 1/7/2018

2) AngularJS relies in the $$hashKey property for object comparison, it is highly unlikely they remove this property as it will force them to perform deep comparisons between objects

3) The $$hashKey property is the only property assigned to all entity objects. If you have any better ideas I am open to suggestions

mportuga: Well, what I am concerned about is the update to Angular 5. This is something that might end up breaking and need more refactoring to keep functional once we are ready to upgrade. And although I understand that there will be no more breaking changes to angularJS. We need to plan for the future and strive for as much reusability when we upgrade as possible.

yonatankarimish: While your concern addresses point 1), it ignores points 2), and 3). It would be the same as preparing for breaking changes in a core library such as jquery (or jqlite): Could a breaking change technically happen ? Yes. How probable is such a change? Almost not at all.

Take into note that without this pull request, if a user selects a grouping header together with a standard row, he will receive two completely different row entities. It is much more probable this will cause bugs for end users of THIS library than any future change to angular 5.

mportuga: Good point, I will consider this and decide how to handle this. Perhaps, we can make this a toggleable feature, that is turned off by default, in order to prevent bugs for end users of UI-Grid, but still provide a way for users such as yourself which do not want group entities.

Issue #3978, opened 2015-07-14
Grid menu animation issue in Firefox 


rquiroz6:
When closing the grid menu in Firefox 39.0, the menu's fly-out animation gets 'stuck' above the grid, as seen in this screenshot:

![firefoxuigridmenu](https://cloud.githubusercontent.com/assets/11983747/8679752/6acecac0-2a11-11e5-897a-7dde38ec8bfd.PNG)

This issue occurs with the column menus as well.

fberthel:
I can confirm this bug for Firefox 40.0.2

tringuyenminh92:
i got this issue in FF ESR 38.2.0

susharma-equinix:
Iam also facing the same issue. Can some one please look into this at the earliest. It would be very grate full if i can get an work around for the same temporarily.

hohwille:
http://ui-grid.info/docs/#/tutorial/121_grid_menu
Still buggy in FF 41.0.1

hohwille:
http://ui-grid.info/
also the regular sort menus do not work with FF. Only working with Chorme.
IE also works but without annimaiton.

Better disable animation gimmicks in FF than having this broken stuff.

SoundLogic:
any easy fix in the meantime is to do the following in your app.js file (requires ngAnimate module):

```javascript
    app.config([
        '$animateProvider', function($animateProvider) {
            $animateProvider.classNameFilter(/^((?!(ui-grid-menu)).)*$/);
        }
    ]);
```

veegandhi:
Run into the same issue , @SoundLogic solution works

fberthel:
@SoundLogic  works like a charm, thanks :)

Issue #3921, opened 2015-07-02
Undesired grid menu animation change


brandonr1:
Basically the same as the column menu bug (#3436), but seems to be a grid menu problem as well.

When switching from 3.0.0-rc.20 to 3.0.0-rc.22, I noticed a change in the grid menu animation. It used to be nice and subtle, but it now comes down from the top of the page.

On my pages the problem doesn't happen when I use RC20 css with RC22 code.  When I use RC22 css the behavior presents itself. 

You can see it on the grid menu demo page (using Chrome ).
http://ui-grid.info/docs/#/tutorial/121_grid_menu

BTW, ui-grid is awesome!

PaulL1:
Ah, that's a good finding.  I had assumed it was a css change, but I thought I did a bisect and didn't find it. Perhaps I did the bisect wrong.  If it's css it'd be much easier to fix.

Any chance you could work out which line in the css does it?  I'm guessing it's a position absolute or something like that somewhere on the menu css.

brandonr1:
I'll double check to verify css is the most likely cause and dig in next week when I get back to work.  I hope it is on the easier side :)

anilnatha:
I just began adding animation to my application and prior to adding the ngAnimate module to my app, I had no issues with the menu; however, after adding ngAnimate, I'm experiencing the same issue that brandonr1 is reporting running 3.0.0-rc.22 as well.

anilnatha:
Just a side note: I really wish there was a way to disable animation for the grid, but because of this issue, I have to remove the animation module from my application entirely and forgo adding animation to the rest of my app because of the menu's jumpy behavior, unless there is a workaround that I'm unaware of.

Allowing developers to choose if they want the grid to use animation or not would be a step in the right direction, rather than forcing grid features to always be animated if the animation module is added to an app.

mikyladewitt:
+1 I have the same problem. It would be great to be able to disable this for developers that use ng-animate, but don't want to have the column animation.

antoinebrault:
gridOptions.onRegisterApi = function (gridApi) {
    $animate.enabled(gridApi.grid.element, false);
}

anilnatha:
Thanks @antoinebrault! I'll give that a try!

Maus3rVonDutch:
I think it's just a css issue. Not sure if this is the appropiate fix, but in the mean time you could apply this css to fix it as an alternative to disabled the animation altogether as mentioned above:
```
[...]

PR #6588, opened 2018-03-03
fix(menus): Fix menu positioning/animation

### LINKED ISSUE #3436: Undesired column menu animation change
When switching from 3.0.0-rc.20 to 3.0.0-rc.21, I noticed a change in the column menu animation. It used to be nice and subtle, but it now comes down from the top of the page.
Hopefully this is just a bug and not an intentional change.

You can see it on the main demo page (using Chrome if that matters).
http://ui-grid.info/

### LINKED ISSUE #6587: If column/grid menu is wider than 170px, or padding is greater than 10px, the menu is shifted when it is first opened.
See an example here: https://plnkr.co/edit/Sr45hW2zhXcbPxvzEBTB?p=preview.

Open a column menu - it is initially positioned incorrectly.

I've increased the width of the menu (as well as the padding) to exaggerate the issue. Internally, repositionMenu is called again and then the menu is positioned correctly. Subsequent menu opens are positioned correctly.

Will submit a PR shortly with a fix.

caseyjhol:
Current behavior: https://plnkr.co/edit/Sr45hW2zhXcbPxvzEBTB?p=preview
New behavior: https://plnkr.co/edit/mSNHjvbSVCRklVrawk6m?p=preview

Since v3.0.0-rc.21, the column/grid menu animation slides down from
above the grid, but it is visible the entire time. This causes the
animation to appear clunky. This has been resolved by setting `overflow:
hidden` on the menu.

Fixes #3436, #3921, #3978.

Showing/hiding the menu can seem slow. The menu animation has been sped
up as well, and when it is hidden, the animation is now twice as fast.

Previously, to calculate the menu position, the width of the menu needed
to be calculated, which meant calling `repositionMenu` twice - once when
first opening the menu, and again after the menu is visible to allow the
width of the menu to be determined. This also meant there was always a
"shift" when the menu was first opened (as a default width of 170px was
used until the actual width of the menu could be determined. This
implements the CSS trick `right: 100%` so the width is not needed when
determining the menu's position, meaning the menu can be opened much
more smoothly and without needing to call `repositionMenu` twice.

Fixes #6587.

`lastMenuWidth` has been removed as well, as it is no longer needed.

Looking forward to hearing your feedback.

### DIFF
--- a/src/js/core/directives/ui-grid-column-menu.js
+++ b/src/js/core/directives/ui-grid-column-menu.js
@@ -274,17 +274,11 @@ function ( i18nService, uiGridConstants, gridUtil ) {
 
       var containerScrollLeft = renderContainerElm.querySelectorAll('.ui-grid-viewport')[0].scrollLeft;
 
-      // default value the last width for _this_ column, otherwise last width for _any_ column, otherwise default to 170
-      var myWidth = column.lastMenuWidth ? column.lastMenuWidth : ( $scope.lastMenuWidth ? $scope.lastMenuWidth : 170);
       var paddingRight = column.lastMenuPaddingRight ? column.lastMenuPaddingRight : ( $scope.lastMenuPaddingRight ? $scope.lastMenuPaddingRight : 10);
 
       if ( menu.length !== 0 ){
         var mid = menu[0].querySelectorAll('.ui-grid-menu-mid');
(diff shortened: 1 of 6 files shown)


caseyjhol:
@mportuga Thanks for merging this! I just realized, though, that I got the animation classes mixed up (it's 0.04s to hide and 0.02s to show - I meant it to be the opposite). Obviously not a huge difference maker, but let me know if you want me to submit another PR.

PR #6629, opened 2018-03-22
fix(ui-grid-column-menu.js): Added keyboard navigation to column menu

### LINKED ISSUE #5075: Column Menu does not retain focus when tabbing through menu options - Accessibility buug
When using tab to navigate through the grid, when you open up the column menu the focus should remain inside the menu until the user performs an action.  Instead, the menu shows and odd X icon to close the menu, then lets you tab through the menu options.  After tabbing to the last item another tab takes to random locations on your page.    Also, there is a "hidden" Close menu that shows up only when tabbing through the menu options.

Proper behavior:  User should be able to open the menu with a keyboard action (clicking on enter when focus is on the menu toggle icon).  User should be able to tab through options and, when the user hits the last menu option, focus is retained in the menu and focus will move to the first item in the menu.  This should work in both ways when tab or shift-tab is pressed.

DanielDiTommaso:
Provided keydown handlers for uiGridColumnMenu so you can tab-cycle through the menu items
correctly. Escape also now closes an open menu.

fix #5075

### DIFF
--- a/src/js/core/directives/ui-grid-column-menu.js
+++ b/src/js/core/directives/ui-grid-column-menu.js
@@ -414,6 +414,9 @@ function ($timeout, gridUtil, uiGridConstants, uiGridColumnMenuService, $documen
           delete $scope.colElementPosition;
           delete $scope.columnElement;
         });
+        $timeout(function() {
+          addKeydownHandlers(getVisibleMenuItems());
+        }, 0, false);
       });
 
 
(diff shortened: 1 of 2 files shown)


src/js/core/directives/ui-grid-column-menu.js
```diff
@@ -435,6 +445,44 @@ function ($timeout, gridUtil, uiGridConstants, uiGridColumnMenuService, $documen
...
+          };
+
+          menuItems = menu.querySelectorAll('.ui-grid-menu-item:not(.ng-hide)');
+          angular.forEach(menuItems, function filterVisibleItems(item) {
+            if (item.offsetParent !== null) {
+              this.push(item);
+            }
+          }, visibleMenuItems);
+
+          if (visibleMenuItems.length) {
```
mportuga: What happens if there is only a single column header menu item? For example the company column of the first grid in the plunker bellow:
http://plnkr.co/edit/?p=preview

It can be seen in the 102 Sorting tutorial.

src/js/core/directives/ui-grid-column-menu.js
```diff
@@ -435,6 +445,54 @@ function ($timeout, gridUtil, uiGridConstants, uiGridColumnMenuService, $documen
...
+            if (visibleMenuItems.length === 1) {
+              visibleMenuItems[0].onkeydown = function singleItemHandler(event) {
+                circularFocusHandler(event, true);
```
mportuga: Can't you just call event.preventDefault directly here?

mportuga: Never mind, I talked to you in person and see the if statement.

PR #6651, opened 2018-04-02
fix(ui-grid-menu.js): Give bySelector correct parent.

### LINKED ISSUE #5206: UI Grid Logs Warn messages when hiding a column (angular.js:12330 Trying to focus on an element that isn't an element)
Issue is more a "heads-up" than an actual bug as I believe functionality of the grid is unaffected.

However, if you open up the following Tutorial (http://ui-grid.info/docs/#/tutorial/102_sorting) and open up the browser console and hide any of the columns in any of the sample tables, you get the *warning* message below:
image

You can see 4 times ui-grid is trying to focus on a null or non-angular element.

I've seen this "warning" occur on all UI-grids in all the tutorials (as well as my app which is where I first found it and thought perhaps I was doing something "inappropriate").

defields923:
Verify that the element sent to the bySelector method is correct so that querySelector will never return null.

fix #5206


src/js/core/directives/ui-grid-menu.js
```diff
@@ -294,7 +294,13 @@ function ($compile, $timeout, $window, $document, gridUtil, uiGridConstants, i18
                 $scope.$emit('hide-menu');
               } else {
                 // Maintain focus on the selected item
-                gridUtil.focus.bySelector(angular.element($event.target.parentElement), 'button[type=button]', true);
+                var correctParent = $event.target.parentElement;
+
+                if (angular.element($event.target)[0].nodeName === 'I') {
```
mportuga: Could you add a comment above the if statement explaining what is 'I' in this case? This is mainly because this line is not self explanatory and I don't want us to lose track as to what it was added for.

defields923: Updated a while ago. Hope that's what you were looking for, and glad to be of help.

PR #6668, opened 2018-04-11
fix(Grid.js): ScrollIfNecessary determines top boundary wrong

### LINKED ISSUE #6653: 4.4.5: Infinite loop with cellnav and enableCellEditOnFocus
In ui-grid 4.4.5, with cell-nav and cell-edit features enabled with enableCellEditOnFocus set to true. 
Navigating to the last row in the grid gets stuck in an infinite loop (cpu in the tab goes often to 100%) or becomes un-editable

I am able to reproduce the error in the following plunker for both chrome and firefox.
Steps to reproduce: expand all rows, scroll to the bottom, select a cell in the bottom row
http://plnkr.co/edit/cVIvvBzZ6clPMEZzWAjr?p=preview

- [x] This is a bug report, not a question on how to use the grid.  
Use [Stack Overflow](http://stackoverflow.com/questions/tagged/angular-ui-grid) or [Gitter](https://gitter.im/angular-ui/ui-grid) for questions.
- [x] You have searched the open issues to see if this bug has been filed before.  We don't want duplicate issues.
- [x] You have reviewed the tutorials and documentation to ensure you are using the grid correctly. [ui-grid.info](http://ui-grid.info/docs/#/tutorial)
- [x] You are using the latest version of the grid. Older versions are not patched or supported.
- [x] You have provided steps to recreate your bug.  A plunkr is even better.

m4m4m4:
Adding headerHeight to topBoundary calculation makes it always larger than pixelsToSeeRow causing an infinite loop.

fixes #6653

### DIFF
--- a/src/js/core/factories/Grid.js
+++ b/src/js/core/factories/Grid.js
@@ -2392,7 +2392,8 @@ angular.module('ui.grid')
       /*-- Get the top, left, right, and bottom "scrolled" edges of the grid --*/
 
       // The top boundary is the current Y scroll position PLUS the header height, because the header can obscure rows when the grid is scrolled downwards
-      var topBound = self.renderContainers.body.prevScrollTop + self.headerHeight;
+      // var topBound = self.renderContainers.body.prevScrollTop + self.headerHeight;
+      var topBound = self.renderContainers.body.prevScrollTop;
 
       // Don't the let top boundary be less than 0
       topBound = (topBound < 0) ? 0 : topBound;


mportuga:
Why did you close this?

m4m4m4:
While it solved editable cells where enableCellEditOnFocus was turned on, it broke top scrolling for non-editable cells. I'll deep dive into this again today and see if I can find a solution.

PR #6669, opened 2018-04-12
fix(Grid.js): ScrollIfNecessary does not account for scrollWidth correctly

### LINKED ISSUE #6653: 4.4.5: Infinite loop with cellnav and enableCellEditOnFocus

m4m4m4:
Since scrollIfNecessary is called multiple times when enableCellEditOnFocus is true we need to make sure the scrollbarWidth and footerHeight is accounted for to not cause a loop.

fixes #6653

### DIFF
--- a/src/js/core/factories/Grid.js
+++ b/src/js/core/factories/Grid.js
@@ -2444,6 +2444,11 @@ angular.module('ui.grid')
           //   to get the full position we need
           scrollPixels = self.renderContainers.body.prevScrollTop - (topBound - pixelsToSeeRow);
 
+          //Since scrollIfNecessary is called multiple times when enableCellEditOnFocus is true we need to make sure the scrollbarWidth and footerHeight is accounted for to not cause a loop.
+          if (gridCol.colDef.enableCellEditOnFocus === true) {
+            scrollPixels = scrollPixels - self.footerHeight - self.scrollbarWidth;
+          }
+
           scrollEvent.y = getScrollY(scrollPixels, scrollLength, self.renderContainers.body.prevScrolltopPercentage);
         }
         // Otherwise if the scroll position we need to see the row is MORE than the bottom boundary, i.e. obscured below the bottom of the self...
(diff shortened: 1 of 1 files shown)


src/js/core/factories/Grid.js
```diff
@@ -2444,6 +2444,11 @@ angular.module('ui.grid')
           //   to get the full position we need
           scrollPixels = self.renderContainers.body.prevScrollTop - (topBound - pixelsToSeeRow);
 
+          //Since scrollIfNecessary is called multiple times when enableCellEditOnFocus is true we need to make sure the scrollbarWidth and footerHeight is accounted for to not cause a loop.
+          if (gridCol.colDef.enableCellEditOnFocus === true) {
+            scrollPixels = scrollPixels - self.footerHeight - self.scrollbarWidth;
```
mportuga: Is there a reason why we wouldn't want to check this every time? Why are we only checking for the footer and scroll bar when enableCellEditOnFocus is true?

m4m4m4: The downside that I have found in leaving it on for everything is that it will scroll a bit too much for the grids not using enableCellEditOnFocus. It makes navigating by keys not appear as smooth and that might make it confusing.

mportuga: Hmm, I see, makes sense. Also, as a follow up, could you add a null or undefined check for gridCol instead of assuming it will be there. I ask, because i noticed that a few lines now, we are checking for that and I don't want this to ever blow up on users of this function. I believe refactoring it to the following should work:

`if (gridCol && gridCol.colDef && gridCol.colDef.enableCellEditOnFocus) {`

There should be no need to explicitly compare enableCellEditOnFocus to true either.

PR #6680, opened 2018-04-20
Excel export to support tree and standard selection column

monster910:
This will fix the the other case for standard selections. It also adds a scaling factor that can be used to scale the column widths on excel because the user may vary fonts which will allow ultimate flexibility to set column widths on excel


mportuga:
@monster910 Would you be able to squash all your commits and resolve the merge conflict? If not, I might try the cherry-pick solution that I did last time and merge it as a separate pull request when I have a little more time.

monster910:
Ok. Will make another request.

Issue #6666, opened 2018-04-10
v4.4.6 Export to CSV not working or showing errors.


brandonpearce2:
The  `exporterMenuCsv` option fails in version 4.4.6. I reverted back to 4.4.5  as a quick fix.

There are no error messages or network calls.

mportuga:
At a quick glance, the only change in 4.4.5 to export feature was the getDisplayName changes that I did. So it stands to reason that it is causing the issue somehow:
https://github.com/angular-ui/ui-grid/commit/4c63239186c73fd3ba1723481ed67618897e9fe4

brandonpearce2:
Just to be clear, 4.4.6 is broken. 4.4.5 is good.

mportuga:
Sorry, I made a typo. Yeah, that is what I meant. 4.4.6 had a change for exporter to change the header names that I did. I suspect that would be the cause. Although, could you describe the issue a little more. I tried clicking the export data as CSV on this plunker and it worked for me: http://plnkr.co/edit/T3IeR2EIuZAj7RlnwsKo?p=preview

mportuga:
I was also able to see it working with the plunker above on Mac Safari and Mac Firefox. What browser and operating system are you currently using?

brandonpearce2:
I've been going through your plnkr, and adding my options, but so far it is working. I'm using windows/chrome. Basically I have an angularjs app that is reading from a REST call and displaying charts and tables. A dashboard essentially.

mportuga:
I see. Well, once you manage to break my plunker let me know. I am curious to see what causes the issue.

mportuga:
@brandonpearce2 Any luck reproducing it in my plunker?

brandonpearce2:
I've spent an hour plugging in my options and templates and a few other fields and haven't been able to find the culprit. Unfortunately it's quite a beast of a dashboard I have so it could be 'n' number of things.

jprr773:
After migrating ui-gris from version v3.2.1 to v4.4.6 , the export menu button is visible but when I click Export to PDF or CSV nothing happens.  3rd party JS libs csv.js and pdfmake.js wre e also updated. Chrome console does not fire any JS error. As recommended by @mportuga , I downgrade to version using 4.4.5.
Thank you for your time

brandonpearce2:
@mportuga  I found it. I'm surprised I didn't try this earlier. I added a hidden field to the grid, and that is what's stopping the download.
[...]

PR #6683, opened 2018-04-20
fix(Export): Eliminate selection column. Add export scale factor for …

monster910:
…excel and fonts

This will fix when a selection column is present vs a tree selection column and not put the drawnwidth column into an array of column definitions for exporting resulting in all columns being the wrong size.

It also introduces a scaling factor used for excel to scale the width for fonts or implementation specific exports.

@mportuga  - I think there is a bug in your change to defaultExporterFieldCallback on 4/5/18. The issue is that when you have a cellFilter like 'cellFilter: currency' the raw data is converted in line 984 using the cellFilter. So, in this case the data goes from 100 to '$100'. It then goes through the exporterFieldFormatCallback and applies the filter for a second time. '$100' | currency become an empty string resulting in all empties in the export. I think you need to review that check in as it broke our code

### DIFF
--- a/misc/tutorial/410_excel_exporting_data_and_header.ngdoc
+++ b/misc/tutorial/410_excel_exporting_data_and_header.ngdoc
@@ -39,12 +39,6 @@ Additionally, information about the Excel Builder may be found at https://github
 
 If the example below we show how to make a custom header based on a callback method and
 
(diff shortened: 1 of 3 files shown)


mportuga:
@monster910 thanks! I just figured that out in issue #6666. Looking at it tonight.

PR #6690, opened 2018-04-23
fix(Grid.js) Vertical scroll calculates height wrong with enableHorizontalScrollbar: NEVER

m4m4m4:
Use scrollbarHeight instead of scrollbarWidth for vertical scroll calculations. scrollbarHeight has the value 0 when enableHorizontalScrollbar is set to NEVER.

Round calculated boundary values as computed style may have decimal number which will not match pixelsToSeeRow

Related to: #6653


m4m4m4:
Found two more scenarios where a loop could occur, seem to be specifically firefox.
It's easy to reproduce on firefox with this plunkr:
http://plnkr.co/edit/4RnRdl5pSlmdKhO3K1pa?p=preview

PR #6693, opened 2018-04-24
fix(selection.js): Raise rowSelectionChangedBatch after toggling selectAll flag

### LINKED ISSUE #5411: selectAllRows() method raises rowSelectionChangedBatch prior to setting the selectAll flag
It appears that the _selectAllRows_ method is raising _rowSelectionChangedBatch_ prior to setting the _selection.selectAll_ flag.  Due to this, you're unable to determine if all rows were selected via _getSelectAllState()_.  

**selectAllRows**

`service.decideRaiseSelectionBatchEvent(grid, changedRows, evt);`
`grid.selection.selectAll = true;`

However, this doesn't appear to be the case with the service.toggleRowSelection function that's called when a single row has been selected.

**toggleRowSelection**

`selectedRows = service.getSelectedRows(grid);`
`grid.selection.selectAll = grid.rows.length === selectedRows.length;`
`grid.api.selection.raise.rowSelectionChanged(row, evt);`

From looking at the code a bit more, it's actually the _selectAllVisibleRows_ method that get's called when the select all checkbox is clicked and this appears to have the same issue as _selectAllRows_.

**selectAllVisibleRows**

`service.decideRaiseSelectionBatchEvent( grid, changedRows, evt );`
`grid.selection.selectAll = true;`

After updating a forked copy of the code, I noticed that the _clearSelectedRows_ method is invoking the _decideRaiseSelectionBatchEvent_ prior to setting the _selectAll_ and causes _getSelectAllState_ to return true when it should be false inside _rowSelectionChangedBatch_ 

**clearSelectedRows**

`service.decideRaiseSelectionBatchEvent( grid, changedRows, evt );`
`grid.selection.selectAll = false;`
`grid.selection.selectedCount = 0;`

mportuga:
Change order of toggle to match toggleRowSelection function

fix #5411

### DIFF
--- a/src/features/selection/js/selection.js
+++ b/src/features/selection/js/selection.js
@@ -243,8 +243,8 @@
                         service.decideRaiseSelectionEvent(grid, row, changedRows, evt);
                       }
                     });
-                    service.decideRaiseSelectionBatchEvent(grid, changedRows, evt);
                     grid.selection.selectAll = true;
+                    service.decideRaiseSelectionBatchEvent(grid, changedRows, evt);
                   }
                 },
                 /**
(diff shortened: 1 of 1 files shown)


mportuga:
When the next version is released. It will also update the CHANGELOG.md file. I am planning on doing a version update this Friday though if you want a timeline.

PR #6722, opened 2018-05-11
fix(ui-grid-util.js): Replace angular.uppercase and .lowercase with S…

### LINKED ISSUE #6715: angular.lowercase and angular.uppercase methods removed in angular 1.7
angular-ui-grid v4.4.9 is using angular.lowercase and angular.uppercase methods that are now removed in angular 1.7.0-rc.0.

https://github.com/angular/angular.js/blob/v1.7.0-rc.0/CHANGELOG.md

The helper functions angular.lowercase and angular.uppercase have been removed.

These functions have been deprecated since 1.5.0. They are internally used, but should not be exposed as they contain special locale handling (for Turkish) to maintain internal consistency regardless of user-set locale.

Developers should generally use the built-ins toLowerCase and toUpperCase or toLocaleLowerCase and toLocaleUpperCase for special cases.

Further, we generally discourage using the angular.x helpers in application code.

mportuga:
…tring.toUpperCase and .toLowerC

This will allow us to support angular 1.7 when it becomes available.

fix #6715

### DIFF
--- a/src/js/core/services/ui-grid-util.js
+++ b/src/js/core/services/ui-grid-util.js
@@ -243,11 +243,11 @@ module.service('gridUtil', ['$log', '$window', '$document', '$http', '$templateC
       return columnName.replace(/_+/g, ' ')
         // Replace a completely all-capsed word with a first-letter-capitalized version
         .replace(/^[A-Z]+$/, function (match) {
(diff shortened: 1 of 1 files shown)


enderdickerson:
@mportuga  Is there a commit with @oklymenk's change included?

Issue #6784, opened 2018-06-29
v4.6 Export to CSV not working if cellFilter is present


gine:
When i try to export a table in csv and there is cellFilter in some column i recive this error: 

Unknown provider: date:'dd/MM/yy'FilterProvider <- date:'dd/MM/yy'FilterUnknown provider

Unknown provider: bool FilterProvider <- bool provider

This is the colDef conf:

```
{ "field": "azienda", "name": "Cliente", "width": "**", "sort": { "direction": 'asc', "priority": 0 }},
{ "field": "cliente.tipol", "name": "Tipologia", "width": "*"},
{ "field": "nome","name": "nome", "width": "*"},
{ "field": "cognome", "name": "cognome", "width": "*"},
{ "field": "scadPS", "name": "Scad. PS" ,"cellFilter": 'date:\'dd/MM/yy\'', "width": "6%"},
{ "field": "boolstato", "name": "Assunto", "cellFilter":'bool', "width": "5%"}, 
{ "field": "statoCli","name": "Stato Cliente", 'visible': false,  'filters': [{"term":true}]}
```
boolstato have a custom filter 'bool'.
scadPS have a classic date filter: "cellFilter": 'date:\'dd/MM/yy\''

Removing "cellFilter":'bool', and setting "cellFilter": 'date:\'dd/MM/yy\'' to "cellFilter": 'date:'' the export work correctly.

This filters work correctly on the table.

PR #6787, opened 2018-07-03
Fix export csv with cellFilter. Handle filters with 3 arguments

gine:


gine:
This should fix issue #6784

PR #6793, opened 2018-07-05
fix(selection): Fix multiSelect + modifierKeysToMultiSelect

### LINKED ISSUE #6791: options multiSelect + modifierKeysToMultiSelect does not work
I just upgraded from version 4.4.9 and up, and from version 4.4.11 I found that the options multiSelect + modifierKeysToMultiSelect does no longer do what they used to do, namely enable selection of multiple rows without holding down the ctrl key. 

I have compared version 4.6 and 4.4.9 of the ui-grid.js file and I think you have an error on line 27899 (in version 4.6) where you should have "($scope.grid.options.multiSelect && !$scope.grid.options.modifierKeysToMultiSelect)" as the 4'th parameter to the toggleRowSelection function call instead of false or perhaps just add the else from the old code in the end of the if-elseif

mportuga:
Revert part of 2485652

fix #6791

### DIFF
--- a/src/features/selection/js/selection.js
+++ b/src/features/selection/js/selection.js
@@ -955,7 +955,8 @@
               }
               else if ($scope.grid.options.enableSelectRowOnFocus) {
                 uiGridSelectionService.toggleRowSelection($scope.grid, $scope.row, evt,
-                  false, $scope.grid.options.noUnselect);
+                  ($scope.grid.options.multiSelect && !$scope.grid.options.modifierKeysToMultiSelect),
+                  $scope.grid.options.noUnselect);
               }
               $scope.row.setFocused(!$scope.row.isFocused);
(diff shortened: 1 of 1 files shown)


vestergaardmartin:
additional to what you changed, I also removed the line:  $scope.row.setFocused(!$scope.row.isFocused); just bellow your changes because it gave me wrong coloring of the lines (a non-selected line was colored as selected) but I didn't include it in my issue as I thought you should decide what to do regarding this as I would guess that you added this line as a part of some functionality that I don't use and that removing line this would possibly break this functionality.

mportuga:
I addressed that by removing the css which made a focus row have color as part of a different bug fix. Try 4.6.1 and if it is still reproducible, I might remove that line as well.

PR #6820, opened 2018-08-01
fix(fonts): changed space endings on font files

### LINKED ISSUE #6809: ui-grid.ttf is not a valid font file
In 4.6.2, when double clicking on src/font/ui-grid.ttf in Windows, it shows a dialog that indicates "The requested file ...\ui-grid.ttf is not a valid font file".

Looking at a diff of the file between commit 8096ed043b and ec13255f61, it looks like a "0D 0A" was changed to just a "0A", so it appears that a CRLF to LF conversion was done.

The ui-grid.eot file seems to also have gone through a CRLF to LF conversion.

IceDune:
replaced the font files. could'nt open the .eot file. but i guess it will work.

closes #6809


IceDune:
this is for issue #6809

Issue #5227, opened 2016-03-14
Allow formatting of cells through exporterFieldCallback


jmondragon:
I needed to be able to customize the PDF export a little bit more, so I started looking through the documentation and code to see how to do that.  It seems that if you allowed an object return from `exporterFieldCallback` it would give the end user a lot of flexibility.

In my local copy, I just added an `else if` check for object type and return the raw field.value.

Here's the full `formatFieldAsPdfString` function:
```
        formatFieldAsPdfString: function (field) {
          var returnVal;
          if (field.value == null) { // we want to catch anything null-ish, hence just == not ===
            returnVal = '';
          } else if (typeof(field.value) === 'number') {
            returnVal = field.value.toString();
          } else if (typeof(field.value) === 'boolean') {
            returnVal = (field.value ? 'TRUE' : 'FALSE') ;
          } else if (typeof(field.value) === 'string') {
              returnVal = field.value.replace(/"/g, '""');
          } else if (typeof (field.value) === 'object') {
              returnVal = field.value;
          } else {
              returnVal = JSON.stringify(field.value).replace(/^"/,'').replace(/"$/,'');
          }

          if (field.alignment && typeof(field.alignment) === 'string' ){
            returnVal = { text: returnVal, alignment: field.alignment };
          }

          return returnVal;
        }
```

This now allows me to add code like this:
```
gridOptions.exporterFieldCallback = function (grid, row, col, value) {
if (col.name === 'Status') {
  if (value === 'Active')
     value = { text: value, color: 'green' };
  else
     value = { text: value, color: 'red' };
}
return value;
}
```

Thank you,
Josh

jmondragon:
I should note that this does not work if you define `exporterPdfAlign` on the columnDef.

Also, a check should be done in the `formatFieldAsCsv` to check for object type and grab just the text property:

```
        formatFieldAsCsv: function (field) {
          if (field.value == null) { // we want to catch anything null-ish, hence just == not ===
            return '';
          }
          if (typeof(field.value) === 'number') {
            return field.value;
          }
          if (typeof(field.value) === 'boolean') {
            return (field.value ? 'TRUE' : 'FALSE') ;
          }
          if (typeof(field.value) === 'string') {
            return '"' + field.value.replace(/"/g,'""') + '"';
          }
          if (typeof (field.value) === 'object') {
              return field.value.text;
          }
          return JSON.stringify(field.value);
        },
```

Thank you,
Josh

[...]

PR #6852, opened 2018-09-21
Fix CSV export for objects

jmondragon:
When overriding exports with the method described in #5227, the CSV export should still only use the 'text' field of the object.  Otherwise the entire JSON is dumped into the file.


mportuga:
Can you clear the whitespace from the empty line after your if statement?

jmondragon:
I've removed the whitespace, but now the commit-lint-pr seems to be stuck (4 days).  Not sure how to force or restart that process.

Issue #6881, opened 2018-10-30
Sorting behavior is inadvertently altered when minified with uglify default `collapse_vars`


teddy-error:
## Reproduce
- Minify a project that requires ui-grid
  - Webpack (4)
  - Both the built-in minifier and UglifyJS plugin affect ui-grid

[ Will try to add a small PoC project here later]

## Results
- Column sorting (default) _does not_ work as expected

## Workaround
- Disabling the option [collapse_vars](https://github.com/mishoo/UglifyJS2/tree/master#compress-options) appears to resolve the issues.

PR #6882, opened 2018-10-30
Wrap rowSortFn while innards in function to prevent sorting bug

defields923:
When using this module with Webpack and UglifyJs, utilizing the compress with all default options will result in the sorting functionality breaking due to it hoisting and collapsing the propA and propB variables in the rowSortFn method. One simple solution to this issue is to wrap that content in an IIFE to prevent that hosting/collapsing behavior.


defields923:
Issue:

https://github.com/angular-ui/ui-grid/issues/6881

teddy-error:
Thanks for fixing, @defields923

PR #6901, opened 2018-12-03
feat(filter): add filterContainer option to allow a column filter to be placed in column menu

### LINKED ISSUE #3989: Improving user experience - Moving filter section in column menu or make it available in other dropdown
I've started using it for past few weeks, it has really improved a lot with such a great performance. I would like to share few thoughts on filter section, which works great but need few improvements on look and feel. Right now, all filter options are displayed right below the column header and it will be visible all the time. As a user, I want to see it whenever I want to filter, is there any way to move the complete filter section in column menu i.e., after sort options?

Also, I was trying to position the filter section so that I can make it available in a dropdown on user demand. But this breaks the grid header when we have horizontal scroll bar.

http://stackoverflow.com/questions/31422422/angular-ui-grid-moving-filter-section-in-column-menu

Do you have any improvement plans for filter section? Can you please help to improve the look and feel?

caseyjhol:
New behavior: http://next.plnkr.co/edit/nJbWFbLy7qutQslTQovF?preview

filterContainer has a default value of "headerCell" which is the same as existing behavior. A filterContainer value of "columnMenu" moves the filter into the column menu. This improves the appearance of the filter in situations where there are multiple filters for a single column (preventing the header from simply getting taller and taller) and offers a streamlined alternative to placing the filter directly in the header cell.

Fixes #3989.


caseyjhol:
@mportuga Any thoughts on this? Not sure why the commit lint is taking over a week.

src/js/core/directives/ui-grid-header-cell.js
```diff
@@ -328,6 +333,14 @@
 
             updateHeaderOptions();
 
+            if ($scope.col.filterContainer === 'columnMenu' && $scope.col.filterable) {
+              $rootScope.$on('menu-shown', function() {
```
mportuga: Why is $rootScope needed here?

mportuga:
@caseyjhol Overall it looks good. Just one question. And I am not sure why commit lint is being slow.

src/js/core/directives/ui-grid-header-cell.js
```diff
@@ -1,8 +1,8 @@
 (function() {
   'use strict';
 
-  angular.module('ui.grid').directive('uiGridHeaderCell', ['$compile', '$timeout', '$window', '$document', 'gridUtil', 'uiGridConstants', 'ScrollEvent', 'i18nService',
-  function ($compile, $timeout, $window, $document, gridUtil, uiGridConstants, ScrollEvent, i18nService) {
+  angular.module('ui.grid').directive('uiGridHeaderCell', ['$compile', '$timeout', '$window', '$document', 'gridUtil', 'uiGridConstants', 'ScrollEvent', 'i18nService', '$rootScope',
+  function ($compile, $timeout, $window, $document, gridUtil, uiGridConstants, ScrollEvent, i18nService, $rootScope) {
```
mportuga: Could you move rootScope to the start? It is more of a personal preference, but I like having all of the $ variables grouped together.
```suggestion
  function ($compile, $timeout, $window, $document, $rootScope, gridUtil, uiGridConstants, ScrollEvent, i18nService) {
```

src/js/core/directives/ui-grid-header-cell.js
```diff
@@ -1,8 +1,8 @@
 (function() {
   'use strict';
 
-  angular.module('ui.grid').directive('uiGridHeaderCell', ['$compile', '$timeout', '$window', '$document', 'gridUtil', 'uiGridConstants', 'ScrollEvent', 'i18nService',
-  function ($compile, $timeout, $window, $document, gridUtil, uiGridConstants, ScrollEvent, i18nService) {
+  angular.module('ui.grid').directive('uiGridHeaderCell', ['$compile', '$timeout', '$window', '$document', 'gridUtil', 'uiGridConstants', 'ScrollEvent', 'i18nService', '$rootScope',
```
mportuga:
```suggestion
  angular.module('ui.grid').directive('uiGridHeaderCell', ['$compile', '$timeout', '$window', '$document', '$rootScope', 'gridUtil', 'uiGridConstants', 'ScrollEvent', 'i18nService',
```

src/js/core/directives/ui-grid-header-cell.js
mportuga: Also, it might make sense to have tests that specifically change that setFilter works as expected when menu-shown event is triggered,

caseyjhol: The `'menu-shown'` event is emitted from the `uiGridMenu` directive, which is outside the `uiGridHeaderCell` directive, so $rootScope is needed to listen for the event from within `uiGridHeaderCell`. I'm open to suggestions if you think there might be a better way. We could perhaps move the filter stuff to another directive (`'uiGrid'`?).

caseyjhol:
@mportuga You want me to submit another PR to move $rootScope? I can work on some more advanced tests, too.

PR #6903, opened 2018-12-10
fix(uiGridColumMenu): Check column visibility before trying to focus …

### LINKED ISSUE #6871: Possibly unhandled rejection: not-element
Angular.js version: 1.7.3;
ui-grid.js version:  4.6.3;
plunker: https://plnkr.co/edit/GV64cnmcTBedfyA160gW?p=preview

I prepared plunker with different column definition options.

1. If `enableColumnMenu: true` and `enableHiding: false, enableSorting: false` we have empty column menu and we try to open it we got `Possibly unhandled rejection: not-element` error in console.
Steps to reproduce:
   - Open plunker which linked before and Run it
   - In scripts can see that first column definition have options which described before.
   - Try to open 1st column menu
   - Got error in console

2. If `enableColumnMenu: true` and `enableHiding: true, enableSorting: false`  we have column menu with 1 menu item inside `Hide column` and when we try hide it we got `Possibly unhandled rejection: not-element` error in console.
Steps to reproduce:
   - Open plunker which linked before and Run it
   - In scripts can see that second column definition have options which described before.
   - Try to open 2nd column menu
   - Got error in console

MartijnWelker:
…it in the menu-hidden callback

Fixes #6871.

> it's trying to focus on the column even though $scope.col.visible is false (and therefore the column is not rendered and not focusable).
> Skipping the column focus will be fine because setFocusOnHideColumn will handle focussing of the visible columns.

### DIFF
--- a/src/js/core/directives/ui-grid-column-menu.js
+++ b/src/js/core/directives/ui-grid-column-menu.js
@@ -381,7 +381,8 @@ function ($timeout, gridUtil, uiGridConstants, uiGridColumnMenuService, $documen
 
           if ($scope.col && $scope.col.visible) {
             // Focus on the menu button
-            gridUtil.focus.bySelector($document, '.ui-grid-header-cell.' + $scope.col.getColClass()+ ' .ui-grid-column-menu-button', $scope.col.grid, false);
+            gridUtil.focus.bySelector($document, '.ui-grid-header-cell.' + $scope.col.getColClass()+ ' .ui-grid-column-menu-button', $scope.col.grid, false)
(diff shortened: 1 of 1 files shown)


MartijnWelker:
This doesn't fix the core of the issue (not all bySelector/byElement calls have a .catch), but it does fix invalid usage of the bySelector function

mportuga:
Your pull request appears to have broken linting rules by leaving trailing whitespace.

MartijnWelker:
@mportuga My bad ! should be fixed now

PR #6940, opened 2019-02-22
fix(core): prevent uiGrid styles breaking, columns collapsing

Nickproger:
Prevent uiGrid styles breaking(columns collapsing) for those cases when some uiGrids are invisible on page (in some hidden dialogs or tabs).
We don't need to do gridResize() when grid is not visible.


tomo357349:
Please review what work with AngularJs. jqLite may not support is() method...

PR #6945, opened 2019-03-13
feat(filterChanged): pass the changed column as first argument to filterChanged function

### LINKED ISSUE #4775: Shouldn't the "filterChanged" event at least return the column for which the filter has changed?
UI Grid 3.0.6

The "filterChanged" event does not currently return anything as an argument.

The document specifies the following.

> **filterChanged**
> 
> is raised after the filter is changed. The nature of the watch expression doesn't allow notification of what changed, so the receiver of this event will need to re-extract the filter conditions from the columns.

This should mean that due to the watch expression, the old & new filter terms cannot be returned.

There seems to be only 2 instances in which the "filterChanged" event is raised.

Both of them have access to the column for which they are being raised & can be changed to the following.

> **restoreColumns**
> 
> ``` javascript
> grid.api.core.raise.filterChanged ( currentCol );
> ```
> 
> **onDownEvents**
> 
> ``` javascript
> uiGridCtrl.grid.api.core.raise.filterChanged ( $scope.col );
> ```

Is there any problem returning the column for for which the filter has changed?

caseyjhol:
When using the filterChanged function, it's now possible to use the first argument - `column` - to simply do `column.filters[0].term` to retreive the search term for the column being filtered. This is especially useful if using external filtering.

I do not see a simple way to determine the column being changed without this.

Closes #4775.

### DIFF
--- a/packages/core/src/js/directives/ui-grid-header-cell.js
+++ b/packages/core/src/js/directives/ui-grid-header-cell.js
@@ -237,7 +237,7 @@
                   $scope.col.filters.forEach( function(filter, i) {
                     filterDeregisters.push($scope.$watch('col.filters[' + i + '].term', function(n, o) {
                       if (n !== o) {
-                        uiGridCtrl.grid.api.core.raise.filterChanged();
+                        uiGridCtrl.grid.api.core.raise.filterChanged( $scope.col );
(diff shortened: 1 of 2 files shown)


caseyjhol:
@mportuga Not sure why the commit linter is taking so long again. Anything I might be able to check on my end?

PR #7046, opened 2019-10-13
update(ui-grid): changed sequence of right and left container div elements

MsDev88:
**Issue Description :** NVDA is reading random row no while focus comes to first header cell when grouping  grid data.

**Root Cause:** As expand all controls of grouping is part of left container in ui-grid.html, it's div aligned first and then header cell div. Somehow  Narrator is also reading no of grouped row along with header cell row. 

**Proposed Fix:**  Updated right and left container div sequence in ui-grid.html file.

**Before Changes:** 
BeforeChanges
[BeforeChanges_NVDA.zip](https://github.com/angular-ui/ui-grid/files/3721649/BeforeChanges_NVDA.zip)

**After Changes:**
[AfterChange_NVDA.zip](https://github.com/angular-ui/ui-grid/files/3721650/AfterChange_NVDA.zip)

**Testing Details:**
Unit tests - PASS
Manual testing - PASS
E2E tests - Unable to run locally.

### DIFF
--- a/packages/core/src/templates/ui-grid/ui-grid.html
+++ b/packages/core/src/templates/ui-grid/ui-grid.html
@@ -28,7 +28,8 @@
   <div class="ui-grid-contents-wrapper" role="grid">
     <div ui-grid-menu-button ng-if="grid.options.enableGridMenu"></div>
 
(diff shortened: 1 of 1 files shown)


mportuga:
This looks like it would cause a different NVDA problem for Western countries. Where we want the left container to be read first. I don't think that I can approve/merge this change.

PR #7048, opened 2019-10-16
update(keyboard-navigation): keyboard navigation changes of expand and selection buttons

MsDev88:
**Issue Description:** User is not able to access expand and selection buttons through key board.

**Proposed Fix:**  Added tab index and key down events for select and expand buttons.

**After Changes :** Please refer attached video. [AfterChanges_KeyBoardNavigation.zip](https://github.com/angular-ui/ui-grid/files/3733878/AfterChanges_KeyBoardNavigation.zip)

### DIFF
--- a/packages/selection/src/js/selection.js
+++ b/packages/selection/src/js/selection.js
@@ -788,7 +788,7 @@
           }
 
           function selectButtonKeyDown(row, evt) {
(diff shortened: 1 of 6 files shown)


MsDev88:
Thanks for Reply.

Please let me know when can we have new release with below changes.

Thanks,
Umang

PR #7051, opened 2019-11-15
docs: update 317_custom_templates.ngdoc

aland:
docs(templates): update standard template link

Update standard template link as it was 404


calvinmario (commented):
ALANNDXS  : UYY I AM Calvinstore

PR #7059, opened 2019-11-25
refactor: 💡 Expose pdf docDefinition

bhammondbovisync:
Allow use of docDefinition with other technologies


mportuga:
@bhammondbovisync Please fix the unit tests

PR #7090, opened 2020-06-27
fix: update angular to 1.8.0

caok2709:
1. update angular to 1.8.0

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2157,15 +2157,15 @@
       }
     },
     "angular": {
-      "version": "1.7.9",
-      "resolved": "https://registry.npmjs.org/angular/-/angular-1.7.9.tgz",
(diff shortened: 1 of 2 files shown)


caok2709:
https://docs.angularjs.org/guide/migration#migrating-from-1-7-to-1-8

Copy from Angular Team:

Migrating from 1.7 to 1.8
Generally updating to 1.8.0 from 1.7.x should be a straightforward process and is highly recommended. AngularJS 1.8 is a breaking change release from 1.7 to mitigate a security issue.

JqLite no longer turns XHTML-like strings like <div /><span /> to sibling elements when not in XHTML mode: <div></div><span></span>. Instead it will leave the elements alone. In non-XHTML mode the browser will convert these to nested elements: <div><span></span></div>.

This is a security fix to avoid an XSS vulnerability if a new jqLite element is created from a user-controlled HTML string. If you must have this functionality and understand the risk involved then it is posible to restore the original behavior by calling

angular.UNSAFE_restoreLegacyJqLiteXHTMLReplacement();
But you should adjust your code for this change and remove your use of this function as soon as possible.

Note that this only patches jqLite. If you use jQuery 3.5.0 or newer, please read the jQuery 3.5 upgrade guide for more details about the workarounds.

PR #7129, opened 2020-10-24
fix: pass both tree parent nodes and children nodes when exporting

### LINKED ISSUE #6819: Exporter Excludes Parent Nodes (TreeView)
The "All data" option for the exporter is insufficient because it does not actually export all data. When combined with TreeView, it presumes that any parent nodes are merely for grouping and categorizing data, and not part of the data itself. It seems to me that there should be another gridOption to use with exporter that lets you include _all_ data, not just leaf nodes. (And in my opinion that should be the default behavior.)

This is regardless of which export format is chosen. It comes down to the getData and getDataSorted functions in ui-grid.exporter.js.

### LINKED ISSUE #7127: Export all data as excel with Tree View bug.
http://plnkr.co/edit/hFPAO6W3EW8nlzt2
Click "export all data as Excel". Only the leaf nodes are exported.

mportuga:
Fix export all data to excel with treeview bug. Example: http://plnkr.co/edit/hFPAO6W3EW8nlzt2

fix #7127, fix #6819

### DIFF
--- a/packages/exporter/src/js/exporter.js
+++ b/packages/exporter/src/js/exporter.js
@@ -952,21 +952,24 @@
          * recurse down into the children to get to the raw data element
          * which is a row without children (a leaf).
          * @param {Node} aNode the tree node on the grid
-         * @returns {Array} an array of leaf nodes
+         * @returns {Array} an array with all child nodes from aNode
          */
         getRowsFromNode: function(aNode) {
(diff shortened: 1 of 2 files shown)


mportuga:
@alancqueiroz I cleaned up your code and will be merging it today

PR #7136, opened 2020-11-26
fix: change $scope.$parent to appScope where applicable

popsail:
This change prevents an issue where combining a string uiGrid.data and wrapping ui-grid with another component causes data not to be watched and evaluated correctly.
Seems to me we'd want to watch the variable on appScope, not necessarily ui-grid's direct parent.

Otherwise, when using a wrapped uigrid, we'll have to pass '$parent.data' instead of 'data', which is a leaky abstraction at best.

### DIFF
--- a/packages/core/src/js/directives/ui-grid.js
+++ b/packages/core/src/js/directives/ui-grid.js
@@ -44,8 +44,8 @@
       if (self.grid.options.fastWatch) {
         self.uiGrid = $scope.uiGrid;
         if (angular.isString($scope.uiGrid.data)) {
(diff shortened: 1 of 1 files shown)


popsail:
This change expects appScope to be a $scope. 
Passing a vm instead will caused it to fail (it not having $watch etc).

For backward-compatibility, we can condition using $scope.$parent if appScope doesn't have a $watch function.

popsail:
Added commit for backward-compatibility - we now condition watching appScope on it being a valid $scope.
In other cases (i.e it is a vm) we revert to watching $scope.$parent.

popsail:
Seems like appScopProvider: vm use cases don't really get fixed by my solution.
Will work on another approach and open a new pull/issue.

PR #7141, opened 2021-03-10
fix: clearSelectedRows should respect enableSelection 

blackwolf12333:
The issue we had was that we have rows that are selected, but also have `enableSelection: false`. We expected select all to behave similarly when selecting and unselecting all rows in that select all does not touch disabled rows. But unselect all does touch disabled rows.

This fixes the above issue by using the same check that is used during select all behavior in `clearSelectedRows`.

### DIFF
--- a/packages/selection/src/js/selection.js
+++ b/packages/selection/src/js/selection.js
@@ -627,7 +627,7 @@
         clearSelectedRows: function (grid, evt) {
           var changedRows = [];
           service.getSelectedRows(grid).forEach(function (row) {
-            if (row.isSelected) {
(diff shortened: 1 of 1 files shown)


Scrum (approved):

Scrum:
@mportuga

blackwolf12333:
@Scrum should I do something about the lint error? I don't see what's wrong with the commit now tbh.

Scrum:
> @Scrum should I do something about the lint error? I don't see what's wrong with the commit now tbh.

unfortunately, I also do not have access there

mportuga:
That lint error was just an issue with the PR title apparently

Issue #7153, opened 2021-06-29
date with timepart in cellFilter is not exported correctly


rpmelet:
When exporting a date column with a cellFilter 'date:"dd-MM-yyyy HH:mm"' only the date and hours are exported.

This has to do with the way the cellFilter is split in exporter/src/js/exporter.js in the function defaultExporterFieldCallback

I think this will solve the issue. I'll try to create a pull request later.

```
function defaultExporterFieldCallback(grid, row, col, value) {
    // fix to handle cases with 'number : 1' or 'date:"MM-dd-YYYY HH:mm"', etc.. We needed to split the string
    if (col.cellFilter) {
      var args, filter, arg1, arg2;
      // Split on ':' except when in double quotes.
      args = col.cellFilter.match(/(?:[^:"]+|"[^"]*")+/g);
      // remove space, single/double to maintain retro-compatibility, but keep spaces in second argument (arg[1])
      filter = args[0] ? args[0].replace(/[\'\"\s]/g, "") : null;
      arg1 = args[1] ? args[1].replace(/[\'\"]/g, "").trim() : null;
      arg2 = args[2] ? args[2].replace(/[\'\"\s]/g, "") : null;
      return $filter(filter)(value, arg1, arg2);
    } else {
      return value;
    }
  }
```

PR #7154, opened 2021-06-29
fix: export filter with time part. ('date:"MM-dd-YYYY HH:mm').

rpmelet:


rpmelet:
This should fix issue #7153.

mportuga (approved):

mportuga:
@Commit-Lint review

Issue #3527, opened 2015-05-18
Columns to the right of Image Cells disappear while Horizontally Scrolling


JLLeitschuh:
Plunker demonstrating bug:  http://plnkr.co/edit/WhyKGn

System specs: Chrome 42.0.2311.152 (64-bit) on MacBookPro running OSX Yosemite

As you scroll to the right in this example you'll notice that the columns on the left start disappearing.
This seems to be related to how the headers that are disappearing are shortened and using ellipses. If you change the column width so that the column header names are no longer shortened (ie set `dataElementsAfterImagesWidth = 75`) the problem disappears.

PaulL1:
Looks like there's something busted in virtualisation to me.  @swalters, @c0bra - have you seen this before?  Broadly speaking it looks to me like the virtualisation window is being miscalculated, so some visible columns aren't actually there.

PR #7155, opened 2021-07-22
fix: 3527 Columns to the right disappear while Horizontally Scrolling

prx-lmo:
- adjustColumns now calculates the colIndex instead of guessing it with scrollpercentage
 - this is fixing #3527 

Use Case: grid has 100 columns, the first column is enlarged to 50% of grid width, it is scrolled to the middle(50%)

previously: assumes that all columns have the same width; 50% of 100 columns is 50 -> colIndex is set to 50; if excessColumns isn't high enough(default is 4) the cached Columns begin at Index 46

now: accumulates the drawnWidth of all columns till 50%(or whatever left position it has) and returns the index of the column which reached the treshold -> colIndex is set to 0; the cached Columns begin at Index 0


mportuga (approved):

mportuga:
NOTE to self: retest behavior described by the following commit after merging this:
https://github.com/angular-ui/ui-grid/commit/767e0228006b7ddcc9685b502c25733db1d6ff94

Issue #6335, opened 2017-07-25
SyntaxError: Invalid regular expression: /^/[\s\S]*?+$/: Nothing to repeat when entering the filter term "*+" or "*(" etc


shahbazkhan252:
I m using the latest version v4.0.6, before this I was using v4.0.4, I get the following stacktrace after entering "*+", "*(", "*)" as the filter expression. 
    
    SyntaxError: Invalid regular expression: /^[\s\S]*?+$/: Nothing to repeat
    at RegExp (<anonymous>)
    at Object.guessCondition (ui-grid.js:9731)
    at Object.setupFilters (ui-grid.js:9782)
    at Object.search (ui-grid.js:9997)
    at Grid.searchRows (ui-grid.js:5579)
    at startProcessor (ui-grid.js:5073)
    at handleProcessedRows (ui-grid.js:5089)
    at angular.js:15757
    at m.$eval (angular.js:17025)
    at m.$digest (angular.js:16841)

I have used to sort out this by  "condition: uiGridConstants.filter.CONTAINS", it works on grids having server side filtering while its not working for other grids

orneryd:
Thanks for contributing! This repository has been transferred, and we're preparing the new 5.x release. We're closing older issues and pull requests as part of the reset. Please join us on Discord to follow the work or continue the conversation: https://discord.gg/Baz4w8ZWN

Issue #5301, opened 2016-04-08
SyntaxError: Invalid regular expression when entering the filter term "*?"


ret0:
In the following example: http://jsfiddle.net/ts8h2w94/3/ I get the following stacktrace after entering "*?" as the filter expression. This seems to affect all versions that I tested, including the example on the website ("Complex Example" on http://ui-grid.info/).

```
SyntaxError: Invalid regular expression: /^[\s\S]*??$/: Nothing to repeat
    at RegExp (native)
    at Object.guessCondition (http://ui-grid.info/release/ui-grid-unstable.js:9545:14)
    at Object.setupFilters (http://ui-grid.info/release/ui-grid-unstable.js:9592:45)
    at Object.search (http://ui-grid.info/release/ui-grid-unstable.js:9805:59)
    at Grid.searchRows (http://ui-grid.info/release/ui-grid-unstable.js:5400:24)
    at startProcessor (http://ui-grid.info/release/ui-grid-unstable.js:4894:33)
    at handleProcessedRows (http://ui-grid.info/release/ui-grid-unstable.js:4910:20)
    at wrappedCallback (http://ajax.googleapis.com/ajax/libs/angularjs/1.2.26/angular.js:11573:81)
    at wrappedCallback (http://ajax.googleapis.com/ajax/libs/angularjs/1.2.26/angular.js:11573:81)
    at http://ajax.googleapis.com/ajax/libs/angularjs/1.2.26/angular.js:11659:26
```

This issue appears to be similar to http://stackoverflow.com/questions/29554683/angularjs-ui-utils-highlight-filter-breaks-application-on-search

nkoterba:
👍 Just encountered this on our app as well.  When users put in non-standard regex terms (such as `*?` or `?*`), they get this error.

Looks like ui-grid is trying to guess the condition of the filter based on its value (leading *, trailing *, surrounded by *).

We worked around this by explicitly setting a filter condition for each column. See: http://ui-grid.info/docs/#/tutorial/103_filtering

for examples on how to set conditions on filters.  We do most of our processing on the backend/server-side so the actual value of the front-end filter condition doesn't really matter.  By using a value of `uiGridConstants.filter.CONTAINS` we no longer get the Regex exception and functionality appears to be the same (again because we are doing all our filtering server side).

stadler:
Same issue here.

PR #7159, opened 2021-07-26
fix: 6360 error when entering *+ combination in the column filter

prx-lmo:
- updated escapeRegExp according to https://github.com/sindresorhus/escape-string-regexp (where it was originally probably taken from)
 - now escapes term first, before changing *
 - changed * replacement from "[\s\S]*?" to ".*?": [\s\S] == .
 - fixes #7159 , #6335 and #5301


prx-lmo:
this fixes also: https://github.com/angular-ui/ui-grid/issues/6335

mportuga (approved):

PR #7200, opened 2021-12-16
feat (exporter) : Add possibility to create custom style header

NicolasLOJO:
Add the possibility to do custom format on each header cell before put them in excel in case the user want to apply some style or anything else


NicolasLOJO:
Sorry for have some error it's my first contribution

PR #7201, opened 2021-12-17
feat(exporter): add custom header styling callback

NicolasLOJO:
add the possibility to customize your header with a callback for each header cell


mportuga:
@NicolasLOJO This PR appears to have broken the build, you might need to look into why and provide a fix before I can merge it.

PR #7350, opened 2026-05-22
Fix exceptions not logged 7211

### LINKED ISSUE #7211: Exceptions raised by ui-grid or ui-grid plugin code is caught and not logged in the console
When an exception is raised, the end of the following code:

```
function startProcessor(i, renderedRowsToProcess) {
      // Get the processor at 'i'
      var processor = self.rowsProcessors[i].processor;

      // Call the processor, passing in the rows to process and the current columns
      //   (note: it's wrapped in $q.when() in case the processor does not return a promise)
      return $q.when( processor.call(self, renderedRowsToProcess, self.columns) )
        .then(function handleProcessedRows(processedRows) {
          // Check for errors
          if (!processedRows) {
            throw "Processor at index " + i + " did not return a set of renderable rows";
          }

          if (!angular.isArray(processedRows)) {
            throw "Processor at index " + i + " did not return an array";
          }

         // Processor is done, increment the counter
          i++;

          // If we're not done with the processors, call the next one
          if (i <= self.rowsProcessors.length - 1) {
            return startProcessor(i, processedRows);
          }
          // We're done! Resolve the 'finished' promise
          else {
            finished.resolve(processedRows);
          }
        }).catch(angular.noop);
    }
```

In the original code, the exception is rethrown using the following code.

```
}).catch(function(error) {
    throw error;
});
```

The reason for raising this bug is that when using the [ui-grid-auto-fit-columns](https://github.com/Den-dp/ui-grid-auto-fit-columns) plugin with in empty array of columnDefs, the plugin throws an exception but this exception cannot be seen in the console log to see the reason why the plugin is not working.

sardarMrashed:
Fixes #7211 - Changed .catch(angular.noop) to .catch(function(err) { $log.error(err); }) 
so that exceptions are logged to the console instead of being silently ignored.

### DIFF
--- a/packages/core/src/js/directives/ui-grid.js
+++ b/packages/core/src/js/directives/ui-grid.js
@@ -31,7 +31,7 @@
               self.grid.preCompileCellTemplates();
 
               self.grid.refreshCanvas(true);
-            }).catch(angular.noop);
+            }).catch(function(err) { $log.error(err); });
         }) );
       }
 
@@ -118,7 +118,7 @@
             promises.push(self.grid.buildColumns()
               .then(function() {
                 self.grid.preCompileCellTemplates();
-              }).catch(angular.noop));
+              }).catch(function(err) { $log.error(err); }));
           }
 
           $q.all(promises).then(function() {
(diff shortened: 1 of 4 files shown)


orneryd:
Thanks for contributing! This repository has been transferred, and we're preparing the new 5.x release. We're closing older issues and pull requests as part of the reset. Please join us on Discord to follow the work or continue the conversation: https://discord.gg/Baz4w8ZWN

